PR #103, opened 2014-01-10
Create Visualizer::BodyFollower class.

chrisdembia:
To cause the camera to follow objects around in a visualization. The class is a
generalization of the BodyWatcher class in TimsBox.cpp. I have replaced
BodyWatcher in TimsBox.cpp with this new class. I ran TimsBox before and after the change and the camera behavior looks the same to me. I've also used this BodyFollower in the rolling disc example, and it works well there too.

I'm not sure where exactly I should put the new code; let me know if I should relocate the additions.

See some related discussion in #74. @sherm1 said "Attaching the camera rigidly to moving bodies doesn't usually work well either because then the image bounces around too much." I didn't do anything to account for this.


chrisdembia:
Ready for review again.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
 
+/** Causes the camera to point at and follow a point fixed on a body (a
+ station). This might be useful if your system translates substantially
+ and would thus leave the field of view of a stationary camera. **/
+// This class is based on the BodyWatcher class that used to be in TimsBox.cpp.
+class SimTK_SIMBODY_EXPORT Visualizer::BodyFollower :
+    public Visualizer::FrameController {
+public:
+    /**
+    @param[in] bodyB
```
sherm1: This name doesn't match the formal parameter name in the constructor (`body`). I have been switching to `mobod` for MobilizedBodies to avoid ambiguity regarding the Body class (which I should have called something like `BodyInfo`). So I would suggest `mobodB` for this parameter name.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+       The body to follow.
+    @param[in] stationPinB
+       The location of the station P on the body to follow, expressed in B.
```
sherm1: Since it is optional you should mention the default behavior.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+    @param[in] offset
+       Position of the camera from P, expressed in ground.
```
sherm1: Default behavior?

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+    @param[in] upDirection
+       Controls the rotation of the camera about the offset vector. The
+       camera's up (+y) direction will be aligned with this vector as best as
+       possible. Expressed in ground. By default, this is the system's up
+       direction (see getSystemUpDirection); it's unlikely that you want
+       something other than the default.
```
sherm1: Good! Trivia: probably should capitalize `System` here since it refers to an object of that class; add `()` after `getSystemUpDirection()` as the conventional way of designating a function reference.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+    **/
+    BodyFollower(const MobilizedBody& body,
+                          const Vec3&          stationPinB = Vec3(0, 0, 0),
+                          const Vec3&          offset = Vec3(NaN),
+                          const Vec3&          upDirection = Vec3(NaN));
+    
+    virtual void generateControls(
+            const Visualizer&             viz,
+            const State&                  state,
+            Array_< DecorativeGeometry >& geometry) override;
```
sherm1: Since this is an `override`, it should not also say `virtual`.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+       Controls the rotation of the camera about the offset vector. The
+       camera's up (+y) direction will be aligned with this vector as best as
+       possible. Expressed in ground. By default, this is the system's up
+       direction (see getSystemUpDirection); it's unlikely that you want
+       something other than the default.
+    **/
+    BodyFollower(const MobilizedBody& body,
+                          const Vec3&          stationPinB = Vec3(0, 0, 0),
+                          const Vec3&          offset = Vec3(NaN),
+                          const Vec3&          upDirection = Vec3(NaN));
```
sherm1: Alignment of the parameters looks funny here -- the first param doesn't align with the others.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
```diff
@@ -760,6 +764,44 @@ class SimTK_SIMBODY_EXPORT Visualizer::FrameController {
...
+    virtual void generateControls(
+            const Visualizer&             viz,
+            const State&                  state,
+            Array_< DecorativeGeometry >& geometry) override;
+    
+private:
+    const MobilizedBody& m_body;
+    const Vec3 m_stationPinB;
+    const Vec3 m_offset;
+    const Vec3 m_upDirection;
```
sherm1: Style comment: you might consider aligning all the data members here, like:
```
    const MobilizedBody& m_body
    const Vec3           m_stationPinB;
    const Vec3           m_offset;
    const Vec3           m_upDirection;
```
I think it is marginally easier to scan that way, but feel free to ignore.

Simbody/Visualizer/src/Visualizer.cpp
```diff
@@ -1107,6 +1108,52 @@ const Array_<Visualizer::FrameController*>& Visualizer::getFrameControllers() co
...
 
+//============================================================================
+//                             BODY FOLLOWER
+//============================================================================
+Visualizer::BodyFollower::BodyFollower(
+        const MobilizedBody& body,
+        const Vec3&          stationPinB,
+        const Vec3&          offset,
+        const Vec3&          upDirection)
+    :   m_body(body), m_stationPinB(stationPinB), m_offset(offset),
```
sherm1: Suggest: `m_mobod(mobodB)` or something like that.

Simbody/Visualizer/src/Visualizer.cpp
```diff
@@ -1107,6 +1108,52 @@ const Array_<Visualizer::FrameController*>& Visualizer::getFrameControllers() co
...
+        const Visualizer&             viz,
+        const State&                  state,
+        Array_< DecorativeGeometry >& geometry)
+{
+    // Offset.
+    Vec3 offset(m_offset);
+    if (m_offset.isNaN()) {
+        // Default: offset is based on system up direction and ground height.
+        offset = Vec3(1, 1, 1);
+        offset[viz.getSystemUpDirection().getAxis()] += viz.getGroundHeight();
```
sherm1: Slick! I like it.

Simbody/Visualizer/src/Visualizer.cpp
```diff
@@ -1107,6 +1108,52 @@ const Array_<Visualizer::FrameController*>& Visualizer::getFrameControllers() co
...
+    }
+
+    // Up direction.
+    const Vec3* upDirection;
+    if (m_upDirection.isNaN()) {
+        // Default: use system up direction.
+        upDirection = &UnitVec3(viz.getSystemUpDirection()).asVec3();
+    } else {
+        upDirection = &m_upDirection;
+    }
```
sherm1: Use of a pointer here is scary (I assumed it was heap allocated until I read further). A reference is always preferable if you can do it. In this case you can:
```
const Vec3& upDirection = m_upDirection.isNaN() ? UnitVec3(viz.getSystemUpDirection()).asVec3() : m_upDirection;
```
One further thought: can you make `m_upDirection` a `UnitVec3`? I think it is always a unit vector.

Simbody/tests/adhoc/UnilateralPointContactWithFriction.cpp
```diff
@@ -1205,30 +1205,6 @@ class ShowContact : public DecorationGenerator {
 
 
 //==============================================================================
-//                            BODY WATCHER
-//==============================================================================
-// Prior to rendering each frame, point the camera at the given body's
```
sherm1: It's great to see all this code deletion!

sherm1:
I tried the modified examples and the camera behavior looks good. Thanks, Chris!

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: Addressed.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: Done.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: Done.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: Done.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: eek! Done.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: Fixed.

Simbody/Visualizer/include/simbody/internal/Visualizer.h
chrisdembia: I agree; done.

Simbody/Visualizer/src/Visualizer.cpp
chrisdembia: Addressed.

Simbody/Visualizer/src/Visualizer.cpp
chrisdembia: > A reference is always preferable if you can do it.

Done.



> One further thought: can you make m_upDirection a UnitVec3? I think it is always a unit vector.

I tried this, but ran into an issue: I get a segmentation fault when I initialize the UnitVec3 to NaN's. So I just went with Vec3's. I think using a Vec3 also makes the constructor easier to use.

chrisdembia:
Ready for review again.

sherm1:
> I get a segmentation fault when I initialize the UnitVec3 to NaN's. 

The default constructor for UnitVec3 is defined to make it all-NaN:
```
    /// Default constructor initializes to all-NaN even in Release mode so that
    /// we maintain the above-promised contract.
    UnitVec() : BaseVec(NTraits<P>::getNaN()) {}
```
> So I just went with Vec3's. I think using a Vec3 also makes the constructor easier to use.

But if (0,0,0) isn't an allowed value (because it is really a direction) then you have to do that check yourself; UnitVec3 understands that already so will do the checking. Also (probably not an issue here) a Vec3 must be normalized if it is to be used as a direction. That is an expensive operation so generally it is nice to avoid it if possible. If the supplied vector is given as a UnitVec3 then it doesn't have to be normalized, in case the user already has a unit vector available (common).

chrisdembia:
> The default constructor for UnitVec3 is defined to make it all-NaN:

Ah so I must have messed up somewhere else.

> That is an expensive operation so generally it is nice to avoid it if possible.

It seems the expense would be incurred [here](https://github.com/simbody/simbody/blob/48c3cc037e47c42770bf2e0f708814976bf8a739/SimTKcommon/Mechanics/src/Rotation.cpp#L553), when we get the norm of the up direction vec3. Does `normSqr()` on `UnitVec3` avoid computation?

I'll try again to use a UnitVec3.

chrisdembia:
I pushed a new commit. Now, `upDirection` is a `UnitVec3`, and the member variables are const references.

sherm1:
> It seems the expense would be incurred here, when we get the norm of the up direction vec3.

Yes, you are right. In this case there isn't a performance advantage since the unit vector gets passed to a method that expects a vec3. But you should decide based on what the interface to your routine is. If it really does require a unit vector it should accept one.

chrisdembia:
> But you should decide based on what the interface to your routine is. If it really does require a unit vector it should accept one.

Good point; I want my interface to take a UnitVec3.

sherm1:
Actually, I can't tell if the called routine actually normalizes the vector. normSqr() alone is much cheaper than normalization. normSqr() is a dot product: 5 flops. normalization is that same dot product, a square root, and at least one divide. Roughly 25 flops. Anyway I think it is safe to say there is no real performance issue to worry about in this context so I'm happy with whatever you think is the nicest interface.

sherm1:
Looks good now -- ready to merge pending CI.

chrisdembia:
Thanks for reviewing.

PR #271, opened 2014-11-27
SuperEllipsoid Contact Geometry

danielsimoeslopes:
The code I edited is for creating a new contact geometry, the superellipsoid. 
Please take a look and let me now what else is required to properly integrate this surface to the contact family.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/271)


adamkewley:

Issue #264, opened 2014-11-11
Handling relative paths. Confusion with deconstructPathname


chrisdembia:
I am working on https://github.com/opensim-org/opensim-core/issues/118. @sherm1 , in this issue you claim that

> FYI, Simbody has a utility class SimTK::Pathname that provides platform-independent handling of relative path names so that changing directory isn't required.

Based on the Pathname API, this does not seem possible. So I have been trying to alter Pathname::getAbsolutePathname to suit my needs.

My objective is to take a path `pathname` that is relative an arbitrary path `base`, and express it as either (1) a relative path that is relative to my cwd, or (2) an absolute path.

I made the necessary changes to `getAbsolutePathname` so that it could use a `base` that is not the cwd. However, I misunderstood what `deconstructPathname` does. The documentation contains the confusing statement:

> "." starts an absolute path name which is relative to the current working directory

How can an absolute path be relative to something else? I fed this method a path like `../../myfile` and was told that this is an absolute path. I learned that `deconstructPathname` is trying to do more than it advertises: it converts relative paths to absolute paths if you pass it "keywords" such as `../`, `./`, `@`. So from this perspective, indeed, `./` specifies an absolute path. I feel this is improper; `getAbsolutePathname` should do that; I simply wanted to deconstruct the path as given. 

I believe my current method of trying to do (1) or (2) cannot be done simply by modifying `getAbsolutePathname`. 

@sherm1 does Pathname, as is, provide the functionality I seek? If not, I would like to alter `deconstructPathname` so that it strictly decomposes a given path, and does not try to create an absolute path based on keywords. I'm happy to take other suggestions as well.

sherm1:
Chris, I believe PR #307 has addressed this issue which you opened. Please close if you agree.

chrisdembia:
:fireworks:

PR #307, opened 2014-12-20
deconstructPathRelativeToSWD

carmichaelong:
Addresses fixes to deconstructPathname (addresses issue #264). Introduces a new function deconstructPathRelativeToSWD that finds the absolute path to a given path name relative to a specified working directory (swd).

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -89,23 +89,26 @@ class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
     /// Given a specified working directory (swd) and path, this function
     /// evaluates the absolute path of a given path relative to the swd and
-    /// returns the directory, fileName, and extension of the canonicalized absolute
-    /// path. This means that instead of evaluating "." as the current working
-    /// directory (cwd), the swd is used. In general, some rules are as follows
+    /// returns the directory, fileName, and extension of the canonicalized path
+    /// with respect to a swd, if needed. This means, for the path, that instead of 
+    /// evaluating "." as the current working directory (cwd), the swd is used. This
+    /// function also returns a boolean (dontApplySearchPath) to indicate if a search
+    /// path should not be applied. The following rules are applied in order:
     /// 1) If an absolute path is given by path, swd is not used. An absolute path
-    ///    is defined as a a root-relative path name and on Windows it will begin 
-    ///    with an explicit drive letter (e.g. /usr/* or c:/documents/*, respectively).
-    /// 2) If a relative path without a drive letter is given by path (e.g. "./*" or "*"), 
-    ///    then the absolute path of the swd is prepended to path.
+    ///    is defined as a a root-relative path name and on Windows it may begin 
+    ///    with a drive letter (e.g. /usr/file.ext or c:/documents/file.ext).
+    /// 2) If a relative path without a drive letter is given by path (e.g. "./dir/file.ext" 
+    ///    or "dir/file.ext"), then the absolute path of the swd is prepended to path.
     /// 3) To resolve drive ambiguities, if swd provides a drive, it is used. 
-    ///    If not, then the path drive is used (e.g. path = "X:*"). If neither provides 
-    ///    a drive, then the current drive is used.
+    ///    If not, then the path drive is used (e.g. path = "X:dir/file.ext"). If neither 
+    ///    provides a drive, then the current drive is used.
     /// 4) If swd is an empty string, then swd = cwd.
-    static void findAbsolutePathUsingSpecifiedWorkingDirectory(const std::string& swd,
+    static void deconstructPathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
                                              const std::string& path,
                                              std::string& directory,
                                              std::string& fileName,
-                                             std::string& extension);
+                                             std::string& extension,
+                                             bool& dontApplySearchPath);
 
     /// Dismantle a supplied pathname into its component
     /// parts. This can take pathnames like <pre>   
@@ -185,6 +188,17 @@ public:
         if (!absPath.empty() && absPath[absPath.size()-1] != getPathSeparatorChar())
             absPath += getPathSeparatorChar();
         return absPath;
+    }
+
+    static std::string getAbsolutePathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
+                                                                         const std::string& path) {
+        std::string directory, fileName, extension;
+        bool dontApplySearchPath;
+        deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension, dontApplySearchPath);
+        if (!dontApplySearchPath) {
+            directory = getCurrentWorkingDirectory() + directory;
+        }
+        return directory + fileName + extension;
     }
 
     /// Return true if the given pathname names a file that exists and is
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -91,7 +91,7 @@ public:
     /// evaluates the absolute path of a given path relative to the swd and
     /// returns the directory, fileName, and extension of the canonicalized absolute
     /// path. This means that instead of evaluating "." as the current working
-    /// directory (cwd), the swd is used. In general, some rules followed are as follows
+    /// directory (cwd), the swd is used. In general, some rules are as follows
     /// 1) If an absolute path is given by path, swd is not used. An absolute path
     ///    is defined as a a root-relative path name and on Windows it will begin 
     ///    with an explicit drive letter (e.g. /usr/* or c:/documents/*, respectively).
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -163,6 +163,19 @@ public:
                                                                   std::string& directory,
                                                                   std::string& fileName,
                                                                   std::string& extension);
+    
+    /// Give back the deconstructed canonicalized absolute pathname for a given path.
+    /// If the path is not an absolute path, it will be made into an absolute path first
+    /// following the rules of deconstructPathname() (which it uses).
+    static void deconstructAbsolutePathname(const std::string& path,
+                                            std::string& directory,
+                                            std::string& fileName,
+                                            std::string& extension) {
+        bool dontApplySearchPath;
+        deconstructPathname(path, dontApplySearchPath, directory, fileName, extension);
+        if (!dontApplySearchPath)
+            directory = getCurrentWorkingDirectory() + directory;
+    }
 
     /// Get canonicalized absolute pathname from a given pathname which 
     /// can be relative or absolute. Canonicalizing means
@@ -182,11 +195,8 @@ public:
     /// from deconstructPathname(), plus inserting the current working
     /// directory in front if the path name was relative.
     static std::string getAbsolutePathname(const std::string& pathname) {
-        bool isAbsolutePath;
         std::string directory, fileName, extension;
-        deconstructPathname(pathname, isAbsolutePath, directory, fileName, extension);
-        if (!isAbsolutePath)
-            directory = getCurrentWorkingDirectory() + directory;
+        deconstructAbsolutePathname(pathname, directory, fileName, extension);
         return directory + fileName + extension;
     }
 
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -128,7 +128,7 @@ public:
     /// for the currently running platform (i.e. backslash for
     /// Windows and forward slash everywhere else).
     static void deconstructPathname(    const std::string& name,
-                                        bool&        isAbsolutePath,
+                                        bool&        dontApplySearchPath,
                                         std::string& directory,
                                         std::string& fileName,
                                         std::string& extension);
@@ -139,7 +139,7 @@ public:
     /// the canonicalized path with respect to a swd, if needed. This means, for 
     /// the path, that instead of evaluating "." as the current working directory 
     /// (cwd), the swd is used. Unlike deconstructPathname(), this function will
-    /// always return an absolute path, and no bool isAbsolutePath is returned.
+    /// always return an absolute path, and no bool dontApplySearchPath is returned.
     /// Rules:
     /// - If the swd is empty (after removing whitespace), deconstructPathname()
     ///   is called, and cwd is prepended if needed to make it an absolute path.
@@ -151,8 +151,8 @@ public:
     /// 2) If path is a root-relative path name (and on Windows this includes a drive) 
     ///    (e.g. /usr/file.ext or c:/documents/file.ext), then swd is ignored, and the
     ///    absolute path is returned.
-    /// 3) Otherwise, if a path is given relative to the cwd (e.g. "./dir/file.ext" 
-    ///    or "dir/file.ext"), then the swd is prepended to path.
+    /// 3) Otherwise, if a path is given relative to a directory that is not the root 
+    ///    (e.g. "./dir/file.ext" or "dir/file.ext"), then the swd is prepended to path.
     /// 4) To resolve drive ambiguities, if swd provides a drive, it is used. If not, 
     ///    then the path drive is used. If neither provides a drive, then the current 
     ///    drive is used.
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -87,28 +87,7 @@ namespace SimTK {
  */
 class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
-    /// Given a specified working directory (swd) and path, this function
-    /// evaluates the absolute path of a given path relative to the swd and
-    /// returns the directory, fileName, and extension of the canonicalized path
-    /// with respect to a swd, if needed. This means, for the path, that instead of 
-    /// evaluating "." as the current working directory (cwd), the swd is used. This
-    /// function also returns a boolean (dontApplySearchPath) to indicate if a search
-    /// path should not be applied. The following rules are applied in order:
-    /// 1) If an absolute path is given by path, swd is not used. An absolute path
-    ///    is defined as a a root-relative path name and on Windows it may begin 
-    ///    with a drive letter (e.g. /usr/file.ext or c:/documents/file.ext).
-    /// 2) If a relative path without a drive letter is given by path (e.g. "./dir/file.ext" 
-    ///    or "dir/file.ext"), then the absolute path of the swd is prepended to path.
-    /// 3) To resolve drive ambiguities, if swd provides a drive, it is used. 
-    ///    If not, then the path drive is used (e.g. path = "X:dir/file.ext"). If neither 
-    ///    provides a drive, then the current drive is used.
-    /// 4) If swd is an empty string, then swd = cwd.
-    static void deconstructPathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
-                                             const std::string& path,
-                                             std::string& directory,
-                                             std::string& fileName,
-                                             std::string& extension,
-                                             bool& dontApplySearchPath);
+
 
     /// Dismantle a supplied pathname into its component
     /// parts. This can take pathnames like <pre>   
@@ -154,6 +133,35 @@ public:
                                         std::string& fileName,
                                         std::string& extension);
 
+    /// An extension of deconstructPathname(). Given a specified working directory 
+    /// (swd) and path, this function evaluates the absolute path of a given path
+    /// relative to the swd and returns the directory, fileName, and extension of 
+    /// the canonicalized path with respect to a swd, if needed. This means, for 
+    /// the path, that instead of evaluating "." as the current working directory 
+    /// (cwd), the swd is used. Unlike deconstructPathname(), this function will
+    /// always return an absolute path, and no bool isAbsolutePath is returned.
+    /// Rules:
+    /// - If the swd is empty (after removing whitespace), deconstructPathname()
+    ///   is called, and cwd is prepended if needed to make it an absolute path.
+    /// - Otherwise, we evaluate path relative to the swd. These steps are as follows:
+    /// 1) Preprocess the swd. This means that if the swd is of any form that denotes
+    ///    an absolute path (i.e. "C:/file.ext", "C:file.ext", "./file.ext", "/file.ext")
+    ///    we change the swd to reflect the absolute path (e.g. "./file.ext" may change
+    ///    to "/cwd/file.ext" or "C:/cwdOnC/file.ext").
+    /// 2) If path is a root-relative path name (and on Windows this includes a drive) 
+    ///    (e.g. /usr/file.ext or c:/documents/file.ext), then swd is ignored, and the
+    ///    absolute path is returned.
+    /// 3) Otherwise, if a path is given relative to the cwd (e.g. "./dir/file.ext" 
+    ///    or "dir/file.ext"), then the swd is prepended to path.
+    /// 4) To resolve drive ambiguities, if swd provides a drive, it is used. If not, 
+    ///    then the path drive is used. If neither provides a drive, then the current 
+    ///    drive is used.
+    static void deconstructPathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
+                                                                  const std::string& path,
+                                                                  std::string& directory,
+                                                                  std::string& fileName,
+                                                                  std::string& extension);
+
     /// Get canonicalized absolute pathname from a given pathname which 
     /// can be relative or absolute. Canonicalizing means
     ///   - drive designator is recognized if we're on Windows;
@@ -190,14 +198,10 @@ public:
         return absPath;
     }
 
-    static std::string getAbsolutePathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
-                                                                         const std::string& path) {
+    static std::string findAbsolutePathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
+                                                                          const std::string& path) {
         std::string directory, fileName, extension;
-        bool dontApplySearchPath;
-        deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension, dontApplySearchPath);
-        if (!dontApplySearchPath) {
-            directory = getCurrentWorkingDirectory() + directory;
-        }
+        deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension);
         return directory + fileName + extension;
     }
 
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -87,21 +87,21 @@ namespace SimTK {
  */
 class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
-	/// Given a specified working directory (swd) and path, this function
-	/// evaluates the absolute path of a given path relative to the swd and
-	/// returns the directory, fileName, and extension of the final absolute
-	/// path. In general, some rules followed are as follows
-	/// 1) If a full absolute path is given by path, swd is not used.
-	/// 2) To resolve drive ambiguities, in general, if swd provides a drive,
-	///    it is used. If not, then the path drive is used. If neither provides
-	///    a drive, then the current drive is used.
-	/// 3) If swd is an empty string, then the path is evaluated as normal
-	///    in a shell.
-	static void deconstructPathRelativeToSWD(const std::string& swd,
-											 const std::string& path,
-											 std::string& directory,
-											 std::string& fileName,
-											 std::string& extension);
+    /// Given a specified working directory (swd) and path, this function
+    /// evaluates the absolute path of a given path relative to the swd and
+    /// returns the directory, fileName, and extension of the final absolute
+    /// path. In general, some rules followed are as follows
+    /// 1) If a full absolute path is given by path, swd is not used.
+    /// 2) To resolve drive ambiguities, in general, if swd provides a drive,
+    ///    it is used. If not, then the path drive is used. If neither provides
+    ///    a drive, then the current drive is used.
+    /// 3) If swd is an empty string, then the path is evaluated as normal
+    ///    in a shell.
+    static void deconstructPathRelativeToSWD(const std::string& swd,
+                                             const std::string& path,
+                                             std::string& directory,
+                                             std::string& fileName,
+                                             std::string& extension);
 
     /// Dismantle a supplied pathname into its component
     /// parts. This can take pathnames like <pre>   
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -89,15 +89,19 @@ class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
     /// Given a specified working directory (swd) and path, this function
     /// evaluates the absolute path of a given path relative to the swd and
-    /// returns the directory, fileName, and extension of the final absolute
-    /// path. In general, some rules followed are as follows
-    /// 1) If a full absolute path is given by path, swd is not used.
-    /// 2) To resolve drive ambiguities, in general, if swd provides a drive,
-    ///    it is used. If not, then the path drive is used. If neither provides
+    /// returns the directory, fileName, and extension of the canonicalized absolute
+    /// path. This means that instead of evaluating "." as the current working
+    /// directory (cwd), the swd is used. In general, some rules followed are as follows
+    /// 1) If an absolute path is given by path, swd is not used. An absolute path
+    ///    is defined as a a root-relative path name and on Windows it will begin 
+    ///    with an explicit drive letter (e.g. /usr/* or c:/documents/*, respectively).
+    /// 2) If a relative path without a drive letter is given by path (e.g. "./*" or "*"), 
+    ///    then the absolute path of the swd is prepended to path.
+    /// 3) To resolve drive ambiguities, if swd provides a drive, it is used. 
+    ///    If not, then the path drive is used (e.g. path = "X:*"). If neither provides 
     ///    a drive, then the current drive is used.
-    /// 3) If swd is an empty string, then the path is evaluated as normal
-    ///    in a shell.
-    static void deconstructPathRelativeToSWD(const std::string& swd,
+    /// 4) If swd is an empty string, then swd = cwd.
+    static void findAbsolutePathUsingSpecifiedWorkingDirectory(const std::string& swd,
                                              const std::string& path,
                                              std::string& directory,
                                              std::string& fileName,
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -141,16 +141,18 @@ public:
     /// (cwd), the swd is used. Unlike deconstructPathname(), this function will
     /// always return an absolute path, and no bool dontApplySearchPath is returned.
     /// Rules:
+    /// - If path is empty, directory, fileName and extension will be returned empty.
+    ///   This case probably should just use getCurrentworkingDirectory().
     /// - If the swd is empty (after removing whitespace), deconstructPathname()
     ///   is called, and cwd is prepended if needed to make it an absolute path.
     /// - Otherwise, we evaluate path relative to the swd. These steps are as follows:
-    /// 1) Preprocess the swd. This means that if the swd is of any form that denotes
+    /// 1) If path is a root-relative path name (and on Windows this includes a drive) 
+    ///    (e.g. /usr/file.ext or c:/documents/file.ext), then swd is ignored, and the
+    ///    absolute path is returned.
+    /// 2) Preprocess the swd. This means that if the swd is of any form that denotes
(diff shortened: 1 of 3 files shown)


carmichaelong:
Updated documentation and split test cases for what would be expected for Windows and not-Windows platforms (e.g. X:/dir is an absolute path on Windows but is a relative path on not-Windows). Please review again, it'll need more bashing for documentation and to choose the right set of test cases (they are currently enumerating a ton of possible combinations, but they are not all uniquely useful).

chrisdembia:
I think that `findAbsolutePathUsingSpecifiedWorkingDirectory()` should return a single string, since that's what its name implies it will do.

> or else you'd have to "deconstruct" again.

Why would you have to deconstruct again?

I think another reason to move a lot of the code into a separate method is for re-use by `deconstructPathname()`.

carmichaelong:
When finding the absolute path, as currently written, you automatically have the directory, fileName, and ext all separated, so to put everything together as one string is another step. Then, if this method were used in `deconstructPathname()`, then it would have been a waste to put everything together first (though, not like this is that expensive).

I agree with what you said that a method named like this would be expected to return a single string,

chrisdembia:
> Then, if this method were used in deconstructPathname(), then it would have been a waste to put everything together first (though, not like this is that expensive).

Ah! I see. I figured that the guts of `findAbsolutePathUsingSpecifiedWorkingDirectory()` would be moved to a new method `deconstructPathnameUsingSpecifiedWorkingDirectory()`. You, I think, were thinking of the opposite.

How about if we end up with these methods in the end:

* `deconstructPathname()`: same functionality as it currently has, but implementation will call `deconstructPathnameUsingSpecifiedWorkingDirectory()`
* `findAbsolutePathUsingSpecifiedWorkingDirectory()` returns a single string by concatenating the results from `deconstructPathnameUsingSpecifiedWorkingDirectory()`.
* `deconstructPathnameUsingSpecifiedWorkingDirectory()` will contain the functionality that was previously in both the above two methods.

?

The name `deconstructPathnameUsingSpecifiedWorkingDirectory()` might be the wrong name.

sherm1:
:+1:

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -87,6 +87,26 @@ namespace SimTK {
  */
 class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
+    /// Given a specified working directory (swd) and path, this function
+    /// evaluates the absolute path of a given path relative to the swd and
+    /// returns the directory, fileName, and extension of the canonicalized absolute
+    /// path. This means that instead of evaluating "." as the current working
+    /// directory (cwd), the swd is used. In general, some rules are as follows
```
chrisdembia: We should mention that the order of the rules here gives their precedence.

chrisdembia:
If "./" appears at the beginning of "path", it is treated as if that didn't exist, unless swd == "", right?

chrisdembia:
I just looked at every test case. Very thorough! I think your documentation of the new method makes sense. I may take a pass through it at some point later on.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -87,6 +87,26 @@ namespace SimTK {
...
+    /// 1) If an absolute path is given by path, swd is not used. An absolute path
+    ///    is defined as a a root-relative path name and on Windows it will begin
```
sherm1: "and on Windows it *may* begin with a drive letter" (it is still root-relative if it begins with a slash, but it is the root of the current drive). I suppose if there were an swd with a drive letter it would grab that drive rather than the current one, right? But then ignore the rest of the swd?

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -87,6 +87,26 @@ namespace SimTK {
...
 public:
+    /// Given a specified working directory (swd) and path, this function
+    /// evaluates the absolute path of a given path relative to the swd and
+    /// returns the directory, fileName, and extension of the canonicalized absolute
+    /// path. This means that instead of evaluating "." as the current working
+    /// directory (cwd), the swd is used. In general, some rules are as follows
+    /// 1) If an absolute path is given by path, swd is not used. An absolute path
+    ///    is defined as a a root-relative path name and on Windows it will begin 
+    ///    with an explicit drive letter (e.g. /usr/* or c:/documents/*, respectively).
+    /// 2) If a relative path without a drive letter is given by path (e.g. "./*" or "*"),
```
sherm1: I don't think it is good to use `*` here because it looks like a glob pathname (file regular expression). It would be easy for people to misinterpret this to mean they can provide patterns (or am I worrying too much?). An alternative might be something innocuous like `xxx`.

sherm1:
> If "./" appears at the beginning of "path", it is treated as if that didn't exist, unless swd == "", right?

The documentation says swd=="" means swd=cwd. So then I guess the `./` still wouldn't change anything. AFAIK, the leading `./` is only significant if you are trying to decide whether to apply a search path. So if the shell sees `grep` it applies the search path, but `./grep` must be found in the current directory. Since the new method doesn't attempt to make this determination (like the old one does), the `./` shouldn't matter.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
chrisdembia: I actually was thinking the same thing. `xxx` sounds fine, or just making up a normal-looking path.

chrisdembia:
@sherm1 thanks for clarifying.

The current test cases are so thorough! Really nice. However, could you add test cases for when `path` is `.`, `..`, etc., or a directory? It seems like the method has some logic for handling these cases.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
carmichaelong: Yeah this is why the wording here has to be careful. On Windows, if the path is root-relative with a drive letter, it will ignore swd. If it the path does not have a drive letter, it will then look at the swd for the drive letter, to resolve ambiguities (rule 3). How to clearly state this is tricky, since we expect different behaviors on a Windows machine vs. not-Windows.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
carmichaelong: Didn't consider that, but that's a good point. Will replace with innocuous or normal looking path things.

carmichaelong:
> How about if we end up with these methods in the end:
>
> `deconstructPathname()`: same functionality as it currently has, but implementation will call `deconstructPathnameUsingSpecifiedWorkingDirectory()`
> `findAbsolutePathUsingSpecifiedWorkingDirectory()` returns a single string by concatenating the results from `deconstructPathnameUsingSpecifiedWorkingDirectory()`.
> `deconstructPathnameUsingSpecifiedWorkingDirectory()` will contain the functionality that was previously in both the above two methods.

Looking at this more, I realized there are some important differences that may change how we structure this.
* An important function of `deconstructPathname()` is to check for `isAbsolutePath`, or rather check if the search path should apply. Currently the new function `deconstructPathnameUsingSpecifiedWorkingDirectory()` does not do that.
* To deal with this, one way would actually be to see that in the case where `swd` is the empty string, then `deconstructPathnameUsingSpecifiedWorkingDirectory()` just calls on `deconstructPathname()` (reversing the structure above)
* Doing the above, then it's clear what `isAbsolutePath` should return when `swd` is the empty string (preserving prior behavior). But then another question comes up which is, how should one handle `isAbsolutePath` when `swd` is not the empty string? At first it seemed to me that this should always return `true`, but I'm not convinced of that now. Or even a higher level question, should this function be dealing with `isAbsolutePath`?
* deconstructPathnameUsingSpecifiedWorkingDirectory()` currently returns `directory`, `fileName`, and `ext`, but the `directory` is the absolute path. This also differs from `deconstructPathname()` because that function only returns the part of the directory that was passed in since it does not try to resolve what the user really means by the "working directory".

All of this makes me wonder if we actually have two functions whose purposes are actually distinct.

sherm1:
> All of this makes me wonder if we actually have two functions whose purposes are actually distinct.

Somewhere way back I think we discussed the fact that I had misnamed the original routine since it is really about whether or not a search path should apply, not really about whether there is an absolute pathname. Stretching my memory a bit though ...

chrisdembia:
Oh gosh...I don't think I was born with a great ability to work with paths.

carmichaelong:
> Somewhere way back I think we discussed the fact that I had misnamed the original routine since it is really about whether or not a search path should apply, not really about whether there is an absolute pathname.

I tried to just refer to that as returning the correct answer to `isAbsolutePath` above, since the meaning of it here doesn't make a difference. But yes, I think it should be changed to `shouldApplySearchPath` at some point, but this will flip all true and false answers.

The big point for this PR is that, I realize that `deconstructPathname()` has two purposes. 1) separate directory, fileName, and ext; and 2) assign a true or false to `isAbsolutePath`. The second purpose brings up a question about how the new `deconstructPathnameUsingSpecifiedWorkingDirectory()` should behave. With our first iterations of what this should do, it didn't need to deal with assigning something to `isAbsolutePath`. But if this is the lower-level function, then it needs to. However, then that begs the question, what does that flag mean if you pass in a swd? If it doesn't mean anything, then `deconstructPathname()` should just be called as a "special case" when swd is the empty string. Otherwise, we have to think about what behavior we want when swd is not just the empty string.

sherm1:
Better to rename the flag something like `dontApplySearchPath` or `shouldUseLiterally` or some other name that has the same sense as `isAbsolutePath`. Otherwise existing code would break.

carmichaelong:
Ready to be looked at again. `deconstructPathnameUsingSpecifiedWorkingDirectory()` has been reworked, and `deconstructPathname()` uses it now. `getAbsolutePathUsingSpecifiedWorkingDirectory()` has also been implemented and tests added.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -87,6 +87,29 @@ namespace SimTK {
  */
 class SimTK_SimTKCOMMON_EXPORT Pathname {
 public:
+    /// Given a specified working directory (swd) and path, this function
+    /// evaluates the absolute path of a given path relative to the swd and
+    /// returns the directory, fileName, and extension of the canonicalized path
+    /// with respect to a swd, if needed. This means, for the path, that instead of 
+    /// evaluating "." as the current working directory (cwd), the swd is used. This
+    /// function also returns a boolean (dontApplySearchPath) to indicate if a search
+    /// path should not be applied. The following rules are applied in order:
```
sherm1: This should be more clear about the meaning of this flag. Something like "... to indicate whether it would be appropriate to apply a search path if this is being used in a search path context, like command look up in a shell. If a specific directory was specified in some way then a search should not be done. For example `mypath` or `mypath/myfile` could be searched, but `./mypath` could not."

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -167,6 +190,17 @@ class SimTK_SimTKCOMMON_EXPORT Pathname {
         return absPath;
     }
 
+    static std::string getAbsolutePathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
```
sherm1: Let's call this `findAbsolute ...` rather than `getAbsolute ...`. Simbody's convention is that `get` means "things you can get for free with no calculation", that is, some kind of accessor. Some other verb could work to, like `construct`.

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+    // Check for a drive, then map swd to an absolute path name if no drive given
+    String swdcleaned = String::trimWhiteSpace(swd).replaceAllChar('\\', '/');
+    if (!swdcleaned.empty() && swdcleaned[swdcleaned.size() - 1] != '/')
+        swdcleaned += '/';
+    removeDriveInPlace(swdcleaned, swddrive);
+
+    String processed;
+    if (pathcleaned.substr(0, 1) == "/" || pathcleaned.substr(0, 2) == "@/") {
+        processed = pathcleaned;
+        if (!pathdrive.empty()) swddrive = "";
```
sherm1: Here (path is absolute) it looks like a drive conflict between path and swd is resolved in favor of the path. I don't remember seeing that in the documentation -- I think it says drive conflicts are resolved in favor of swd. I think it does make sense to handle it this way here, but the exception should be noted.

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+    }
+    else if (pathcleaned.substr(0, 2) == "./") {
+        if (!swd.empty()) {
```
sherm1: `swd` was not cleaned, so it will not test empty even if it is just a bunch of blanks. So I don't think this is the right test. That also suggests that the test suite should include some tests where `swd` is non-empty but just whitespace.

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+        if (!swddrive.empty()) finaldrive = swddrive;
+        else if (!pathdrive.empty()) finaldrive = pathdrive;
+        else finaldrive = getCurrentDriveLetter();
+    }
+    else if (processed.substr(0, 2) == "./") {
+        dontApplySearchPath = true;
+        String discard;
+        // If swd was empty, then "./" came from path. Make sure not to write
+        // over current pathdrive or swddrive.
+        if (swd.empty()) {
```
sherm1: Same comment here -- I don't think this is the right test.

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+            removeDriveInPlace(processed, discard);
+        }
+        // Otherwise, "./" came from swd
+        else {
+            processed.replace(0, 2, getCurrentWorkingDirectory(swddrive));
+            removeDriveInPlace(processed, discard);
+        }
+        if (!swddrive.empty()) finaldrive = swddrive;
+        else if (!pathdrive.empty()) finaldrive = pathdrive;
+        else finaldrive = getCurrentDriveLetter();
```
sherm1: There is something wrong with the logic here. Why is it so complicated compared to the "@/" case?

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+    }
+    else if (processed.substr(0, 2) == "@/") {
+        dontApplySearchPath = true;
+        processed.replace(0, 2, getThisExecutableDirectory());
+        removeDriveInPlace(processed, finaldrive);
```
sherm1: Nice!

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+    }
+    // Looks like a relative path name. But if either swd or path had 
+    // an initial drive specification, e.g. X:something.txt, that is 
+    // supposed to be interpreted relative to the current working directory
+    // on drive X, just as though it were X:./something.txt.
```
sherm1: Ah, I think this makes sense but I don't think it was in the API documentation.

SimTKcommon/src/Pathname.cpp
```diff
@@ -110,94 +110,117 @@ static void removeDriveInPlace(string& inout, string& drive) {
...
+        removeDriveInPlace(processed, finaldrive);
+    }
+    else if (!pathdrive.empty()) {
+        dontApplySearchPath = true;
+        // Even if swd did not provide a drive, swd may not have been empty.
+        // Make sure to still use pathdrive.
+        if (!swd.empty()) {
+            processed.insert(0, getCurrentWorkingDirectory(swddrive));
+            removeDriveInPlace(processed, swddrive);
+            finaldrive = pathdrive;
```
sherm1: I confused by this code and the comment above it. Why is `swddrive` being used here when `pathdrive` is the only non-empty one? Is there a test case missing for this?

SimTKcommon/src/Pathname.cpp
```diff
@@ -218,13 +241,13 @@ void Pathname::deconstructPathname( const string&   name,
     }
 
     // Now we can put together the canonicalized directory.
-    if (isAbsolutePath) {
-        if (!drive.empty())
-            directory = drive + ":";
+    if (dontApplySearchPath) {
```
sherm1: This seems confusing. `dontApplySearchPath` is an instruction to the caller, but we're not trying to search anything here. What is the real condition here? I think it might still be whether we have an absolute path. Then you can translate into a caller instruction just prior to returning, but this code is already complicated enough that I think the variable names should have the right meanings, even if you have to introduce an extra bool.

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +261,43 @@ void Pathname::deconstructPathname( const string&   name,
...
+//      Leading "../" is interpreted as "./..".
+//      Leading "@/" means relative to executable location.
+// Above leading characters are removed and replaced with the
+// full path name they represent, then the entire path is divided
+// into components. If a component is "." or "" (empty) it is
+// removed. If a component is ".." it and the previous component
+// if any are removed. (".." as a first component will report as
+// ill formed.)
+// If there is something ill-formed about the file name we'll return
+// false.
```
sherm1: Should this internal comment be moved out to the header so it shows in doxygen? It seems like it has some useful info that isn't given elsewhere.

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +261,43 @@ void Pathname::deconstructPathname( const string&   name,
...
+void Pathname::deconstructPathname( const string&   name,
+                                    bool&           isAbsolutePath,
```
sherm1: `dontApplySearchPath`?

SimTKcommon/tests/TestPlugin.cpp
```diff
@@ -190,6 +190,488 @@ void testPathname() {
...
+
+#ifdef _WIN32
+    std::string swd, path, dir, pathname;
+    bool dontApplySearchPath;
+    const std::string cwd = Pathname::getCurrentWorkingDirectory();
+    std::string cwd_nodrive = cwd; cwd_nodrive.erase(0,3);
+    const std::string cwdX = Pathname::getCurrentWorkingDirectory("x");
+    const std::string cwdY = Pathname::getCurrentWorkingDirectory("y");
+
+    swd = ""; path = "X:/topdir/seconddir/myFileName.ext";
```
sherm1: You could just make some of these `swd`s be strings of blanks and tabs.

SimTKcommon/tests/TestPlugin.cpp
```diff
@@ -190,6 +190,488 @@ void testPathname() {
...
+    dir = "y:" + sep + "specified" + sep + "topdir" + sep + "seconddir" + sep;
+    directory = fileName = extension = "junk"; dontApplySearchPath = false;
+    Pathname::deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension, dontApplySearchPath);
+    SimTK_TEST(directory == dir && fileName == "myFileName" && extension == ".ext" && dontApplySearchPath);
+    pathname = "junk";
+    pathname = Pathname::getAbsolutePathnameUsingSpecifiedWorkingDirectory(swd, path);
+    SimTK_TEST(pathname == dir + "myFileName.ext");
+
+    swd = "Y:specified"; path = "./topdir/seconddir/myFileName.ext";
+    dir = "y:" + sep + "specified" + sep + "topdir" + sep + "seconddir" + sep;
```
sherm1: Will this work on a machine that actually has a "Y:" drive? I would think the cwd for Y: would get prepended in that case.

SimTKcommon/tests/TestPlugin.cpp
```diff
@@ -190,6 +190,488 @@ void testPathname() {
...
+    std::string cwd_nodrive = cwd; cwd_nodrive.erase(0,3);
+    const std::string cwdX = Pathname::getCurrentWorkingDirectory("x");
+    const std::string cwdY = Pathname::getCurrentWorkingDirectory("y");
+
+    swd = ""; path = "X:/topdir/seconddir/myFileName.ext";
+    dir = "x:" + sep + "topdir" + sep + "seconddir" + sep;
+    directory = fileName = extension = "junk"; dontApplySearchPath = false;
+    Pathname::deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension, dontApplySearchPath);
+    SimTK_TEST(directory == dir && fileName == "myFileName" && extension == ".ext" && dontApplySearchPath);
+    pathname = "junk";
```
sherm1: You don't need these `pathname="junk"` lines because pathname is overridden by the next line. That could only fail if the compiler is broken -- it doesn't matter what the routine does.

SimTKcommon/tests/TestPlugin.cpp
sherm1: Might be better to use "C:" and "D:" drives since most Windows machines have both.

sherm1:
Done reviewing. Looks good except for a few things I mentioned.

Also, I realize there is a bug in handling of `..` on machines that have symbolic links. In that case a path like `blah/symlink/../xyz` is not equivalent to `blah/xyz` which is what these functions would return. I don't think we should fix that now but it would be worth adding a `@bug` to the doxygen and filing an issue.

carmichaelong:
Recapping a discussion with @sherm1 (feel free to chime in, @chrisdembia):

The design of this function will be changed from the last one proposed. This function will give back the `directory`, `fileName`, and `ext` separately as before, but it will NOT return a boolean `dontApplySearchPath` and will always return an absolute path through `directory`. We also noted that an empty `swd` is a special case, and so this new function will call on `deconstructPathname()` for those cases, but then return an absolute path always (e.g. `path` = "./dir/f.ext" is equivalent to `path` = "dir/f.ext" with an `swd` is not given).

If the given `path` is completely non-ambiguous (e.g. "/dir/f.ext" on Unix, "C:/dir/f.ext" on Windows), then the `swd` is ignored. In the corner case of Windows if only a drive is missing, "/dir/f.ext", then we use the drive of the `swd` if it's given, otherwise we use the current working directory's (cwd's) drive.

Otherwise, if an `swd` is given but is not an absolute path, then the `swd` is pre-processed to make it an absolute path first (with drive on Windows). The `swd` is then replaces the "./" of the `path` or prepended to `path` if `path` is not root-relative (e.g. "./file.ext" and "dir/file.ext", respectively). The final drive chosen in this case is in order of precedence: swd's drive > path's drive > cwd's drive.

Some examples that lead to drive changes that may be odd: Consider `path` = "C:dir/f.ext". (Note that this the same as "C:./dir/f.ext") and `cwd` = "D:/cwd"
* `swd` = "." = "D:/cwd" --> `finalpath` = "D:/cwd/dir/f.ext"
* `swd` = "D:" = "D./" = "D:/cwd" --> `finalpath` = "D:/cwd/dir/f.ext (note for this line that "D:./" means "D's current working directory"
* `swd = "D:/" --> `finalpath` = "D:/dir/f.ext"

sherm1:
Great summary, Carmichael. One thing I'm not sure about:
> if an swd is given but is not an absolute path, then the swd is pre-processed to make it an absolute path first (with drive on Windows). 

In that case I believe the `path`'s drive specification still wins (which is what you might have meant but it wasn't clear). So let's say 
- the user's current drive is "C:" and C's current working directory is "C:/ctemp"
- there is a "D:" drive also and D's current working directory is "D:/dtemp"
- Then I think: `swd` = "one/two", `path`="D:myfile.ext" --> `finalpath` = "D:/dtemp/one/two/myfile.ext"

Do you agree?

carmichaelong:
I'm okay with this working either way. It seems that in the case you've laid out, there is a distinction here (correct me if I'm wrong):
* `swd` = "one/two", `path`="D:myfile.ext" --> `finalpath` = "D:/dtemp/one/two/myfile.ext"
* `swd` = "./one/two", `path`="D:myfile.ext" --> `finalpath` = "C:/ctemp/one/two/myfile.ext"

Thus, the "./" is very important, as it means the swd will be pre-processed with "./" is present, but it won't be changed to be an absolute path if there is no "./", since technically "./one/two" indicates an absolute path whereas "one/two" does not.

sherm1:
Right, that's exactly what I was thinking for those two cases and for exactly the reason you said -- "./one/two" is really an absolute pathname. So the general rule would be that if an swd begins with a ".", that gets replaced by the current drive's cwd, including the drive name. But any other swd is left alone. 

These are very obscure cases, I think. I expect that 99% of the time a specified swd will be an absolute pathname, with drive. I think that is the intended use case in OpenSim where some low-level file that gets included by an osim file contains a relative pathname. We would set the swd to the absolute directory containing the low-level file. But there will probably be other uses we haven't thought of so it is good to get all this stuff documented and working. Any other opinions?

/cc @aseth1 @aymanhab @chrisdembia

aymanhab:
@sherm The problem is that this PR (and the other issues it refers to) is way removed from anything that OpenSim currently uses and it's not clear how this will be incorporated in our system so the arguments are mostly hypothetical! What I'd like to see, and doesn't belong in this PR, is a design/scheme that describes how these will be used at a high level to handle running OpenSim code  from GUI, or with setup files or from Matlab or Python scripts where the working directory could be arbitrary.

Currently the GUI has a mechanism in place to convert AbsolutePath  to RelativePaths and vice versa, while the API is unaware of these issues..

carmichaelong:
@sherm1, ready to be looked at again.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -126,11 +128,40 @@ class SimTK_SimTKCOMMON_EXPORT Pathname {
...
+    /// always return an absolute path, and no bool dontApplySearchPath is returned.
+    /// Rules:
+    /// - If the swd is empty (after removing whitespace), deconstructPathname()
+    ///   is called, and cwd is prepended if needed to make it an absolute path.
+    /// - Otherwise, we evaluate path relative to the swd. These steps are as follows:
+    /// 1) Preprocess the swd. This means that if the swd is of any form that denotes
+    ///    an absolute path (i.e. "C:/file.ext", "C:file.ext", "./file.ext", "/file.ext")
+    ///    we change the swd to reflect the absolute path (e.g. "./file.ext" may change
+    ///    to "/cwd/file.ext" or "C:/cwdOnC/file.ext").
+    /// 2) If path is a root-relative path name (and on Windows this includes a drive)
```
sherm1: I think (2) actually is the first rule applied. If so should come first in this list.

SimTKcommon/include/SimTKcommon/internal/Pathname.h
```diff
@@ -126,11 +128,40 @@ class SimTK_SimTKCOMMON_EXPORT Pathname {
...
+    ///    to "/cwd/file.ext" or "C:/cwdOnC/file.ext").
+    /// 2) If path is a root-relative path name (and on Windows this includes a drive) 
+    ///    (e.g. /usr/file.ext or c:/documents/file.ext), then swd is ignored, and the
+    ///    absolute path is returned.
+    /// 3) Otherwise, if a path is given relative to a directory that is not the root 
+    ///    (e.g. "./dir/file.ext" or "dir/file.ext"), then the swd is prepended to path.
+    /// 4) To resolve drive ambiguities, if swd provides a drive, it is used. If not, 
+    ///    then the path drive is used. If neither provides a drive, then the current 
+    ///    drive is used.
+    static void deconstructPathnameUsingSpecifiedWorkingDirectory(const std::string& swd,
```
sherm1: The comments for this method are great, Carmichael!

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +240,161 @@ void Pathname::deconstructPathname( const string&   name,
...
+{
+    directory.erase(); fileName.erase(); extension.erase();
+    string pathdrive, swddrive, finaldrive;
+    String processed;
+
+    // Remove all the white space and make all the slashes be forward ones.
+    // (For Windows they'll be changed to backslashes later.)
+    String pathcleaned = String::trimWhiteSpace(path).replaceAllChar('\\', '/');
+    if (pathcleaned.empty())
+        return; // path consisted only of white space
```
sherm1: Do you think this is the right policy for an empty pathname? Would it make more sense to return with directory=processed swd, fileName=extension=""? If it should be changed you could do it now or we could just file an issue and deal with it later.

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +240,161 @@ void Pathname::deconstructPathname( const string&   name,
...
+        swdcleaned += '/';
+    removeDriveInPlace(swdcleaned, swddrive);
+
+    // If swd is empty (or only white space) then use deconstructPathname().
+    if (swdcleaned.empty() && swddrive.empty()) {
+        bool dontApplySearchPath;
+        deconstructPathname(path, dontApplySearchPath, directory, fileName, extension);
+        if (!dontApplySearchPath)
+            directory = getCurrentWorkingDirectory() + directory;
+
```
sherm1: Better to put a `return` here, so you can remove indentation from the rest of the code. It is also much easier to follow that way.

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +240,161 @@ void Pathname::deconstructPathname( const string&   name,
...
+        // Also preprocess the swd if it leads with "/". Just grab current drive letter.
+        else if (swdcleaned.substr(0, 1) == "/" && swddrive.empty()) {
+            swddrive = getCurrentDriveLetter();
+        }
+        // Also preprocess if swd starts with something of the form "C:folder/". Resolve
+        // by finding the current directory of this drive.
+        else if (!swddrive.empty()) {
+            swdcleaned.insert(0, getCurrentWorkingDirectory(swddrive));
+            removeDriveInPlace(swdcleaned, swddrive);
+            swdcleaned.insert(0, "/"); // ensure swd starts with "/" to denote full path
```
sherm1: Won't this make it start with `//`? I think getCurrentWorkingDirectory() always returns an absolute path. Is that wrong?

carmichaelong: getCurrentWorkingDirectory() does return an absolute path, but removeDriveInPlace() does a str.erase(0,2), so it removes the leading "/". I considered just erasing the first two by a direct call instead of using the helper function, but that wouldn't work correctly on not-Windows platforms.

sherm1: > removeDriveInPlace() does a str.erase(0,2) ...

That's erase(startPosition,length), so doesn't it just remove the two characters `C:` from the start of the path?

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +240,161 @@ void Pathname::deconstructPathname( const string&   name,
...
+        // Looks like a relative path name (i.e. pathcleaned starts immediately with
+        // a directory or file name).
+        else {
+            processed = swdcleaned + pathcleaned;
+        }
+
+        // If the pathname in its entirety is just one of these, append 
+        // a slash to avoid special cases below.
+        if (processed == "." || processed == ".." || processed == "@")
+            processed += "/";
```
sherm1: Here you turn "." into "./" but it looks like you handled the "./" case already on line 305. Or can this only happen if swd="." and pathname=""? Please double check to make sure this is OK.

SimTKcommon/src/Pathname.cpp
```diff
@@ -238,6 +240,161 @@ void Pathname::deconstructPathname( const string&   name,
...
+
+        // If the path begins with "../" we'll make it ./../ to simplify handling.
+        if (processed.substr(0, 3) == "../")
+            processed.insert(0, "./");
+
+        /* RESOLVING THE FULL PATH */
+        // Full path is determined except for drive letter. Resolve drive letter with
+        // swd, then path, then current drive.
+        if (processed.substr(0, 1) == "/") {
+            processed.erase(0, 1);
```
sherm1: You are removing a leading slash here. Speculation -- this only works because you made it "//" above per my earlier comment! I think you want to get rid of both changes. If you look at the next case below (for "@/") you can see that one leaves the "/" in place (because getThisExecutableDirectory() starts with a "/" after you remove the drive).

SimTKcommon/tests/TestPlugin.cpp
```diff
@@ -190,6 +190,478 @@ void testPathname() {
...
+    pathname = Pathname::findAbsolutePathnameUsingSpecifiedWorkingDirectory(swd, path);
+    SimTK_TEST(pathname == dir + "myFileName.ext");
+
+    swd = "specified"; path = "topdir/seconddir/myFileName.ext";
+    dir = cwd + "specified" + sep + "topdir" + sep + "seconddir" + sep;
+    directory = fileName = extension = "junk";
+    Pathname::deconstructPathnameUsingSpecifiedWorkingDirectory(swd, path, directory, fileName, extension);
+    SimTK_TEST(directory == dir && fileName == "myFileName" && extension == ".ext");
+    pathname = Pathname::findAbsolutePathnameUsingSpecifiedWorkingDirectory(swd, path);
+    SimTK_TEST(pathname == dir + "myFileName.ext");
```
sherm1: These tests are awesome!

sherm1:
Done reviewing -- looks great, Carmichael. Ready except for the few minor comments above.

SimTKcommon/src/Pathname.cpp
carmichaelong: Looks fine to me since the other way is swd="" and path=".", which would be sent to deconstructPathname().

SimTKcommon/src/Pathname.cpp
carmichaelong: Hm. I keep going back and forth with this. Given how I envisioned the use cases, I didn't think an empty path would be okay. I imagine any use like this should have just used getCurrentWorkingDirectory() anyway? For now I'm going to keep this as is and then add some more comments.

carmichaelong:
Fixed the "//" problems. Also chose to do some rearranging of the logic to move some things that definitely could be called by "deconstructPathname()" earlier. It lengthens the code a little, so let me know if the logic is easier to follow. If not, the logic can be reverted if it's not worth it.

I'm also considering that a bit of code at the end could be removed by just calling on "deconstructPathname()" after figuring out how the paths are combined. Do you think it's worth it to remove some of that copied code?

sherm1:
> I'm also considering that a bit of code at the end could be removed by just calling on "deconstructPathname()" after figuring out how the paths are combined. Do you think it's worth it to remove some of that copied code?

Yes, definitely. In fact there are several cases where you can get rid of a bunch of code by using [`getAbsolutePathname()`](https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/Pathname.h#L151) rather than `deconstructPathname()`. I think it already has the same 6 lines of code you are using.

sherm1:
Oops -- never mind my last comment. I forgot you need the individual segments not just the absolute path.

SimTKcommon/src/Pathname.cpp
```diff
@@ -218,13 +220,187 @@ void Pathname::deconstructPathname( const string&   name,
...
+        deconstructPathname(path, dontApplySearchPath, directory, fileName, extension);
+        if (!dontApplySearchPath) 
+            directory = getCurrentWorkingDirectory() + directory;
+        return;
+    }
+
+    // Check if pathcleaned is a root-relative path. For Windows, this means
+    // that path gave a drive. For not-Windows, no drives will exist.
+    if (pathcleaned.substr(0, 1) == "/") {
+        if (getCurrentDriveLetter().empty() || !pathdrive.empty()) {
```
sherm1: This doesn't make sense to me. The current drive would only be empty on a non-Windows system but then why would pathdrive not be empty?

SimTKcommon/src/Pathname.cpp
```diff
@@ -218,13 +220,187 @@ void Pathname::deconstructPathname( const string&   name,
...
+    String pathcleaned = String::trimWhiteSpace(path).replaceAllChar('\\', '/');
+    if (pathcleaned.empty())
+        return; // path consisted only of white space
+    removeDriveInPlace(pathcleaned, pathdrive);
+
+    String swdcleaned = String::trimWhiteSpace(swd).replaceAllChar('\\', '/');
+    
+    // If swd was empty, then just call deconstructPathname() on the path.
+    if (swdcleaned.empty()) {
+        bool dontApplySearchPath;
```
sherm1: Rather than repeating this code, how about adding a new method `deconstructAbsolutePathname()` that does it? Then replace `getAbsolutePathname()` with:
```c++
static std::string getAbsolutePathname(const std::string& pathName) {
    std::string directory, fileName, extension;
    desconstructAbsolutePathname(pathName, directory, fileName, extension);
    return directory + fileName + extension;
}

SimTKcommon/src/Pathname.cpp
sherm1: Currently I think there is no directly platform-dependent code in this method. Instead everything is hidden by platform-dependent methods that understand how to interpret the path characters so that "C:" just comes in as ordinary file characters on Linux. It would be good to keep it that way if possible. Testing whether `getCurrentDriveLetter().empty()` is really a hidden platform dependency; those should be done with `#ifdef WIN32` and localized with platform dependencies elsewhere. In contrast testing `pathdrive.empty()` is not a platform dependency since it could be empty anywhere.

sherm1:
I didn't notice any test cases with ".." segments in the pathnames. It would be good to have one or two of those just to make sure that features is still working right in the new methods.

carmichaelong:
@sherm1, ready again. Has the changes you suggested and some directed tests at weirder cases.

sherm1:
Sweet! I love it when the code gets shorter! Thanks so much for all this work, Carmichael. Merging now.

Issue #246, opened 2014-09-29
CMAES Optimizer: initial inclusion, threading, MPI, resuming an optimization


chrisdembia:
Sherm just got news that cmaes will be released under the Apache 2.0 license, so we will be able to include it directly in Simbody. I am willing to take the brunt of the work doing the integration, but will want help from @sherm1 and, if interested, @carmichaelong and @msdemers.

First question:

Since CMAES is a derivative-free optimizer, when someone creates an OptimizerSystem and they try to define a derivative function, what do we do?

Things to keep in mind:
- [x] How to handle constraint tolerance?
- [x] Satisfying license requirements?
- ~~[ ] Do all OptimizerRep subclasses need to override clone()? There's a definition in OptimizerRep.~~
- [x] cmaes outputs to the console; should we allow this?
- [x] Does hitting the maximum number of iterations count as finding a solution?
- [ ] Check for memory leaks.
- [x] Create simple CMAES example file.
- [ ] Create example with an optimization, optimizing something about a model's behavior, using visualization to show the improvement with iterations of the optimization.
- [ ] Allow restarts.
- ~~[ ] Use cmaes' boundary transformation?~~
- [ ] Allow a single parameter, but lambda >= 2 (change [this line](https://github.com/simbody/simbody/blob/master/SimTKmath/Optimizers/src/CMAESOptimizer.cpp#L51))
- [ ] MPI: should the master run objective function evaluations? Make this an option. If master cannot run obj func, then having only one process will cause the solver to hang; prevent option=off with nproc = 1.
- [ ] Make sure CMAES is well behaved with MPI when only using 1 proc.
- [ ] Test that a diagnostics level of 2 causes the allcmaes.dat file to be written.
- [ ] Make sure I only use "popsize" and "stepsize", never "lambda" or "sigma".
- [ ] In readme, clarify how to find the MPI library with cmake.
- [ ] SIMBODY_ENABLE_MPI - > SIMBODY_MPI.
- [ ] Better MPI test: maybe actually run an executable using the mpi executable.
- [ ] In SimbodyConfig.cmake, provide if mpi was used and which library was used. (use components?)
- [ ] Diagnostics should print numbers using scientific notation.
- [ ] Show how to use restarts in CMAESOptimization.cpp
- [ ] processing cmaes errors (c-cmaes calls exit()?).
- [ ] Allow turning off parallel.
- [ ] Add MPI test to travis.
[...]

PR #316, opened 2014-12-24
[cmaes] Initial work on allowing use of MPI

chrisdembia:
The use of MPI with CMAES is now functional, though it is definitely not complete. I have a long todo list over at #246 (only some of those items are related to MPI). Note that this PR goes into cmaes-feature, not master.

I've created a CMAESMPITest, which runs CMAESTest but with MPI. I've added another travis build that uses MPI and runs CMAESMPITest. 

The test should fail right now. The error is from exceptions that are thrown by the master process. I don't want to address this failure here though; I'd like to discuss it at @246. I might actually like to exclude fixing those errors from this PR; I could comment-out the two subtests that are failing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/316)


chrisdembia:
VTK uses MPI; maybe I'll try to follow their example. http://stackoverflow.com/questions/17729967/compiling-mpi-vtk-hello-world. They seem to have created their own MPI interface. Even though that might be a lot of work, that is desirable since then the user does not need to include MPI themselves.

chrisdembia:
VTK actually is able to create a unified interface for MPI and threading.

sherm1:
Really? That’s very nice.

adamkewley:

PR #321, opened 2014-12-31
Fix Simbody 3.5 install problem on Mac.

sherm1:
This addresses osrf/homebrew-simulation#33. 

The Mac installation failed due to inability to copy a symbolic link. The code that failed was over a year old so we think this may be due to a change in CMake behavior (seen with CMake 3.1). The fix here pushes the symbolic link into a subdirectory and then copies the contents of the whole subdirectory rather than trying to copy the link directly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,3 +58,6 @@ install:
 script:
   ## Test Simbody.
   - ctest -j8 --output-on-failure
+
+  ## Make sure we can install with no issues.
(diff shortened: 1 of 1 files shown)


sherm1:
Simbody 3.5.1 release is now available with this fix in it. And @scpeters updated the osrf simbody homebrew formula to grab that, so all should be well now. Thanks, Steve and Chris!

PR #322, opened 2014-12-31
[travis] Test that simbody installs.

chrisdembia:


sherm1:
Both fixes are in the cherry-picks that I used for PR #323, ready for review.

PR #327, opened 2015-01-20
Add copy-on-write smart pointer CloneOnWritePtr

sherm1:
Simbody already has a `ClonePtr` which is like `std::unique_ptr` but with deep copy semantics (`std::unique_ptr` doesn't allow copying). This PR adds another smart pointer `CloneOnWritePtr` that acts like `std::shared_ptr` when the managed object is being shared for reading, but acts like `ClonePtr` when someone tries to write on it. That means that it is very fast to copy construct or copy assign, because no copy actually occurs until someone tries to write. At that point the managed object's `clone()` method gets called and the writer ends up writing on their own private copy, while all the other sharers see the use count decremented.

This should allow for very fast copying of complicated objects like States where large chunks of the object remain unchanged from source to copy. In particular, discrete state variables are unchanged while integrators need to make trial copies of the continuous ones.

Note: I modeled this as closely as possible on C++11 smart pointers rather than on the existing `ClonePtr` which needs to be upgraded to match C++11 better.

### DIFF
--- a/SimTKcommon/tests/StateTest.cpp
+++ b/SimTKcommon/tests/StateTest.cpp
@@ -38,66 +38,6 @@ using std::endl;
 using namespace SimTK;
 
 
-
-
-//void testLowestModified() {
-//    const SubsystemIndex Sub0(0), Sub1(1);
-//    State s;
-//    s.setNumSubsystems(2);
-//    SimTK_TEST(s.getSystemStage()==Stage::Empty);
-//    SimTK_TEST(s.getSubsystemStage(Sub0)==Stage::Empty && s.getSubsystemStage(Sub1)==Stage::Empty);
-//    SimTK_TEST(s.getLowestStageModified()==Stage::Topology);
-//
-//    const DiscreteVariableIndex dvxModel = s.allocateDiscreteVariable(Sub1, Stage::Model, new Value<Real>(2));
-//
-//    // "realize" Topology stage
-//    s.advanceSubsystemToStage(Sub0, Stage::Topology);
-//    s.advanceSubsystemToStage(Sub1, Stage::Topology);
-//    s.advanceSystemToStage(Stage::Topology);
-//    SimTK_TEST(s.getLowestStageModified()==Stage::Topology);    // shouldn't have changed
-//
-//    const DiscreteVariableIndex dvxInstance = s.allocateDiscreteVariable(Sub0, Stage::Instance, new Value<int>(-4));
-//
-//    // A Model-stage variable must be allocated before Topology is realized, and this condition
-//    // should be tested even in Release mode.
-//    try {
-//        s.allocateDiscreteVariable(Sub0, Stage::Model, new Value<int>(0));
-//        SimTK_TEST(!"Shouldn't have allowed allocation of Model-stage var here");
-//    } catch (...) {}
-//
-//    // "realize" Model stage
-//    s.advanceSubsystemToStage(Sub0, Stage::Model);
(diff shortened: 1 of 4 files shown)


chrisdembia:
What happens if someone gets the underlying pointer using `get()` then const-casts and writes to that pointer? [The OpenSim code contains about 49 const_casts](https://github.com/opensim-org/opensim-core/search?p=1&q=const_cast&utf8=%E2%9C%93). I didn't search exhaustively, but [here](https://github.com/opensim-org/opensim-core/blob/2c0d85e05493d2fa68325e9e24456298a8602817/OpenSim/Simulation/Model/CMCActuatorSubsystem.cpp#L129)'s a const-cast on a state object.

SimTKcommon/tests/StateTest.cpp
```diff
@@ -250,7 +190,6 @@ int main() {
 
 
     SimTK_START_TEST("StateTest");
-        //SimTK_SUBTEST(testLowestModified);
```
chrisdembia: What is the reason for removing these tests?

sherm1: I didn't remove them. They were already commented out because they were tests for long-ago dead code.

sherm1:
> What happens if someone gets the underlying pointer using get() then const-casts and writes to that pointer?

In that case they would be modifying the shared object rather than writing to a unique one. That is a very bad thing to do!

sherm1:
>  The OpenSim code contains about 49 const_casts. I didn't search exhaustively, but here's a const-cast on a state object.

Probably some of those are OK and some aren't. But they wouldn't affect the copy-on-write activity of the `CloneOnWritePtr` class. To mess that up, you actually have to extract a const pointer from it and then const_cast it. That would be a foolish thing to do to an object whose whole purpose in life is to use read-vs-write as a way to avoid copying.

chrisdembia:
Something I had tried to discern with `ClonePtr` was whether it was @sherm1's invention, or it was a common type of smart pointer. Could you perhaps clarify in documentation if such pointers are general ideas or are just in Simbody?

chrisdembia:
> In that case they would be modifying the shared object rather than writing to a unique one. That is a very bad thing to do!

I guess I could see it happening, perhaps by someone like me.

The documentation says the copy occurs upon write, but does it more accurately occur upon calling `upd()`? I quickly tried peeking at the code to discern this; perhaps I am wrong.

chrisdembia:
Don't wait for me to "approve" this PR to merge it.

sherm1:
> The documentation says the copy occurs upon write, but does it more accurately occur upon calling upd()? 

The copy occurs whenever an attempt is made to get writable access to the contents of the smart pointer. upd() is an example, but so are updRef(), operator->() and operator*(), and release().

sherm1:
> Could you perhaps clarify in documentation if such pointers are general ideas or are just in Simbody?

The concepts are nothing new. Copy-on-write (COW) is a venerable computer science design pattern. The idea of using a clone() method is the standard approach to the lack of virtual constructors in C++. [Here](http://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-on-write) is a limited-functionality COW pointer implementation. You can see the family resemblance. That one won't work with abstract classes; combining it with the clone() requirement makes it more useful. The rest of the implementation (such as the names of methods and the use of templates to allow COW to work across related classes) is just copied from the C++11 standard implementations of std::unique_ptr and std::shared_ptr.

sherm1:
> Don't wait for me to "approve" this PR to merge it.

OK. Thanks for looking at it. I hope you enjoyed it!

PR #330, opened 2015-01-21
Mingw

xantares:


CMakeLists.txt
```diff
@@ -451,25 +451,26 @@ IF(BUILD_USING_OTHER_LAPACK)
...
-        SET(MATH_LIBS_TO_USE_VN ${LAPACK_BEING_USED} pthreadVC2_x64)
+        SET(CMAKE_THREAD_LIBS_INIT pthreadVC2_x64)
     else()
-        SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} pthreadVC2)
-        SET(MATH_LIBS_TO_USE_VN ${LAPACK_BEING_USED} pthreadVC2)
+        SET(CMAKE_THREAD_LIBS_INIT pthreadVC2)
     endif()
 ENDIF()
 
+SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} ${CMAKE_THREAD_LIBS_INIT} ${REALTIME_LIB} ${DL_LIBRARY} ${MATH_LIBRARY})
```
chrisdembia: Can you make this line under 80 characaters?

CMakeLists.txt
```diff
@@ -451,25 +451,26 @@ IF(BUILD_USING_OTHER_LAPACK)
...
-    SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} pthread ${REALTIME_LIB} dl m)
-    SET(MATH_LIBS_TO_USE_VN ${LAPACK_BEING_USED} pthread ${REALTIME_LIB} dl m)
-ELSE()
+    SET(MATH_LIBRARY m)
+    SET(DL_LIBRARY dl)
+    SET(CMAKE_THREAD_LIBS_INIT pthread)
+ENDIF()
+
+IF(MSVC)
     ## Assume Microsoft Visual Studio
```
chrisdembia: You can remove this comment now.

CMakeLists.txt
```diff
@@ -451,25 +451,26 @@ IF(BUILD_USING_OTHER_LAPACK)
 "Basename of the actual Lapack library we're depending on; can't change here; see variable BUILD_USING_OTHER_LAPACK." FORCE)
 ENDIF(BUILD_USING_OTHER_LAPACK)
 
-IF(${CMAKE_C_COMPILER} MATCHES "cc" OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
-    IF(APPLE)
-        SET(REALTIME_LIB)
-    ELSE()
+IF(UNIX)
```
chrisdembia: +1

CMakeLists.txt
```diff
@@ -451,25 +451,26 @@ IF(BUILD_USING_OTHER_LAPACK)
...
+        SET(CMAKE_THREAD_LIBS_INIT pthreadVC2_x64)
     else()
-        SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} pthreadVC2)
-        SET(MATH_LIBS_TO_USE_VN ${LAPACK_BEING_USED} pthreadVC2)
+        SET(CMAKE_THREAD_LIBS_INIT pthreadVC2)
     endif()
 ENDIF()
 
+SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} ${CMAKE_THREAD_LIBS_INIT} ${REALTIME_LIB} ${DL_LIBRARY} ${MATH_LIBRARY})
+SET(MATH_LIBS_TO_USE_VN ${MATH_LIBS_TO_USE})
```
chrisdembia: +1

CMakeLists.txt
```diff
@@ -451,25 +451,26 @@ IF(BUILD_USING_OTHER_LAPACK)
...
+IF(UNIX)
+    IF(NOT APPLE)
         SET(REALTIME_LIB rt)
     ENDIF()
-    SET(MATH_LIBS_TO_USE    ${LAPACK_BEING_USED} pthread ${REALTIME_LIB} dl m)
-    SET(MATH_LIBS_TO_USE_VN ${LAPACK_BEING_USED} pthread ${REALTIME_LIB} dl m)
-ELSE()
+    SET(MATH_LIBRARY m)
+    SET(DL_LIBRARY dl)
+    SET(CMAKE_THREAD_LIBS_INIT pthread)
```
chrisdembia: Nitpicky, but why name this differently? It's not a cmake variable.

chrisdembia:
Can you add a comment in this section of the README [here](https://github.com/simbody/simbody/blob/master/README.md#installing), similar to the OSX comment, saying that on Windows you can also compile with mingw? Or should we leave this as an "undocumented" feature?

SimTKcommon/Geometry/src/DecorativeGeometry.cpp
```diff
@@ -34,17 +34,17 @@ namespace SimTK {
...
+const Vec3 Gray    = Vec3(.5,.5,.5);
+const Vec3 Red     = Vec3( 1, 0, 0);
+const Vec3 Green   = Vec3( 0, 1, 0);
+const Vec3 Blue    = Vec3( 0, 0, 1);
+const Vec3 Yellow  = Vec3( 1, 1, 0);
+const Vec3 Orange  = Vec3( 1,.5, 0);
+const Vec3 Magenta = Vec3( 1, 0, 1);
+const Vec3 Purple  = Vec3(.5, 0,.5);
+const Vec3 Cyan    = Vec3( 0, 1, 1);
+const Vec3 White   = Vec3( 1, 1, 1);
```
sherm1: How do these get exported from the DLL on Windows without `__declspec(dllexport)`? Great if they do but at one time they were undefined symbols without that macro there.

chrisdembia: The appveyor test passes, and I believe at least some of those symbols are used in test cases and examples. So perhaps they are not needed?

sherm1: Yeah -- could have been a bug in an older version of VC++. I see that the header files still have the EXPORT macro, and that should be all that is needed.

@xantares, were the duplicate EXPORT macros causing a failure when compiling with mingw?

sherm1:
Please let me know when this is ready for final review and merge.

CMakeLists.txt
xantares: ok

CMakeLists.txt
xantares: yep

CMakeLists.txt
xantares: usually it comes from findthreads.cmake

SimTKcommon/Geometry/src/DecorativeGeometry.cpp
xantares: yes, the export should only be in the header, never the cpp

xantares:
I dont understand why the  AppVeyor build fails.I did not touch LAPACK_BEING_USED since a942056.

Issue #331, opened 2015-01-22
undefined references to ~PIMPLHandle()


xantares:
undefined refs to ~PIMPLHandle, which is declared protected (and not virtual) in 
SimTKcommon/include/SimTKcommon/internal/PrivateImplementation.h
inherited here:
../SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h
../SimTKcommon/include/SimTKcommon/internal/PrivateImplementation.h
../SimTKcommon/include/SimTKcommon/internal/PrivateImplementation_Defs.h
../SimTKcommon/include/SimTKcommon/internal/ParallelWorkQueue.h
../SimTKcommon/include/SimTKcommon/internal/Parallel2DExecutor.h
../SimTKcommon/include/SimTKcommon/internal/ParallelExecutor.h
and instanciated in SimTKcommon/src/PrivateInstantiations.cpp.

Linking CXX shared library ../../libSimTKmath.dll
CMakeFiles/SimTKmath.dir/objects.a(CMAESOptimizer.cpp.obj): In function `~ParallelExecutor':
/home/xantares/projects/simbody/SimTKcommon/./include/SimTKcommon/internal/ParallelExecutor.h:65: undefined reference to `SimTK::PIMPLHandle<SimTK::ParallelExecutor, SimTK::ParallelExecutorImpl, false>::~PIMPLHandle()'
/home/xantares/projects/simbody/SimTKcommon/./include/SimTKcommon/internal/ParallelExecutor.h:65: undefined reference to `SimTK::PIMPLHandle<SimTK::ParallelExecutor, SimTK::ParallelExecutorImpl, false>::~PIMPLHandle()'
CMakeFiles/SimTKmath.dir/objects.a(ContactGeometry_TriangleMesh.cpp.obj): In function `ZNK5SimTK15ContactGeometry12TriangleMesh19createPolygonalMeshEv':
/home/xantares/projects/simbody/SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h:76: undefined reference to `SimTK::PIMPLHandle<SimTK::PolygonalMesh, SimTK::PolygonalMeshImpl, true>::~PIMPLHandle()'
CMakeFiles/SimTKmath.dir/objects.a(ContactGeometry_TriangleMesh.cpp.obj): In function `~PolygonalMesh':
[...]

PR #332, opened 2015-01-22
Mingw2

xantares:


xantares:
removes a lot of undefined issues, but this one remains:
https://github.com/simbody/simbody/issues/331

sherm1:
Thanks for this change -- I'll merge it since it seems to be an improvement even though it doesn't get you all the way past your mingw troubles. (I tried it on Windows 8.1 with VS 2013 and it works fine there.)

What's your name? I'd like to add you to the contributor list with your real name unless you prefer anonymity. 

Sherm

PR #338, opened 2015-02-06
Improve the operative system detection in preprocessor code

j-rivero:
For some reason (my hypothesis is the use of [\__linux keyword instead of the POSIX \__linux\__](http://sourceforge.net/p/predef/wiki/OperatingSystems/))  I'm receiving [failures of Ubuntu Simbody](
https://launchpad.net/ubuntu/+source/simbody/3.5.1+dfsg-1~exp1~vivid1/+build/6951817/+files/buildlog_ubuntu-vivid-powerpc.simbody_3.5.1%2Bdfsg-1%7Eexp1%7Evivid1_FAILEDTOBUILD.txt.gz) when is compiled in the Linux/PPC architecture (it fails looking for the windows.h header).

This pull request implement a small variation in the logic of operative system checking and fix the linux keyword according to POSIX.

Untested on PPC but should be enough.

### DIFF
--- a/SimTKcommon/src/ParallelExecutor.cpp
+++ b/SimTKcommon/src/ParallelExecutor.cpp
@@ -166,13 +166,13 @@ void ParallelExecutor::execute(Task& task, int times) {
 #ifdef __APPLE__
    #include <sys/sysctl.h>
    #include <dlfcn.h>
+#elif _WIN32
+   #include <windows.h>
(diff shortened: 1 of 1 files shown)


sherm1:
Looks good. I verified that both gcc and clang define both `__linux` and `__linux__` on Ubuntu so that would explain why it has been working there.

PR #339, opened 2015-02-09
SimbodyConfig.cmake allows relocation of installation.

chrisdembia:
Previously, SimbodyConfig.cmake contained hardcoded paths to the installation
directory. This is fine on Linux, but may be an issue on Windows where users
choose the installation directory long after the build process.

To do this, I used CMake module that comes with CMake 2.8.8. We currently only
support up through CMake 2.8.6. Since Ubuntu 12.04 uses 2.8.6 and it is likely
that our users will be running Ubuntu 12.04, we cannot use CMake 2.8.8 yet.
Instead, I copied the necessary CMake module from 2.8.8 and committed it to the
repo.

I tested this on Linux by compiling a Simbody-using client project, and by
changing the name of the install directory after the installation was complete.
In both cases, my client project was able to correctly find Simbody. I do not
think a Windows test is necessary, as I did not change any platform specific
behavior.

@scpeters @j-rivero could you take a peek? Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Dependencies
 
 Simbody depends on the following:
 
-* cross-platform building: [CMake](http://www.cmake.org/cmake/resources/software.html) 2.8.6 or later
+* cross-platform building: [CMake](http://www.cmake.org/cmake/resources/software.html) 2.8.8 or later
 * compiler: [Visual Studio](http://www.visualstudio.com) 2013 or later (Windows only), [gcc](http://gcc.gnu.org/) 4.8.1 or later (typically on Linux), or [Clang](http://clang.llvm.org/) 3.4 or later (typically on Mac, possibly through Xcode)
 * linear algebra: [LAPACK](http://www.netlib.org/lapack/) and [BLAS](http://www.netlib.org/blas/)
 * visualization (optional): [FreeGLUT](http://freeglut.sourceforge.net/), [Xi and Xmu](http://www.x.org/wiki/)
@@ -143,7 +143,7 @@ Windows using Visual Studio
 All needed library dependencies are provided with the Simbody installation on Windows, including linear algebra and visualization dependencies. 
 
 1. Download and install [Microsoft Visual Studio](http://www.visualstudio.com), version 2013 or higher. The "Community Edition" is free for "non-enterprise" use. The "Express" edition is another free option, in that case use *Visual Studio Express for Windows Desktop*.
-2. Download and install [CMake](http://www.cmake.org/download), version 2.8.6 or higher.
+2. Download and install [CMake](http://www.cmake.org/download), version 2.8.8 or higher.
 3. (optional) If you want to build API documentation, download and install Doxygen, version 1.8.8 or higher.
 
 #### Download the Simbody source code
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ framework. Mac's come with the visualization dependencies.
 
 On Ubuntu, we need to get the dependencies ourselves. Open a terminal and run the following commands.
 
-1. Get the necessary dependencies: `$ sudo apt-get install cmake liblapack-dev`
+1. Get the necessary dependencies: `$ sudo apt-get install cmake liblapack-dev`. The cmake on Ubuntu 12.04 is not new enough; you could instead download it from [cmake.org](http://www.cmake.org/download/) or use [this third party PPA](https://launchpad.net/~robotology/+archive/ubuntu/ppa).
 2. If you want to use the CMake GUI, install `cmake-qt-gui`.
 3. For visualization (optional): `$ sudo apt-get install freeglut3-dev libxi-dev libxmu-dev`
 4. For API documentation (optional): `$ sudo apt-get install doxygen`
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,11 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then sudo apt-get update -qq; fi
   - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then sudo apt-get install -qq g++-4.8; fi
   - if [[ "$TRAVIS_OS_NAME" = "linux" && "$CXX" = "g++" ]]; then export CXX="g++-4.8" CC="gcc-4.8"; fi
+  
+  # CMake 2.8.11 on Ubuntu.
+  - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then sudo add-apt-repository ppa:kalakris/cmake -y; fi
+  - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then sudo apt-get update -qq; fi
+  - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then sudo apt-get install cmake; fi
 
 install:
   - mkdir ~/simbody-build && cd ~/simbody-build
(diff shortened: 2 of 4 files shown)


scpeters:
You said you copied the cmake file into the repo, but it still failed with travis (12.04).

chrisdembia:
> You said you copied the cmake file into the repo, but it still failed with travis (12.04).

Ah oops.

j-rivero:
As far as the cmake module works in the same way on Linux, I'm fine with the change, looks to me like it is the right way of getting this new feature.

chrisdembia:
thanks @scpeters; fixed!

chrisdembia:
> As far as the cmake module works in the same way on Linux

Yes it does :) Well, I did a simple test.

sherm1:
I have a few questions/worries about this:
- At some point we baked-in the installation directory into the simbody-visualizer search path at the time Simbody is built. How does this change interact with that one?
- You initially described this as a fix for Windows but then said it doesn't need Windows testing?
- Copying a 2.8.8 module seems like a temporary fix that could cause trouble later if the old copy were used with a newer CMake. Is it only being used for 2.8.6?

I think I might be missing the big picture of this PR. What problem is it solving?

scpeters:
For reference, [here's the code](https://github.com/simbody/simbody/blob/1cd36b037f756929344514686c96862d6c93354d/Simbody/Visualizer/src/VisualizerProtocol.cpp#L231-L273) that searches for the visualizer executable.

chrisdembia:
> What problem is it solving?

The problem is that if I build and install Simbody with CMAKE_INSTALL_PREFIX=abc and then I later move the Simbody installation to /def, the SimbodyConfig.cmake will no longer be useable and one will not be able to build a client project using SimbodyConfig.cmake (unless they edit it). This is because `abc` will be hardcoded into SimbodyConfig.cmake.

This is a problem on all platforms, but it is much more likely on Windows than on other platforms for the user to move the Simbody installation to a place that is not `CMAKE_INSTALL_PREFIX`. The documentation of [CMakePackageConfigHelpers](It helps making the resulting package relocatable by avoiding hardcoded paths in the installed Config.cmake file.) says "It helps making the resulting package relocatable by avoiding hardcoded paths in the installed Config.cmake file.". This is really only a problem for projects that have a binary release, and so it's really only a problem for OpenSim. However, since I plan to do it for OpenSim, I thought I might as well do it for Simbody, and might as well do it for Simbody first.


> At some point we baked-in the installation directory into the simbody-visualizer search path at the time Simbody is built. How does this change interact with that one?

Thanks @scpeters for the reference. This change does not affect the baking-in of CMAKE_INSTALL_PREFIX. This change makes the Simbody installation more flexible (by being relocatable) in that someone can now build a client project after Simbody has been relocated. However, if you relocate Simbody, you will lose the benefit of the baked-in CMAKE_INSTALL_PREFIX and will rely on the rest of the visualizer search path to find the executable. I think further enhancements of the baking-in are beyond the scope of this PR, considering that all previous functionality is intact. I'll also note that when people inevitably choose custom install directories for OpenSim's release, the baking-in will not function (regardless of this PR).


> You initially described this as a fix for Windows but then said it doesn't need Windows testing?

Yes...the problem is platform-indepenedent, it's the culture of Windows that causes the problem to be more relevant on Windows. I should have said this more clearly at first.

> Copying a 2.8.8 module seems like a temporary fix that could cause trouble later if the old copy were used with a newer CMake.

I agree. If this gets merged I'll create an issue to delete the CMakePackageConfigHelpers.cmake file when we move to 2.8.8 or higher.

> Is it only being used for 2.8.6?

Yes. The alternative solution is to upgrade to 2.8.8; I don't like this because I think we should still support Ubuntu 12.04 (for travis and for people using Simbody on a computing cluster). I also want to make this change before the OpenSim 4.0 release, because otherwise the OpenSimConfig.cmake I created for OpenSim is largely useless unless this change is made. And I think that we will still want to support Ubuntu 12.04 when OpenSim 4.0 is released. If others think we should upgrade to a new cmake, then I'm fine with not including CMakePackageConfigHelpers.cmake in our repo; I'd rather not anyway.

sherm1:
Thanks for the explanation, Chris. With the existing implementation, does it work to move the installation and then set SIMBODY_HOME? That was supposed to be the other way to be able to install Simbody anywhere but I haven't tried it lately with a client program.

sherm1:
> If others think we should upgrade to a new cmake, then I'm fine with not including CMakePackageConfigHelpers.cmake in our repo; I'd rather not anyway.

If it isn't too hard to install CMake 2.8.8 or later on Ubuntu 12.04 then I would prefer just to tell people to do that; I don't really like having the module in the repo.

chrisdembia:
> Thanks for the explanation, Chris. With the existing implementation, does it work to move the installation and then set SIMBODY_HOME?

Nope :/

That was supposed to be the other way to be able to install Simbody anywhere but I haven't tried it lately with a client program.

Setting SIMBODY_HOME is still required with or without this PR. I'm not sure I completely understand your question, but this PR fixes paths that are within SimbodyConfig.cmake, not the mechanism for *finding* the installation.

Here's what a resulting SimbodyConfig.cmake looks like now:

```
list(APPEND Simbody_ROOT_DIR 
            "/home/fitze/simtk/simbody")

list(APPEND Simbody_INCLUDE_DIR 
            "/home/fitze/simtk/simbody/include/simbody")

list(APPEND Simbody_LIB_DIR 
            "/home/fitze/simtk/simbody/lib/x86_64-linux-gnu")

list(APPEND Simbody_BIN_DIR 
            "/home/fitze/simtk/simbody/libexec/simbody"
```

The proposed change would remove those absolute paths.

> If it isn't too hard to install CMake 2.8.8 or later on Ubuntu 12.04 then I would prefer just to tell people to do that; I don't really like having the module in the repo.

K I'll address this.

chrisdembia:
Ready for review.

README.md
```diff
@@ -143,7 +143,7 @@ Windows using Visual Studio
 All needed library dependencies are provided with the Simbody installation on Windows, including linear algebra and visualization dependencies. 
 
 1. Download and install [Microsoft Visual Studio](http://www.visualstudio.com), version 2013 or higher. The "Community Edition" is free for "non-enterprise" use. The "Express" edition is another free option, in that case use *Visual Studio Express for Windows Desktop*.
-2. Download and install [CMake](http://www.cmake.org/download), version 2.8.6 or higher.
+2. Download and install [CMake](http://www.cmake.org/download), version 2.8.8 or higher.
```
scpeters: For Ubuntu 12.04, you can recommend a PPA, which makes it easier to install custom packages. [Here's one](https://launchpad.net/~robotology/+archive/ubuntu/ppa) for example.

chrisdembia: Good idea; I did this now.

sherm1:
I tried this on Windows. It works and I get it now! Thanks, Chris.

chrisdembia:
@sherm1 did you try moving the installation after build? Thanks for reviewing :)

sherm1:
>did you try moving the installation after build?

Yes. I was using an independent main program built with our sample CMakeLists.txt file. To get that to work with a random installation directory, I had to provide a value for the `HINTS` argument to `find_package`. That worked but wouldn't have before you made the change. 

On Windows I also had to set the PATH variable to include `randomInstallationDirectory/bin` so it could find the dlls. It couldn't find the simbody-visualizer (since the built-in path was wrong) but setting `SIMBODY_HOME` to the new installation directory fixed that. So it wasn't exactly easy to do but at least it is possible now!

Issue #263, opened 2014-11-07
Get snapshot of current active contacts


texasflood:
There should be some way of getting information of what contact surface are currently in contact. My thoughts on how to implement this would be to have a function like this:

<code>bool isThereContact (State state, ContactTrackerSubsystem contactSystem, MobilizedBody mobod1, ContactSurface surface1, MobilizedBody mobod2, ContactSurface surface2);</code>

which tells you whether the given unique pair of the contact surface associated with a particular <code>MobilizedBody</code> are currently in contact. As multiple Mobods can use the same <code>ContactSurface</code>, both must be passed to the function.

This would be very helpful to implement some control algorithms which change state based on contacts.

sherm1:
Hi, Anas. PR #347 added a method that maps `(MobilizedBody,contactSurfaceNumber)` to `ContactSurfaceIndex`. There is already a method `ContactSnapshot::hasContact(ContactSurfaceIndex,ContactSurfaceIndex)` that tells you whether a particular pair of surfaces are contacting. So this should now let you do what you wanted to do in the first place. This change is currently in the master branch if you want to try it and will be released in Simbody 3.6.

Thanks for pointing out this problem.

Regards,
Sherm

/cc @texasflood, @hsu

PR #346, opened 2015-02-21
Added index-on-body and userRef pointer to ContactSurface

### LINKED ISSUE #98: Contact geometry needs a void* pointer for identifying corresponding model component
Similar to the pointer stored with DecorativeGeometry. (This is useful for mapping back from contact or ray-contact geometry intersection to the higher level model component it was generated from.) Gazebo uses this for Lidar modeling.

sherm1:
Also makes `Body::addContactSurface()` set the index automatically. Fixes #98.
Partially addresses #263.

### DIFF
--- a/Simbody/include/simbody/internal/Body.h
+++ b/Simbody/include/simbody/internal/Body.h
@@ -114,11 +114,13 @@ writable access to decorations even on a const Body -- these are after all
 just decorations. **/
 DecorativeGeometry& updDecoration(int i) const;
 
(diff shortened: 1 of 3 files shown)


sherm1:
John, in case you want to review this before it goes in I'm cc'ing you. I'm not sure if this will actually help in Gazebo but it could make it easier to map back from contacts to the corresponding Gazebo geometry. This PR allows for an arbitrary `void*` to be stored with a `ContactSurface`.

/cc @hsu

PR #347, opened 2015-02-27
Find ContactSurfaceIndex of surface 

### LINKED ISSUE #263: Get snapshot of current active contacts

sherm1:
Previously there was no good way to map between a system's `ContactSurface` objects and the `ContactSurfaceIndex` assigned by the `ContactTrackerSubsystem`. The PR adds a method `ContactTrackerSubsystem::getContactSurfaceIndex(MobilizedBody, int contactSurfaceOrdinal)` where the ordinal is the small integrator returned by `Body::addContactSurface()`. With the `ContactSurfaceIndex` in hand it is possible to ask whether particular surfaces are currently in contact, so this fixes #263.

This PR also cleans up some doxygen and fixes mobilizer-associated inboard and outboard decorations to record their ordinal numbers (obscure).

### DIFF
--- a/Simbody/include/simbody/internal/Body.h
+++ b/Simbody/include/simbody/internal/Body.h
@@ -84,7 +84,7 @@ const MassProperties& getDefaultRigidBodyMassProperties() const;
 
 /** Add a piece of decorative geometry fixed at some pose on this Body. 
 This can be used for visualization of the Body's motion. Returns a small 
-integer that can be used to identify this decoration in any copy of this
+integer ordinal that can be used to identify this decoration in any copy of this
 %Body. The supplied DecorativeGeometry object is copied and transformed by the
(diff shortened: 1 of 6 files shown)


texasflood:
Great thanks, I'll try it out!

PR #354, opened 2015-03-27
Decorative geom opensim40

aymanhab:
This commit adds DecorativeTorus and DecorativeArrow to the set of primitives that can be handled by the Visualizer. As of now the built in visualizer ignores these primitives. The OpenSim GUI/Visualizer which uses the same mechanism/primitives will eventually use these to display forces, torques and WrapObjects.


SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +596,79 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
 };
 
 
+/** This defines a displayable torus  . **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeTorus : public DecorativeGeometry {
+public:
+    explicit DecorativeTorus(Real inner=0.1, Real outer=0.5);
```
sherm1: For consistency, this Torus definition should match Simbody's existing `ContactGeometry::Torus` object that Ian wrote ([click here to see definition](https://github.com/simbody/simbody/blob/master/SimTKmath/Geometry/include/simmath/internal/ContactGeometry.h#L1377)). That specifies how the torus is situated in its local frame, and gives the geometry by the centerline radius and the "tube radius" (if the tube radius is zero it degenerates into a circle).

There should also be a longer comment here describing exactly what you get with this Decoration. Probably the `ContactGeometry::Torus` comment will work.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +596,79 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+just generating the initial arrow in the geometry object's local frame.
+
+There is a default constructor for this object but it is not much
+use unless followed by point, direction and length specifications. By default 
+we produce an arrow going from (0,0,0) in direction (1,1,1) of length 1 just 
+so it will show up if you forget to set it to something meaningful. Having a 
+default constructor allows us to have arrays of these objects. **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeArrow : public DecorativeGeometry {
+public:
+    explicit DecorativeArrow(const Vec3& p1 = Vec3(0), const Vec3& dir=Vec3(1), Real length=1.0); // Arrow
```
sherm1: Directions in Simbody are specified with `UnitVec3` rather than `Vec3`. That avoids a runtime normalization and the awkward possibility that the Vec3 could be (0,0,0) which is not allowed for a unit vector. [Here is an example](https://github.com/simbody/simbody/blob/master/Simbody/include/simbody/internal/Force_Gravity.h#L129).

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +596,79 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+use unless followed by point, direction and length specifications. By default 
+we produce an arrow going from (0,0,0) in direction (1,1,1) of length 1 just 
+so it will show up if you forget to set it to something meaningful. Having a 
+default constructor allows us to have arrays of these objects. **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeArrow : public DecorativeGeometry {
+public:
+    explicit DecorativeArrow(const Vec3& p1 = Vec3(0), const Vec3& dir=Vec3(1), Real length=1.0); // Arrow 
+    const Vec3& getOrigin() const;
+    const Vec3& getDirection() const;
+    const Real& getLength() const;
```
sherm1: Is this enough of a specification for a decorative arrow? How is the arrowhead to be scaled with respect to the line length? Is there an algorithm that can satisfy the user's intent? Or do they need some options? I suppose that could be added later if the autoscaling algorithm isn't satisfactory.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +596,79 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+    // Retain the derived type when setting generic geometry options.
+    DecorativeArrow& setBodyId(int b)          { DecorativeGeometry::setBodyId(b);        return *this; }
+    DecorativeArrow& setIndexOnBody(int x)     { DecorativeGeometry::setIndexOnBody(x);   return *this; }
+    DecorativeArrow& setUserRef(void* p)       { DecorativeGeometry::setUserRef(p);       return *this; }
+    DecorativeArrow& setTransform(const Transform& X_BD) { DecorativeGeometry::setTransform(X_BD); return *this; }
+    DecorativeArrow& setResolution(Real r)     { DecorativeGeometry::setResolution(r);    return *this; }
+    DecorativeArrow& setScaleFactors(const Vec3& s) { DecorativeGeometry::setScaleFactors(s); return *this; }
+    DecorativeArrow& setColor(const Vec3& rgb) { DecorativeGeometry::setColor(rgb);       return *this; }
+    DecorativeArrow& setOpacity(Real o)        { DecorativeGeometry::setOpacity(o);       return *this; }
+    DecorativeArrow& setLineThickness(Real t)  { DecorativeGeometry::setLineThickness(t); return *this; }
```
sherm1: If information is to be conveyed by the thickness of the arrow shaft (I have used logarithmic thickness to represent large values rather than making a very, very long line), will line thickness be sufficient (this maps to OpenGL's line thickness support)? Or will you want to draw a cylinder for the shaft?

sherm1:
Thanks, Ayman. A few other questions:
- are you planning also to add partial surfaces, like a hemisphere?
- can you say here how you intend to generate a properly-sized arrowhead? Then we will have a spec in case someone gets in the mood to update the Simbody Visualizer, and it would be good if it follows the GUI. (Maybe better to make that a separate issue.)

sherm1:
You mentioned in the PR description that you want to use these to display torques. Doesn't a torque require a curved arrow?

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
aymanhab: Sherm, thanks for the pointer. I used the terms from OpenSim::WrapTorus instead but I'll switch to Simbody's for consistency since that's where we're heading..

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
aymanhab: Sherm, I debated that and didn't use UnitVec3 because it's not exposed yet to Scripting so it'd make it unusable for now. Possible ways to proceed: combine direction and length in one non-unit vector that can be 0, or assume unit length , fixed direction and use scale factors/transform to scale/orient (as done by vtk), or use unit-vector as suggested and tackle the wrapping issue sometime later. Do you have a preference?

sherm1: > ... use unit-vector as suggested and tackle the wrapping issue sometime later. Do you have a preference?

I do want the main constructor to use `UnitVec3`, but it would be fine to have a "convenience" constructor that takes a `Vec3` if that helps wrapping. Its entire implementation could be this in the header file:
```c++
DecorativeArrow(const Vec3& p1, const Vec3& dir, Real length=1.0) : DecorativeArrow(p1,UnitVec3(dir),length) {}
```

aymanhab: Thanks Sherm, I started down this route but it got a little messy with possibly unnecessary conversions so I decided to go back and wrap UnitVec3, so next rev. will use UnitVec3 exclusively for direction.

aymanhab:
@sherm1 Good point regarding the Arrow "Tip". The API I have used in the GUI assumes default tip-size (fraction), and tip-radius but these can be exposed for future use/flexibility. Another option is to introduce Cone as a separate primitive and then mix/match to make arrows and that would allow up to make curved arrows if needed. 
Other possible option is a "flipped" flag to reverse the direction of the arrow. 

I'd prefer to leave the parts/sections of shapes to another PR since it'd be easier to test them using the WrapObjects in OpenSim when they become available.

aymanhab:
@sherm1 @tkuchida @chrisdembia I thought about this issue a bit more and I believe a Cone would be useful in general so I'm adding it to the interface. This doesn't preclude using any of the methods above to customize Arrow Tip but gives us the tools necessary to customize/experiment.

sherm1:
A Cone sounds good -- can you say how you would propose parameterizing it? I would like to do it the same way for a future ContactGeometry::Cone as well.

aymanhab:
@sherm I went ahead and implemented DecorativeCone, I'd parametrize it very similar to Cylinder, (you'll have to deal with the singularity at apex). I also added a tipLength option for customization of DecorativeArrow. Scaling the arrow will be done using the Transform attached to DecorativeGeometry and that allows for scaling along shaft or radius (decision will be made by the creator of the DecorativeArrow).

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +597,86 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+    }
+
+    SimTK_PIMPL_DOWNCAST(DecorativeTorus, DecorativeGeometry);
+private:
+    class DecorativeTorusRep& updRep();
+    const DecorativeTorusRep& getRep() const;
+};
+
+
+/** An arrow with origin point, direction, height and length. Note that the actual
```
sherm1: not sure what you mean by height and length -- they seem synonymous

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +597,86 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+};
+
+
+/** An arrow with origin point, direction, height and length. Note that the actual 
+placement can be changed by the parent class transform & scale; here we are 
+just generating the initial arrow in the geometry object's local frame.
+
+There is a default constructor for this object but it is not much
+use unless followed by point, direction and length specifications. By default 
+we produce an arrow going from (0,0,0) in direction (0,1,0) of length 1 and tip
```
sherm1: The default direction ought to be (1,1,1)  to match the default for lines. That maximizes the chance that it will be visible somewhere.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +597,86 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+length of .35 just so it will show up if you forget to set it to something 
+meaningful. Having a  default constructor allows us to have arrays of these 
+objects. **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeArrow : public DecorativeGeometry {
+public:
+    explicit DecorativeArrow(const Vec3& p1 = Vec3(0), const UnitVec3& dir = UnitVec3(1), Real length = 1.0, Real tipLength=0.35); // Arrow
```
sherm1: The default should be `UnitVec3(1,1,1)`. As written it is selecting an axis (0, 1, or 2). I shouldn't have allowed that with a plain `int`. There is a `CoordinateAxis` type whose only values are `XAxis,YAxis,ZAxis` that have implicit conversion to 0,1,2 respectively.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -645,6 +726,51 @@ class SimTK_SimTKCOMMON_EXPORT Decorations : public DecorativeGeometry {
...
+here we are just generating the initial cone in the geometry object's local frame.
+
+There is a default constructor for this object but it is not much
+use unless followed by point, direction, height and base-radius specifications. By default
+we produce a cone going from (0,0,0) in direction (0,1,0) of height 1 and base
+radius of 0.5 just so it will show up if you forget to set it to something meaningful. 
+Having a default constructor allows us to have arrays of these objects. **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeCone : public DecorativeGeometry {
+public:
+    explicit DecorativeCone(const Vec3& p1 = Vec3(0), const UnitVec3& dir = UnitVec3(1),
```
sherm1: Same comment for default direction; also the comment above.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -97,7 +97,8 @@ DecorativeGeometry& operator=(const DecorativeGeometry& source);
 
 /** Drawing modes. **/
 enum Representation {
-    DrawPoints    =  1, ///< Use a cloud of points.
+    Hide = 0, ///< hide geometry for now to display later
```
sherm1: What is `Hide` for?

sherm1:
I'm wondering why the `DecorativeArrow` specification uses (start, direction, length) while the `DecorativeLine` specification uses (start, end). That seems inconsistent.

aymanhab:
@sherm1 Thanks for the great feedback, I updated the documentation and default directions per your suggestions. The main issue you brought up not addressed in the commit is that DecorativeLine specifies start,end while DecorativeArrow specifies, start, direction, length. This's because arrows are normally used to display changing quantities (e.g. Force, torque, ...) in these cases the direction matters more than the actual end point. I can foresee forces with fixed direction and changing length, so not having to compute end point would be more convenient. Suggestions to achieve both consistency and convenience are welcome.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
aymanhab: @sherm1 If display preference in the GUI is changed to Hide it's less overhead if the DecorativeGeometry is marked as Hidden rather than completely removed. It also makes our serializable Geometry::Representation enum consistent with Simbody's counterpart.

sherm1:
> This's because arrows are normally used to display changing quantities ...

That is also true of `DecorativeLine` (at least as currently used). If you are using an arrow to represent a vector quantity v, it is easier and cheaper to provide startPoint and endPoint=startPoint+v than it is to calculate length=norm(v), direction=v/length especially since that requires checking for zero length or very small vectors whose direction is indeterminate. In contrast, startPoint+v is always well defined and works fine for v=0. And since this is now inconsistent with DecorativeLine, I would like to see DecorativeArrow changed to match.

Thanks for the clarification on `Hide`; sounds fine. Should be easy enough to implement in the API visualizer!

aymanhab:
@sherm1 Changes have been made to make the DecortiveArrow and DecorativeLine consistent (specify end-points and tip-length), docs have been updated accordingly.

SimTKcommon/Geometry/include/SimTKcommon/internal/DecorativeGeometry.h
```diff
@@ -596,6 +597,130 @@ class SimTK_SimTKCOMMON_EXPORT DecorativeMeshFile : public DecorativeGeometry {
...
+
+There is a default constructor for this object but it is not much
+use unless followed by point, direction, height and base-radius specifications. By default
+we produce a cone going from (0,0,0) in direction (1,1,1) of height 1 and base
+radius of 0.5 just so it will show up if you forget to set it to something meaningful.
+Having a default constructor allows us to have arrays of these objects. **/
+class SimTK_SimTKCOMMON_EXPORT DecorativeCone : public DecorativeGeometry {
+public:
+    explicit DecorativeCone(const Vec3& p1 = Vec3(0), const UnitVec3& dir = UnitVec3(1,1,1),
+        Real height = 1.0, Real base = 0.5); // Cone
```
sherm1: Better to call this `baseRadius` I think.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -583,6 +583,174 @@ const DecorativeMeshFile& getMyMeshFileHandle() const {
...
+    // This is just a static downcast since the DecorativeGeometry handle class is not virtual.
+
+    const DecorativeTorus& getMyTorusHandle() const {
+        return *static_cast<const DecorativeTorus*>(myHandle);
+    }
+};
+
+
+///////////////////////
+// DecorativArrowRep //
```
sherm1: typo

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -583,6 +583,174 @@ const DecorativeMeshFile& getMyMeshFileHandle() const {
...
+
+    void implementGeometry(DecorativeGeometryImplementation& geometry) const {
+        geometry.implementArrowGeometry(getMyArrowHandle());
+    }
+
+    SimTK_DOWNCAST(DecorativeArrowRep, DecorativeGeometryRep);
+private:
+    Vec3 start_point;
+    Vec3 end_point;
+    Real tip_length;
```
sherm1: These don't match the [coding conventions for variable names](https://github.com/simbody/simbody/blob/master/CONTRIBUTING.md#variables). Should be `startPoint` etc. (could also be `m_startPoint` etc. if you want to separate out data members, but otherwise no underscores).

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -583,6 +583,174 @@ const DecorativeMeshFile& getMyMeshFileHandle() const {
...
+
+
+///////////////////////
+// DecorativeConeRep //
+///////////////////////
+
+class DecorativeConeRep : public DecorativeGeometryRep {
+public:
+    // no default constructor
+    explicit DecorativeConeRep(const Vec3& orig, const Vec3& dir, Real& h, Real& baseR) : origin(orig), direction(dir), height(h), baseRadius(baseR) {
```
sherm1: Line too long -- 80 character max. Suggested formatting:
```c++
explicit DecorativeConeRep(const Vec3& orig, const Vec3& dir, Real& h, Real& baseR) 
:   origin(orig), direction(dir), height(h), baseRadius(baseR)  {
}
```

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -583,6 +583,174 @@ const DecorativeMeshFile& getMyMeshFileHandle() const {
...
+    }
+
+    SimTK_DOWNCAST(DecorativeConeRep, DecorativeGeometryRep);
+private:
+    Vec3 origin;
+    UnitVec3 direction;
+    Real height;
+    Real baseRadius;
+
+    // This is just a static downcast since the DecorativeGeometry handle class is not virtual.
```
sherm1: I think this line is also > 80 characters.

Simbody/Visualizer/src/VisualizerGeometry.h
```diff
@@ -51,6 +51,9 @@ class VisualizerGeometry : public DecorativeGeometryImplementation {
     void implementTextGeometry(const DecorativeText& geom);
     void implementMeshGeometry(const DecorativeMesh& geom);
     void implementMeshFileGeometry(const DecorativeMeshFile& geom) {}; // Not handled yet by this Visualizer
+    void implementArrowGeometry(const DecorativeArrow& geom) {}; // Not handled yet by this Visualizer
+    void implementTorusGeometry(const DecorativeTorus& geom) {}; // Not handled yet by this Visualizer
+    void implementConeGeometry(const DecorativeCone& geom) {}; // Not handled yet by this Visualizer
```
sherm1: We should file an issue on this when this code gets checked in. Maybe some outside contributor will get excited about it!

SimTKcommon/Geometry/src/DecorativeGeometry.cpp
```diff
@@ -371,6 +371,104 @@ const std::string& DecorativeMeshFile::getMeshFile() const {
...
+}
+DecorativeTorus& DecorativeTorus::setTubeRadius(Real tubeR) {
+    DecorativeTorusRep::downcast(*rep).setTubeRadius(tubeR); return *this;
+}
+
+/////////////////////
+// DECORATIVE ARROW //
+/////////////////////
+
+DecorativeArrow::DecorativeArrow(const Vec3& startPoint, const Vec3& endPoint, Real tipLength)
```
sherm1: Line too long. Suggested formatting:
```
DecorativeArrow::DecorativeArrow
   (const Vec3& startPoint, const Vec3& endPoint, Real tipLength) {
    rep = new DecorativeArrowRep(startPoint, endPoint, tipLength);
    rep->setMyHandle(*this);
}
```
Other ways are fine too if you think that's ugly.

SimTKcommon/Geometry/src/DecorativeGeometry.cpp
```diff
@@ -371,6 +371,104 @@ const std::string& DecorativeMeshFile::getMeshFile() const {
...
+
+DecorativeArrow& DecorativeArrow::setTipLength(Real tipLength) {
+    DecorativeArrowRep::downcast(*rep).setTipLength(tipLength); return *this;
+}
+
+/////////////////////
+// DECORATIVE CONE //
+/////////////////////
+
+DecorativeCone::DecorativeCone(const Vec3& p1, const UnitVec3& dir, Real height, Real baseRadius)
```
sherm1: Too long.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -583,6 +583,174 @@ const DecorativeMeshFile& getMyMeshFileHandle() const {
...
 
+
+////////////////////////
+// DecorativeTorusRep //
+////////////////////////
+
+class DecorativeTorusRep : public DecorativeGeometryRep {
+public:
+    // no default constructor
+    explicit DecorativeTorusRep(Real torusR, Real tubeR) : torusRadius(torusR), tubeRadius(tubeR) {
```
sherm1: Too long(?)

sherm1:
> Changes have been made to make the DecorativeArrow and DecorativeLine consistent.

Thanks! Looks good aside from minor things I mentioned above.

Simbody/Visualizer/src/VisualizerGeometry.h
aymanhab: Issue #361 has been opened.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
aymanhab: @sherm1 Any preference as to whether the colon should go with the top or bottom line if we split the long line? Thanks.

sherm1: > Any preference as to whether the colon should go with the top or bottom line

Yes, I prefer the colon on the bottom line where I think it is more visible. And I like to align the first initializer at the first tab stop as though it were code since initializers act like assignments. Then the whole thing looks like
```c++
explicit DecorativeConeRep(const Vec3& orig, const Vec3& dir, Real& h, Real& baseR) 
:   origin(orig), direction(dir), height(h), baseRadius(baseR)  {
    int localVar = 0;
    for (;;) {
        // some code
    }
}
```

This is a mild preference; there are lots of other ways to squish the code into 80 characters that are still readable. I like to avoid stretching the code out vertically and the above format does a pretty good job of being compact but clear (IMO).

aymanhab:
@sherm1 The changes you suggested were made and merged but one of the builds fails for unexplained reason that Chris & I looked at and restarted a couple times already. Please let me know if it's ready to merge and if I should wait for the travis build to pass. Thanks.

sherm1:
Thanks, Ayman -- looks good. The restarted builds passed. Merging.

aymanhab:
Awesome, thanks Sherm :+1:

PR #355, opened 2015-03-28
Added Dzhanibekov Effect example

sherm1:
The Dzhanibekov Effect is a non-intuitive rigid body effect in which a smoothly rotating rigid body suddenly flips 180 degrees, then rotates smoothly, then flips again, etc. It is best seen in zero gravity: compare [this one on the space station](https://youtu.be/L2o9eBl_Gzw) with [this one in Simbody](https://youtu.be/L8B83DUKiiA) produced using this example.
 
Erwin Coumans posted a simulation in Bullet [here](https://plus.google.com/+erwincoumans/posts/TAyj5x4Q1HQ), which inspired me to try it in Simbody.

Here is another real-life [space station video](https://youtu.be/JB0OAt4zQ1E?t=29s) (in Russian but it doesn't matter).

### DIFF
--- a/examples/DzhanibekovEffect.cpp
+++ b/examples/DzhanibekovEffect.cpp
@@ -1,0 +1,126 @@
+/* -------------------------------------------------------------------------- *
+ *                      Simbody(tm) Example: Dzhanibekov Effect               *
+ * -------------------------------------------------------------------------- *
+ * This is part of the SimTK biosimulation toolkit originating from           *
+ * Simbios, the NIH National Center for Physics-Based Simulation of           *
+ * Biological Structures at Stanford, funded under the NIH Roadmap for        *
+ * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
+ *                                                                            *
+ * Portions copyright (c) 2015 Stanford University and the Authors.           *
+ * Authors: Michael Sherman                                                   *
+ * Contributors:                                                              *
+ *                                                                            *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
+ * not use this file except in compliance with the License. You may obtain a  *
+ * copy of the License at http://www.apache.org/licenses/LICENSE-2.0.         *
+ *                                                                            *
+ * Unless required by applicable law or agreed to in writing, software        *
+ * distributed under the License is distributed on an "AS IS" BASIS,          *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+/*                      Simbody Dzhanibekov Effect
+This example demonstrates a non-intuitive behavior of a freely rotating
+rigid body, known as the Dzhanibekov Effect. This is best seen in zero 
+gravity on the space station. Here are some cool real-world videos:
+    https://www.youtube.com/watch?v=L2o9eBl_Gzw
+    https://www.youtube.com/watch?v=JB0OAt4zQ1E
+
+    And here is one generated from this example:
+    https://youtu.be/L8B83DUKiiA
+*/
+
+#include "Simbody.h"
+#include <iostream>
+
+using namespace SimTK;
+
+//==============================================================================
+//                              SHOW ENERGY
+//==============================================================================
+// Generate text in the scene that displays the total energy, which should be 
+// conserved to roughly the number of decimal places corresponding to the 
+// accuracy setting (i.e., acc=1e-5 -> 5 digits).
+class ShowEnergy : public DecorationGenerator {
+public:
+    explicit ShowEnergy(const MultibodySystem& mbs) : m_mbs(mbs) {}
+    void generateDecorations(const State&                state, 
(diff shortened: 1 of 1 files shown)


tkuchida:
:+1: :-1: :+1: :-1: :+1:

sherm1:
The GitHub `:-1:` isn't displaying for me :cry:. But `:thumbsdown:` worked:
:+1::thumbsdown::+1::thumbsdown::+1:

You must be a rigid body with unequal inertias!

sherm1:
Now the `:-1:`s are working too :confused:.

tkuchida:
They ǝɹɐ all sqɯnɥʇ ups! :smile:

sherm1:
Wait ... what? How do you make upside-down text?

sherm1:
¡ʍou ʇı ʇǝƃ ı 'ɥo

tkuchida:
Nice timing. I was just about to hit you with [this](http://lmgtfy.com/?q=type+upside+down) :smile:

sherm1:
Yeah, I had that thought right after I asked that dumb question!

sherm1:
Anyway, since you sent so many thumbs up, and Travis and AppVeyor agree with you, I'm merging this example.

tkuchida:
This would make a nice example for the verification and validation talk!

sherm1:
Check it out, @erwincoumans! (See links in PR description)

scpeters:
I added a [gazebo test](https://bitbucket.org/osrf/gazebo/pull-request/1482/add-link-getworldangularmomentum-enable/diff#Ltest/integration/physics_link.ccT92) that captures this effect, though I didn't know this name for that effect. I guess I need to make it easier to visualize the gazebo tests, which mostly run headless.

scpeters:
Is there an easy way to print out the inertia matrix for the rigid body in this test? I'm sure there is actually, but I don't know the simbody API as well as you.

chrisdembia:
Does this work?https://simtk.org/api_docs/simbody/latest/singletonSimTK_1_1Inertia__.html#a8cb59637de516910c73d319d793f53db

scpeters:
Well it's a `MultibodySystem` with two rigid bodies welded together, so I would need to figure out where the c.g. is located, then compute the lumped moment of inertia at that point. So I guess the hard part isn't printing out the inertia matrix, but rather computing it from the `MultibodySystem`.

sherm1:
There is no unique frame in which to express the inertia, since each of the bodies has its own frame. You could ask for it in the ground (world) frame but then you have to have a State so you can find out how the bodies are positioned in space. Probably better to pick one of the bodies as the "main" body and then rotate and shift the other inertia into that frame. What do you have in mind?

sherm1:
Here's a possibility. If you know that the two welded-together bodies are the only ones in the MultibodySystem, then the inertia you are interested in is the central system inertia (meaning system inertia taken about the system COM). There's an API for that:
```c++
    system.realize(state, Stage::Position);
    Inertia sysInertia = matter.calcSystemCentralInertiaInGround(state);
    std::cout << "System Inertia in Ground = " << sysInertia;
    std::cout << "System COM location = " 
        << matter.calcSystemMassCenterLocationInGround(state) << std::endl;
```
where "matter" is the SimbodyMatterSubsystem contained in the system. This inertia is expressed in Ground though, so it will look different in different orientations. To keep it constant you would need to re-express it in some frame fixed to one of the bodies.

scpeters:
I just wanted to see what the inertia matrix would be for this system if it was modeled as a single rigid body instead of two bodies welded together, since the "[Tennis Racket Theorem](http://en.wikipedia.org/wiki/Tennis_racket_theorem)" says that the rotational stability of a single rigid body about each principal axis depends on the relative moment of inertia about each axis. See also the comments in the [gazebo test](https://bitbucket.org/osrf/gazebo/pull-request/1482/add-link-getworldangularmomentum-enable/diff#Ltest/integration/physics_link.ccT92).

I could compute this by hand using the parallel axis theorem, but I was trying to be lazy.

sherm1:
I understand. But since there are two bodies with two different body frames, you have to decide on a frame in which to express the inertia. If their frames are aligned then it doesn't matter; it isn't clear to me whether that is the case here. Simbody has "shift" methods on Inertia, UnitInertia, and MassProperties classes that will take care of parallel axis stuff for you.

scpeters:
The "central system inertia" sounds like what I want, and since I only care about the principal inertia values, I would take that matrix in any frame (such as the Ground frame) and then compute the eigenvalues.

sherm1:
The system central inertia code I posted gave this as output for Simbody's test case with all the state variables set to zero (the way I built it puts the center of the red shaft at Ground 0,0,0):
image
```
System Inertia in Ground =
[0.0612667,0,0]
[0,0.0644667,0]
[0,0,0.00553333]
System COM location = ~[-0.0466667,0,0]
```

erwincoumans:
In Bullet you can compute the same using void btCompoundShape::calculatePrincipalAxisTransform and pass in the inertias/masses/transforms of each component.

scpeters:
That makes sense then. In this model, `Izz` < `Ixx` < `Iyy`, and the `X` axis is the unstable one.

@erwincoumans that's true, I see that's part of the GyroscopicSetup.cpp file.

PR #357, opened 2015-04-03
Improve simbody-visualizer relocatability on Linux.

chrisdembia:
If one relocates their Simbody installation, the primary mechanism for finding
simbody-visualizer no longer works (the original installation path is baked
into the code). The backup plan is to define SIMBODY_HOME; then, $SIMBODY_HOME/bin is added to a search path. This works for windows, since simbody-visualizer is installed to $SIMBODY_HOME/bin. However, on Linux, simbody-visualizer goes into $SIMBODY_HOME/libexec/simbody. With this PR, the *relative* path of simbody-visualizer is ALSO baked into the code, resolving the platform-dependent issue.

Tested on Ubuntu 14.04 and Windows 8.1: I installed Simbody then built a client project that uses the visualizer. At first, my application can find the visualizer. Then I move the installation. My application can no longer find the visualizer. I define SIMBODY_HOME appropriately, and then the application can again find the visualizer.


sherm1:
Looks perfect -- thanks!

PR #358, opened 2015-04-07
Improve creation of symlink from docs to lib installation folder on UNIX systems

elen4:
As mentioned in #356, right now the symbolic link between the examples docs directory and the libexec directory that contains example binaries is wrong when the installation prefix is changes after the first CMake pass. This PR addresses this issue creating the symlink at installation time and without the need for a temporary symlink in the build directory. Also, the symlink is a relative path, in accordance to https://www.debian.org/doc/debian-policy/ch-files.html#s10.5 (please let me know if I am following the wrong policy :confused: )

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,6 @@ PROJECT(Simbody)
 # C:/Program Files/Simbody, C:/Program Files (x86)/Simbody, or the local
 # language equivalent.
 
-# Include GNUInstallDirs to get canonical paths
-include(GNUInstallDirs)
-
 IF(WIN32)
   set (CMAKE_INSTALL_DOCDIR doc)
 ELSE()
@@ -34,6 +31,9 @@ ELSE()
   # problems with some platforms: NTFS on Win, XFS or JFS variants
   set (CMAKE_INSTALL_DOCDIR ${CMAKE_INSTALL_DATAROOTDIR}/doc/simbody)
 ENDIF()
+
+# Include GNUInstallDirs to get canonical paths
+include(GNUInstallDirs)
 
 SET(SIMBODY_MAJOR_VERSION 3)
 SET(SIMBODY_MINOR_VERSION 6)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,11 +25,11 @@ PROJECT(Simbody)
 # language equivalent.
 
 IF(WIN32)
-  set (CMAKE_INSTALL_DOCDIR doc)
+  set (CMAKE_INSTALL_DOCDIR doc CACHE PATH "documentation root (DATAROOTDIR/doc/PROJECT_NAME)")
 ELSE()
   # Redefine DOCDIR to use the project name in lowercase to avoid
   # problems with some platforms: NTFS on Win, XFS or JFS variants
-  set (CMAKE_INSTALL_DOCDIR ${CMAKE_INSTALL_DATAROOTDIR}/doc/simbody)
+  set (CMAKE_INSTALL_DOCDIR share/doc/simbody CACHE PATH "documentation root (DATAROOTDIR/doc/PROJECT_NAME)")
 ENDIF()
 
 # Include GNUInstallDirs to get canonical paths
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -67,9 +67,17 @@ IF(WIN32)
   SET(EXAMPLES_INSTALL_BIN examples/bin/)
   SET(EXAMPLES_INSTALL_SRC examples/src/)
 ELSE()
-  SET(EXAMPLES_INSTALL_BIN ${CMAKE_INSTALL_FULL_LIBDIR}/simbody/examples/)
+  SET(EXAMPLES_INSTALL_BIN ${CMAKE_INSTALL_LIBDIR}/simbody/examples/) # if this changes, change the corresponding
+                                                                      # FULL version in file(RELATIVE_PATH ) command
   SET(EXAMPLES_INSTALL_SRC ${CMAKE_INSTALL_DOCDIR}/examples/src/)
-  SET(EXAMPLES_SYMLINK_BIN ${CMAKE_INSTALL_DOCDIR}/examples/)
+  # Use full paths for the following as it will be used to create a symlink
+  # The symlink itself will be a relative path, so the package will be relocatable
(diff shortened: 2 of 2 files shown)


examples/CMakeLists.txt
```diff
@@ -67,9 +67,17 @@ IF(WIN32)
...
 ELSE()
-  SET(EXAMPLES_INSTALL_BIN ${CMAKE_INSTALL_FULL_LIBDIR}/simbody/examples/)
+  SET(EXAMPLES_INSTALL_BIN ${CMAKE_INSTALL_LIBDIR}/simbody/examples/) # if this changes, change the corresponding
+                                                                      # FULL version in file(RELATIVE_PATH ) command
   SET(EXAMPLES_INSTALL_SRC ${CMAKE_INSTALL_DOCDIR}/examples/src/)
-  SET(EXAMPLES_SYMLINK_BIN ${CMAKE_INSTALL_DOCDIR}/examples/)
+  # Use full paths for the following as it will be used to create a symlink
+  # The symlink itself will be a relative path, so the package will be relocatable
+  # Cannot use CMAKE_INSTALL_FULL_DOCDIR as it evaluates to a different value,
+  # as CMAKE_INSTALL_DOCDIR is changed as a "normal" variable, not in CACHE
```
sherm1: Hmmm ... this seems awkward. I see that the problem comes from the main CMakeLists file [here](https://github.com/simbody/simbody/blob/master/CMakeLists.txt#L35). I don't think the issue is cache vs. normal; looks like someone changed `CMAKE_INSTALL_DOCDIR` but not `CMAKE_FULL_INSTALL_DOCDIR`. They should match, I think. One fix might be to change them both. Another might be to change the PROJECT name on non-WIN32 platforms from PROJECT(Simbody) to PROJECT(simbody). The latter might have side effects I suppose. I wonder also how we manage to get lower case `simbody` names elsewhere, like in the `lib` and `libexec` directories?

elen4:
I realized (reading - finally - the documentation :sweat_smile: ) that overriding the default values *before* including GNUInstallDirs fixes the generation of FULL variables, which occurs in GNUInstallDirs.cmake. I only tested this (i.e., the correctness of `CMAKE_INSTALL_FULL_DOCDIR`) on Windows, will test on Linux shortly. According to the documentation of GNUInstallDirs, the DOCDIR is the only variable where `PROJECT_NAME` is used, therefore if variable overriding works correctly there should be no need to change the project name.

sherm1:
I like that! Thanks, Elena. Let us know when you're satisfied with your testing.

elen4:
I tested on Ubuntu, and it is working as expected. One thing I noticed is that now `CMAKE_INSTALL_DOCDIR`  does not appear in the cache  (advanced) like the other variables (unless you use an "old" build directory), but I don't think this should be a problem, as user changes to that variable wouldn't have effect, since we override that variable from code.

sherm1:
Maybe it would make more sense to make it a cache variable for consistency? In that case user changes would have an effect, even if it is set in the CMakeLists file. I *think* all you would have to do is change the set command in the main CMakeLists.txt to something like this:
```
IF(WIN32)
  set (CMAKE_INSTALL_DOCDIR doc CACHE STRING "Name of doc directory")
ELSE()
  ...
ENDIF()
```
Probably the prompt string should be the same as whatever `GNUInstallDirs` uses for this variable.

chrisdembia:
I tested this on OSX and Windows. Changing the `CMAKE_INSTALL_DOCDIR` from the cmake gui *does* change where the docs are installed.

I also tested the relative symlink on OSX by renaming my install dir; the symlink still worked.

@elen4 can you change the comment for `CMAKE_INSTALL_DOCDIR` on windows (remove `DATAROOTDIR...`)?

After that, I think this is ready to merge.

sherm1:
Oh, Elena I see you made a fix to this a while back that we didn't notice. GitHub doesn't provide notifications when commits are pushed to an existing PR -- you have to also make a comment like "ready for review"!

elen4:
@chrisdembia I can sure change the comment, but I don't really see the need for it, as the original GNUInstallDirs uses the same comment for all OSs. If you like a "cleaner" Windows comment I'll do it anyway, it will only take a sec :smiley: 
@sherm1 I think that I decided to take some time to think through other possible "issues" that remain, i.e., the fact that if a Linux user sets the `CMAKE_INSTALL_DATAROOTDIR` when first invoking CMake, the `DOCDIR` will still point to `share/doc/simbody` instead of `value_set_by_user/doc/simbody`, but then I forgot about it myself :sweat: 
Anyway, thinking about it now, I'd say I'm being over paranoid, and if a user is "smart" enough to play with GNUInstallDirs variables, s/he can check and fix the `CMAKE_INSTALL_DOCDIR` variable anyway :wink:

chrisdembia:
> @chrisdembia I can sure change the comment, but I don't really see the need for it, as the original GNUInstallDirs uses the same comment for all OSs. If you like a "cleaner" Windows comment I'll do it anyway, it will only take a sec :smiley:

Oh okay forget it then.

CMakeLists.txt
```diff
@@ -24,17 +24,17 @@ PROJECT(Simbody)
...
-include(GNUInstallDirs)
-
 IF(WIN32)
-  set (CMAKE_INSTALL_DOCDIR doc)
+  set (CMAKE_INSTALL_DOCDIR doc CACHE PATH "documentation root (DATAROOTDIR/doc/PROJECT_NAME)")
 ELSE()
   # Redefine DOCDIR to use the project name in lowercase to avoid
   # problems with some platforms: NTFS on Win, XFS or JFS variants
-  set (CMAKE_INSTALL_DOCDIR ${CMAKE_INSTALL_DATAROOTDIR}/doc/simbody)
+  set (CMAKE_INSTALL_DOCDIR share/doc/simbody CACHE PATH "documentation root (DATAROOTDIR/doc/PROJECT_NAME)")
```
sherm1: Chris, can you confirm that "share/doc" is right for OSX as well as Linux?

sherm1:
I verified that this merges cleanly with the branch for #356 and works on Linux. Just waiting for Chris regarding `share/doc` on OSX then this can go in.

chrisdembia:
`share/doc` agrees with the README. Looks good to me.

sherm1:
OK, thanks Chris and Elena. Merging this one.

PR #362, opened 2015-04-20
[travis] Use container-based infra. and ccache.

chrisdembia:
This should speed up gcc builds from 6 minutes to 2 minutes, and clang builds from 4 minutes to 2 minutes.

I'm not sure what will happen to OSX builds. Let's see what happens when Travis tests the PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@
 # framework to run Simbody's tests with every GitHub push or pull-request.
 language: cpp
 
+sudo: false
+
 os:
   - linux
   - osx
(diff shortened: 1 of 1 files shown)


sherm1:
Strange -- 14 minutes. Failed on OSX because it couldn't find ccache at the end.

chrisdembia:
> Strange -- 14 minutes.

Because the cache hadn't been created yet! Feast your eyes on the new build. 

Also, I think I fixed the OSX issue.

chrisdembia:
Also, 2 minutes is the best case in which no files have changed. I don't know how performance will change when source files change. Also I'm not sure the extent to which the cache is used with pull requests. I'm thinking we can at least try it out for a while? I'm more interested in the time savings for OpenSim (I'm almost done ccaching OpenSim).

chrisdembia:
So, the OSX build doesn't get cached and it's the longest build. What do you think about using up the time savings by running the tests in Debug as well?

sherm1:
Wow -- that's fast. Took 25 seconds to do the build. But it wasn't clear to me that it actually built anything. If it does avoid building things, can we be confident that it always builds what needs to be built? OpenSim's CMake is famous for missing a few things now and again.

Seems somewhat abusive to run tests in Debug. I do like to do that once in a while but I don't think it needs to be done continuously.

chrisdembia:
> If it does avoid building things, can we be confident that it always builds what needs to be built? OpenSim's CMake is famous for missing a few things now and again.

Here's what the ccache site says:

> Is it safe? 
> Yes. The most important aspect of a compiler cache is to always produce exactly the same output that the real compiler would produce. This includes providing exactly the same object files and exactly the same compiler warnings that would be produced if you use the real compiler. The only way you should be able to tell that you are using ccache is the speed.

> ccache has been coded very carefully to try to provide these guarantees. However, if you experience any bugs, please report them.

Here's how ccache works:

> The basic idea is to detect when you are compiling exactly the same code a second time and reuse the previously produced output. The detection is done by hashing different kinds of information that should be unique for the compilation and then using the hash sum to identify the cached output. ccache uses MD4, a very fast cryptographic hash algorithm, for the hashing. (MD4 is nowadays too weak to be useful in cryptographic contexts, but it should be safe enough to be used to identify recompilations.) On a cache hit, ccache is able to supply all of the correct compiler outputs (including all warnings, dependency file, etc) from the cache.
ccache has two ways of doing the detection:
the direct mode, where ccache hashes the source code and include files directly
the preprocessor mode, where ccache runs the preprocessor on the source code and hashes the result
The direct mode is generally faster since running the preprocessor has some overhead.

https://ccache.samba.org/manual.html#_synopsis

Also, we can clear the cache every so often.

sherm1:
Sounds good -- more reliable than CMake's method.

chrisdembia:
@sherm1 I'm going to close this for now. Perhaps Simbody's tests run fast enough. If you're interested in the future, we can revisit.

PR #363, opened 2015-04-21
Modernize the SimTK::Array_ class for C++11 compatibility

sherm1:
Added std::initializer_list construction, move construction, and move assignment. Updated doxygen comments. Added test cases. Updated Changelog.

This PR makes `SimTK::Array_` compatible with the C++11 standard `std::vector`, with the exception of `emplace()` and `emplace_back()` method which still need to be done.

The original motivation for this class was that Microsoft put a bunch of junk in its `std::vector` for "safety" that caused it to run too slowly. I ran timings in VS 2013 to see if that is still true (see the TestArray regression test for the timing code). Array_ ran twice as fast as std::vector (roughly 300ms vs 600ms). I ran the test on gcc 4.8.1 and clang 3.4 on the same machine. No advantage for Array_ there, *but* all the tests took 1000-1200ms, twice as slow as Microsoft's *bad* implementation. Not sure what to make of that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,14 @@ Simbody Changelog and Release Notes
 
 This is not a comprehensive list of changes but rather a hand-curated collection of the more notable ones. For a comprehensive history, see the [Simbody GitHub repo](https://github.com/simbody/simbody). You can use the release dates below to find all the PRs and issues that were included in a particular release. 
 
-**Heads up**: Simbody 3.5 will be the last release that will build with C++03 (patch builds with version numbers like 3.5.1, if any, will work too). For 3.6 and above we will permit Simbody developers to use C++11, restricted to the subset that is currently supported on all our platforms. Since the C++03 and C++11 ABIs are not compatible, code that uses Simbody 3.6 will also have to be built with C++11. Time to move up, if you haven't already!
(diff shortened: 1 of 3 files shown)


chrisdembia:
> twice as slow as Microsoft's bad implementation.

I don't want to believe it.

sherm1:
>> twice as slow as Microsoft's bad implementation.

> I don't want to believe it.

OK, I went and fiddled with the settings. For gcc and clang we use "-O2" as the optimization level. There is a more extreme "-O3" available, but we've had trouble with it at times. I switched to "-O3" which had some odd results:
- for gcc it reduced the penalty to about 50% (900 ms for either std::vector or SimTK::Array_)
- for clang std::vector sped up dramatically to 300ms -- as fast as Array_ on VC2013. But ... Array_ runs at 1300ms, slower than at "-O2" :cry:.

That's enough wiggle room for you to believe whatever you want about the relative merits of these compilers!

chrisdembia:
wow! interesting.

sherm1:
Further Array_/std::vector performance developments:

It occurred to me that the default index type for std::vector is 64 bit (unsigned long long, not changeable) while the default for SimTK::Array_ is 32 bit (plain unsigned but settable as a template parameter). So in the speed test I used unsigned long long as the index type for Array_. Here are the results, still using -O3 optimization:
- clang: now both vector and Array_ run at 300ms 
- gcc: Array_ runs around 380ms, very good, but vector is over 900ms
- VS2013: no change (vector at 600ms, Array_ at 300ms)

Probably this means gcc & clang didn't do a very good job optimizing 32 bit integers but paid more attention to 64, while Microsoft was careful with both.

PR #367, opened 2015-04-28
Fixed element-initializing Matrix constructor (failing in Debug builds)

### LINKED ISSUE #366: Filling Matrix with initial value incorrectly requesting length() instead of nelt()?
Initializing a matrix fails [here](https://github.com/simbody/simbody/blob/master/SimTKcommon/BigMatrix/src/MatrixHelper.cpp#L774) in debug since 'length()` assumes 1D matrix (row or column Vector).

sherm1:
Fixes #366. The Matrix element-initializing constructor like Matrix(2,3, 3.14) was failing in Debug builds for any Matrix where all the data is contiguous, due to a broken optimization for that case.

Added test cases.

/cc @aseth1


sherm1:
This PR built cleanly now. Do you want to give it a quick review before it goes in, @aseth1?

aseth1:
looks good to me.

PR #372, opened 2015-05-15
Minor follow up mods to `Pathname` specified working directory additions

sherm1:
I made a few follow-on updates to @carmichaelong's PR #307:
- renamed *find*AbsolutePathnameUsingSpecifiedWorkingDirectory() to *get*... for consistency with the other method names; fixed test program to match
- fixed doxygen comments so that numbered list formats properly
- added missing method getAbsolute*Directory*PathnameUsingSpecifiedWorkingDirectory()
- updated copyright and authorship; added Carmichael to the contributors table
- updated changelog to mention new pathname methods
- some minor additions to comments
- fixed addDirectoryOffset() method so that the result always uses the platform-appropriate slashes

Carmichael, please review (other interested parties also welcome).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 
 3.6 (in development)
 --------------------
-* Added C++11 features to the SimTK::Array_ container including std::initializer_list construction, move construction, and move assignment.
+* Added C++11 features to the `SimTK::Array_` container including `std::initializer_list` construction, move construction, move assignment, and `emplace` methods.
 * Make doxygen run silently so errors will be easier to see.
+* Added new methods to `Pathname` class for interpreting pathnames against a specified working directory instead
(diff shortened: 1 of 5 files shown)


sherm1:
@carmichaelong, I'm going to merge this today unless you object.

carmichaelong:
Only comment is that the function prototypes were changed for some functions in the .h files, but they aren't reflected int he .c files. Not a big deal, just don't know what your coding standards are for that. Feel free to merge whenever.

sherm1:
> function prototypes were changed for some functions in the .h files, but they aren't reflected in the .c files

Thanks, Carmichael. It's not required, but usually better if they match. 
Sometimes there is a good reason for having different names. The names in the header file are part of the user documentation, while the names in the .cpp are seen only by developers. For example, I may use something like `mobilizedBodyIndex` as the name of a parameter in the header, while using `mbx` or even just `i` in the .cpp if the long name would clutter up the code too much.
I'll take a look at these and make them match unless it is too painful.

sherm1:
OK, I fixed the names Carmichael -- thanks! Merging now.

PR #376, opened 2015-06-02
Attempt to create v3.5 nuget package.

chrisdembia:
This is intended to be used for testing the opensim32 repo.

Related to https://github.com/opensim-org/opensim32/issues/19

### DIFF
--- a/.simbody.nuspec
+++ b/.simbody.nuspec
@@ -2,7 +2,7 @@
 <package >
   <metadata>
     <id>simbody</id>
-    <version>0.0.0</version>
+    <version>3.5</version>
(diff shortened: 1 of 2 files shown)


sherm1:
Do you want to be more precise with the version number? Could make it v3.5.3 instead of just v3.5.

Either way feel free to merge.

chrisdembia:
I considered that and decided against it. Basically, I don't want to manually edit those files again if there's a new 3.5 release.

However, if we use CMake to configure the `.simbody.nuspec` file, then we could use the full version number. We can revisit this later on.

chrisdembia:
Thanks

PR #382, opened 2015-06-04
Edit the "Type Erasure" classes to allow for retrieving the stored(type erase) object.

klshrinidhi:
* Add two methods that use the type information provided by caller to 'cast' the abstract type the correct concrete type.
* Add a move constructor to the concrete class.

This modification will allow us create containers of *heterogeneous* types. We erase the original type of the object by storing it in Value class and use a reference/pointer to the AbstractValue class in the containers. The original object(properly typed) can be retrieved through the added methods.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -55,26 +55,22 @@ public:
     /// Retrieve the stored object as const ref -- read-only.
     template<typename T>
     const typename std::remove_reference<T>::type& getValue() {
+      using ErasedType = typename std::remove_reference<T>::type;
 #ifndef NDEBUG
-      return dynamic_cast<const 
-                          Value<typename
-                                std::remove_reference<T>::type>&>(*this).thing;
+      return dynamic_cast<Value<ErasedType>&>(*this).thing;
 #else
-      return static_cast<const 
-                         Value<typename 
-                               std::remove_reference<T>::type>&>(*this).thing;
+      return static_cast<Value<ErasedType>&>(*this).thing;
 #endif
     }
 
     /// Retrieve the stored object as ref -- read-write.
     template<typename T>
     typename std::remove_reference<T>::type& updValue() {
+      using ErasedType = typename std::remove_reference<T>::type;
 #ifndef NDEBUG
-      return dynamic_cast<Value<typename 
-                                std::remove_reference<T>::type>&>(*this).thing;
+      return dynamic_cast<Value<ErasedType>&>(*this).thing;
 #else
-      return static_cast<Value<typename 
-                               std::remove_reference<T>::type>&>(*this).thing;
+      return static_cast<Value<ErasedType>&>(*this).thing;
 #endif
     }
 
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -57,12 +57,12 @@ public:
     const typename std::remove_reference<T>::type& getValue() {
 #ifndef NDEBUG
       return dynamic_cast<const 
-			  Value<typename 
-				std::remove_reference<T>::type>&>(*this).thing;
+                          Value<typename
+                                std::remove_reference<T>::type>&>(*this).thing;
 #else
       return static_cast<const 
-			 Value<typename 
-			       std::remove_reference<T>::type>&>(*this).thing;
+                         Value<typename 
+                               std::remove_reference<T>::type>&>(*this).thing;
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


SimTKcommon/include/SimTKcommon/internal/Value.h
```diff
@@ -47,7 +51,33 @@ class AbstractValue {
...
+    typename std::remove_reference<T>::type& updValue() {
+#ifndef NDEBUG
+      return dynamic_cast<Value<typename 
+				std::remove_reference<T>::type>&>(*this).thing;
+#else
+      return static_cast<Value<typename 
+			       std::remove_reference<T>::type>&>(*this).thing;
+#endif
+    }
+
```
klshrinidhi: The `remove_reference` utility is used here to account for users calling the function with reference types like `getValue<Foo&>`. In such cases, the reference is stripped.

sherm1:
Copied from my email:

I just realized that I recently reworked the Value.h header in Simbody’s develop-4.0 branch, but it isn’t merged into master. That version is a lot cleaner and nicer because I wrote a bunch of doxygen for it. I can merge it later, but in case you prefer the newer version you can take a look at it [here](https://github.com/simbody/simbody/blob/develop-4.0/SimTKcommon/include/SimTKcommon/internal/Value.h). It would be fine to put that in master since it doesn’t depend on any other 4.0 stuff. It would require one small change to avoid compilation errors in master: add an additional non-const overload of `Value<T>::downcast()` that just invokes `updDowncast()`.

sherm1:
Unfortunately I had written the original `AbstractValue/Value<T>` classes many years ago before we settled on our current coding guidelines. The updated one in develop-4.0 is conformant and is probably a better place to start for your mods. Sorry I didn't put it in master!

sherm1:
One possibility: cherry pick the develop-4.0 commit for Value.h over to master, then reapply your changes. I could do the cherry pick if you want -- let me know.

sherm1:
I'm not sure about the `std::remove_reference` stuff. Why would we want to allow anyone to create a `Value<T&>`? We can't actually store a reference so I would prefer that it not compile. Do you have a use case in mind for this?

sherm1:
In AbstractValue, would this work:
```c++
template <typename T>
const T& getValue() const {
    return Value<T>::downcast(*this).get();
}
```
That way AbstractValue doesn't reference `thing` directly. Also the `downcast` method handles the dynamic_cast/static_cast switch between Release/Debug.

sherm1:
I am making a PR that has the develop-4.0 cleaned up code for you to review.

klshrinidhi:
Without `remove_reference`, when users say `getValue<SomeType&>`, compiler will throw an error about not being able to instantiate `Value<SomeType&>`. It will not be intuitive enough to understand that the template argument must not be of a reference type. So I thought I should do one of:
* Allow reference types in the template argument and remove the reference within the function.
* Dis-allow reference types in the template argument, which can be done using SFINAE techniques. 
I chose the first approach for no strong reason.

If you don't think this is a valid concern, we can get rid of the meta-programming.

klshrinidhi:
I think your suggestion of : 
```c++
template <typename T>
const T& getValue() const {
    return Value<T>::downcast(*this).get();
}
```
looks great and can be used. I am assuming there will be another function called `updValue` as well.

sherm1:
> If you don't think this is a valid concern, we can get rid of the meta-programming.

I don't think we have to worry about people directly writing `getValue<T&>()`. That doesn't make logical sense here so I'm happy to have it fail in whatever ugly manner. But conceivably there would be situations where someone else might be doing some metaprogramming where the template argument is produced by `decltype` or whatever where this would be a convenience. OTOH, those people could use `remove_reference` themselves prior to instantiating `getValue<T>()`. My preference would be simply to document any appropriate limitations on T (someday, enforced with Concepts), and count on future metaprogrammers to be smart enough to use the class correctly. One argument in favor of this approach is that we have many templatized classes in Simbody and OpenSim already that do not attempt to tidy up their template arguments, and this hasn't yet been an issue (might be in the future though). A second is that bulletproof template code is so ugly as to be almost unreadable -- a quick look through the standard template library implementation is a horrifying experience. 

My preference would be to favor simpler code except in cases where we know from real use cases that it is worth some layers of metaprogramming. As long as we document what's allowed now, I think we are free to improve it later since it won't change the API except to remove some restrictions. You could easily change my mind on this if you feel strongly about it!

sherm1:
> I will add the two methods to the AbstractValue class in a commit to the branch associated with #383.

Good plan.

klshrinidhi:
Sounds good about the `remove_reference` issue. Let us keep it simple and document the details. 
And I agree that C++ Standard Library is filled with template meta-programming and it is likely to be increasingly so in future for the library as it tries to achieve various ambitious goals without sacrificing type safety and performance.

sherm1:
Shrinidhi made this change in PR #383 so I'm closing this one.

PR #383, opened 2015-06-04
Clean up, document, and modernize AbstractValue and Value<T> code

sherm1:
I cleaned up this code in the develop-4.0 branch; this PR copies the cleaned-up code from there. 

The reason is that @klshrinidhi wants to make some changes to it (see PR #382) but the master branch code is very old and ugly so it would be better to add the changes to this version instead.

The functionality and API are unchanged but now there is actually some doxygen to say what each method is supposed to do.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -35,6 +35,8 @@
 #include <utility>
 
 namespace SimTK {
+  // Forward declaration.
+  template<typename T> class Value;
 
 //==============================================================================
 //                             ABSTRACT VALUE
@@ -70,6 +72,20 @@ public:
     if the source object is compatible, or throw an exception otherwise. **/
     AbstractValue& operator=(const AbstractValue& v) 
     {   compatibleAssign(v); return *this; }
+
+    /** Retrieve the original (type-erased)*thing* as read-only. The template
+    argument must be exactly the non-reference type of the stored *thing*. **/
+    template <typename T>
+    const T& getValue() const {
+      return Value<T>::downcast(*this).get();
+    }
+
+    /** Retrieve the original (type-erased)*thing* as read-write. The template
+    argument must be exactly the non-reference type of the stored *thing* **/
+    template <typename T>
+    T& updValue() {
+      return Value<T>::updDowncast(*this).upd();
+    }
 
     virtual ~AbstractValue() {}   
 };
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
  *                                                                            *
- * Portions copyright (c) 2005-12 Stanford University and the Authors.        *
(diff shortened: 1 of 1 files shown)


sherm1:
Shrinidhi, please take a look at this and comment, modify, or merge it as you see fit. I would rather see your changes made to this version than the older one. Sorry for the inconvenience -- it didn't occur to me that anyone would be modifying these classes prior to the 4.0 switch.

sherm1:
Whoops -- sorry for the delay, Shrinidhi. GitHub doesn't generate a notification when you push to an existing PR and I didn't think to look until just now. I usually add a "ready for review" comment after a new commit to notify reviewers that it is their turn.

SimTKcommon/include/SimTKcommon/internal/Value.h
```diff
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
  *                                                                            *
- * Portions copyright (c) 2005-12 Stanford University and the Authors.        *
+ * Portions copyright (c) 2005-15 Stanford University and the Authors.        *
  * Authors: Michael Sherman                                                   *
  * Contributors:                                                              *
```
sherm1: Add yourself to the contributors list, unless you prefer anonymity. People usual add themselves to Contributors for small changes or to Authors for major changes.

SimTKcommon/include/SimTKcommon/internal/Value.h
```diff
@@ -32,72 +32,204 @@
...
+    that value into this object. Otherwise an exception is thrown. **/
+    virtual void compatibleAssign(const AbstractValue& source) = 0;
+    
+    /** Invokes the compatibleAssign() method which will perform the assignment
+    if the source object is compatible, or throw an exception otherwise. **/
+    AbstractValue& operator=(const AbstractValue& v) 
+    {   compatibleAssign(v); return *this; }
+
+    /** Retrieve the original (type-erased)*thing* as read-only. The template
+    argument must be exactly the non-reference type of the stored *thing*. **/
```
sherm1: If you mean to refer to the literal variable name `thing` in a doxygen comment use code font rather than italics, that is use back single quotes (\`thing\`) rather than italics (\*thing\*). If you just meant this generically then italics is fine, but might be unclear because the `AbstractValue` user won't necessarily know that `Value<T>` calls its internal T a "thing". Maybe call it a "type T object"?

klshrinidhi: I actually meant to have `thing` refer to the class member but didn't bother to look up the markdown. I haven't done much Doxygen documentation before.
I just made the change to enclose *thing* with back-tick.

SimTKcommon/include/SimTKcommon/internal/Value.h
klshrinidhi: I think I am fine without this. If I have a major change to contribute in future, I will include my name.

klshrinidhi:
Ready for review.

sherm1:
Looks great -- thanks, Shrinidhi! 

Regarding doxygen, it is worthwhile to generate it occasionally with `make doxygen` and then look at it in a browser (it's in your binary directory under `html/index.html`) to see how it will look to end users. Doxygen can be quirky so it isn't always easy to predict how things are going to look.

chrisdembia:
:fireworks:

PR #386, opened 2015-06-04
(c|r)poly.cpp: fabs->std::abs.

### LINKED ISSUE #384: clang 3.6 warns about rpoly.cpp fabs vs std::abs
@klshrinidhi got me trying clang-3.6, which gives the following warning:

```
/Users/chris/repos/simbody/1/simbody/SimTKcommon/Polynomial/src/rpoly.cpp:311:29: warning: 
      absolute value function 'fabs' given an argument of type 'long double' but
      has parameter of type 'double' which may cause truncation of value
      [-Wabsolute-value]
        if (ss != 0.0) ts = fabs((ss-oss)/ss);
                            ^
/Users/chris/repos/simbody/1/simbody/SimTKcommon/Polynomial/src/rpoly.cpp:311:29: note: 
      use function 'std::abs' instead
        if (ss != 0.0) ts = fabs((ss-oss)/ss);
                            ^~~~
                            std::abs
```

I'm really scared of editing `rpoly.cpp` because it uses `goto`'s. Are the `goto`'s affected by macros? Two possible fixes: ignore the warning or use `std::abs`.

chrisdembia:
Fixes #384

### DIFF
--- a/SimTKcommon/Polynomial/src/cpoly.cpp
+++ b/SimTKcommon/Polynomial/src/cpoly.cpp
@@ -547,7 +547,7 @@ void CPoly<T>::cauchy( const int nn, T pt[], T q[], T *fn_val )
    dx = x;
    
    // Do Newton iteration until x converges to two decimal places
-   while( fabs( dx / x ) > 0.005 )
+   while( std::abs( dx / x ) > 0.005 )
       {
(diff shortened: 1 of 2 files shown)


sherm1:
Thanks, Chris. Build was clean.

PR #387, opened 2015-06-05
Viz antialias

chrisdembia:
Tested on OSX and Linux. The other view buttons still work.

The saved images are not antialiased. I tried to insert the `glEnable(GL_MULTISAMPLE)` into different places to try and get antialiased images, but I couldn't get it to work. Any ideas?

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/387)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 * Make doxygen run silently so errors will be easier to see.
 * Added new methods to `Pathname` class for interpreting pathnames against a specified working directory instead
 of the current working directory (thanks to Carmichael Ong). See [Issue #264](https://github.com/simbody/simbody/issues/264) and [PR #307](https://github.com/simbody/simbody/pull/307). 
+* The simbody-visualizer now has the option to provide smoother visualization
+via an anti-aliasing option (in the visualizer, click View -> Antialias).
+Note that the
+anti-aliasing may slow down the visualization, and does not affect the quality
+of images saved via "Save Image" or "Save Movie".
 * (There are more that haven't been added yet)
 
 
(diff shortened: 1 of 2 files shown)


sherm1:
Any thoughts on this @peastman? Chris enabled antialiasing for the visualizer and it works fine but the images that are saved to files aren't antialiased.

sherm1:
Chris, I tried this on Windows and could see any difference between antialiasing on or off. Same result on Windows and on my Ubuntu VM. Possibly I don't know what to look for. I tried the Simbody "Rattleback" example which has an ellipsoid. Do you have a better test case?

chrisdembia:
Thanks for testing. I presume you mean "Same result on OSX and on my Ubuntu VM." I've noticed the effect with ellipsoids, but for this PR I was looking at TaskSpaceControl-Atlas. Even with antialiasing, there is still some aliasing but it should definitely look better. 

There is an OpenGL parameter to increase the number of samples in the multisampling.

sherm1:
> I presume you mean "Same result on OSX and on my Ubuntu VM."

No, I'm not running OSX. It is Windows 8.1.

sherm1:
This might provide some info on how to get antialiasing working for saved images and movies: http://stackoverflow.com/questions/20482851/how-to-use-multisampling-with-opengl-fbos

The visualizer makes its own frame buffer to render those into; that also has to be created multisampled.

chrisdembia:
Thanks @sherm1 for that link. I quickly tried to do what they said but then the saved images came out as black. I'd prefer to merge this without anti-aliased images. We can revisit the anti-aliasing of images later. 

Could you send me screenshots of the visualizer on your computer with/without anti-aliasing?

sherm1:
Antialiasing off: 
image

Antialiasing on (or at least I selected it):
image

I don't see any difference between them.

chrisdembia:
Yeah antialiasing isn't working there. Here's on OSX:

without antialiasing:
screen shot 2015-06-07 at 8 12 33 pm

with antialiasing:
screen shot 2015-06-07 at 8 13 35 pm

Focus on the upper edge of the ground.

adamkewley:

Issue #328, opened 2015-01-20
Simbody's smart pointer APIs should better match C++11's


sherm1:
The smart pointers `ClonePtr` and `ReferencePtr` have APIs that are somewhat different from the ones defined by C++11 for `std::unique_ptr` and `std::shared_ptr`. These should be upgraded to match better (in 4.0 since the changes may not be backwards compatible). ClonePtr should get a move constructor and assignment.

The newer `CloneOnWritePtr` matches C++11 and the other two should be made to behave similarly.

sherm1:
Note that in #390 we stopped allowing for null references in `ClonePtr`. Should do the same here.

sherm1:
Fixed in PR #410.

PR #390, opened 2015-06-08
Remove check of reference against nullptr.

### LINKED ISSUE #388: clang 3.6 gives "Undefined bool conversion" in ClonePtr
```
/Users/chris/repos/simbody/1/install/include/simbody/SimTKcommon/internal/ClonePtr.h:84:16
: warning: 
      reference cannot be bound to dereferenced null pointer in well-defined C++
      code; pointer may be assumed to always convert to true
      [-Wundefined-bool-conversion]
    {   reset(&t ? t.clone()  :0); return *this; }
```

http://stackoverflow.com/questions/28689269/how-to-fix-reference-cannot-be-bound-to-dereferenced-null-pointer-warning

Any ideas on how to fix this?

chrisdembia:
Fixes #388.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
@@ -65,7 +65,7 @@ public:
     copy of that object via its clone() method and make this %ClonePtr
     object the owner of the copy. Ownership of the original object is not
     affected. **/
-    explicit ClonePtr(const T& obj) : p(&obj?obj.clone():0) { }
+    explicit ClonePtr(const T& obj) : p(obj.clone()) { }
     /** Copy constructor is deep; the new %ClonePtr object contains a new
     copy of the object in the source, created via the source object's clone()
     method. If the source container is empty this one will be empty also. **/
@@ -81,7 +81,7 @@ public:
     heap-allocated copy of the source object. The copy is created using the 
     source object's clone() method. The currently-held object is deleted. **/    
     ClonePtr& operator=(const T& t)          
-    {   reset(&t ? t.clone()  :0); return *this; }
+    {   reset(t.clone()); return *this; }
     /** This form of assignment replaces the currently-held object by the given
     source object and takes over ownership of the source object. The 
     currently-held object is deleted. **/ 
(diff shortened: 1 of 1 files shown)


sherm1:
I see that I took a different approach in Simbody 4.0 for `CloneOnWritePtr`. I created a little helper routine:
```
    // If src is non-null, clone it; otherwise return null.
    static T* cloneOrNull(const T* src) {
        return src ? src->clone() : nullptr;
    }
```
When given a `const T& t`, this is used like this: `reset(cloneOrNull(&t))`. I didn't do that to avoid the warning but to avoid the many duplicate `?:` tests -- pointers p use `cloneOrNull(p)`. I suspect that would slip by clang 3.6 silently and would still guard against null references.

Please feel free to merge it as-is but if so please add a note to issue #328 to mention that ClonePtr & CloneOnWritePtr should be made to match wrt cloneOrNull() eventually.

chrisdembia:
I'm not sure I understand the reason for giving the user a nullptr instead of causing the segfault. Is it so they can check if they can have logic that tests for the nullptr?

sherm1:
No, it would only be to help out bozos who went to the trouble to create an illegal null reference object. I think it would be better *not* to check the references for null, that is, the way you fixed it. But it would also be better to use cloneOrNull() rather than a bunch of repeated code for the pointers, where it is legitimate to expect nulls. The point of issue #328 is that the two pointer implementations ought to be almost identical.

chrisdembia:
So it sounds like what we should do is, for methods taking references, just go ahead and clone (in both pointer classes), and for methods taking pointers, use `cloneOrNull()`? I can try this.

sherm1:
Yeah, that seems like the best approach. I'm slightly torn because I know people do occasionally create null references but we probably shouldn't encourage the practice by hiding the consequences.

sherm1:
I'm going to have to rework ClonePtr soon anyway to bring it into C++11 compliance, so if you want to leave it with just the existing fix that's fine.

chrisdembia:
> I'm slightly torn because I know people do occasionally create null references but we probably shouldn't encourage the practice by hiding the consequences.

You have the benefit/curse of experience, while I have the benefit/curse of naive idealism.

chrisdembia:
I tried to improve consistency within ClonePtr by adding a cloneOrNull(), but I do not plan on making similar changes to CloneOnWritePtr.

sherm1:
Thanks -- perfect.

PR #405, opened 2015-06-29
Add CloneOnWritePtr smart pointer class

sherm1:
This PR backports the `CloneOnWritePtr` smart pointer from the Simbody 4.0 development branch. This is applicable to any class that has a `clone()` method and provides ordinary, non-shared copy and assignment functionality (like `ClonePtr`) but defers the actual copying operation until someone asks for mutable (non-const) access. So the object is shared for multiple readers, but anyone can write on it without affecting any other user because they will magically get their own just-in-time copy.

The API is intended to come as close as possible to the C++11 APIs for `std::unique_ptr` and `std::shared_ptr`; `CloneOnWritePtr` borrows some functionality from each. Note that the older `ClonePtr` API is somewhat different -- I will update it to match this one after this is reviewed.

/cc @klshrinidhi would you review this, please? (other interested parties welcome too!)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/basics.h
+++ b/SimTKcommon/include/SimTKcommon/basics.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
  *                                                                            *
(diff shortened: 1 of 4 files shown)


klshrinidhi:
I looked at the interface. I think it looks great. Will be very useful.

sherm1:
OK, thanks Shrinidhi. I'll merge this and then make the ClonePtr API match.

PR #407, opened 2015-07-01
Add DEPRECATED_14 macro

sherm1:
This PR adds a `DEPRECATED_14("message")` macro to be used until we can require C++14 everywhere. For now it expands to the standard `[[deprecated("message")]]` when compiled with C++14 but otherwise expands into non-standard compiler-specific hacks that work the same way.

This required turning off general suppression of "deprecated" warnings in VC++, replacing with local deprecation where necessary. That may affect users of Simbody (possibly including OpenSim), if they are using functions that VC++ deems (incorrectly) to have been deprecated, such as `sprintf()`, `getenv()`, and `ctime()`, and were counting on Simbody to silence those warnings.

I also used the new macro to deprecate `ClonePtr::getPtr()` and `updPtr()` in favor of more C++11-like `get()` and `upd()`.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
  *                                                                            *
- * Portions copyright (c) 2005-12 Stanford University and the Authors.        *
+ * Portions copyright (c) 2005-15 Stanford University and the Authors.        *
  * Authors: Michael Sherman                                                   *
  * Contributors:                                                              *
  *                                                                            *
@@ -148,12 +148,17 @@ public:
     the container is non-null (that is, not empty). **/
     operator bool() const { return !empty(); }
 
-    /** Return a writable pointer to the contained object if any, or null. 
-    You can use the "address of" operator\&() instead if you prefer. **/
-    T* updPtr() { return p; }
-    /** Return a const pointer to the contained object if any, or null.  
-    You can use the "address of" operator\&() instead if you prefer. **/
-    const T* getPtr()  const  { return p; }
+    /** Return a const pointer to the contained object if any, or nullptr. Note 
+    that this is different than `%get()` for the standard smart pointers which 
+    return a writable pointer. Use upd() here for that purpose. 
+    @see upd(), getRef() **/
+    const T* get()  const  { return p; }
+
+    /** Return a writable pointer to the contained object if any, or nullptr. 
+    Note that you need write access to this container in order to get write 
+    access to the object it contains.
+    @see get(), updRef() **/
+    T* upd() { return p; }
 
     /** Return a writable reference to the contained object. Don't call this if
     this container is empty. There is also an implicit conversion to reference
@@ -202,7 +207,11 @@ public:
         other.reset(p);
         reset(otherp);
     }
-     
+
+    DEPRECATED_14("use get() instead")
+    const T* getPtr()  const  { return get(); }
+    DEPRECATED_14("use upd() instead")
+    T* updPtr() { return upd(); }
 private:
     // Warning: ClonePtr must be exactly the same size as type T*. That way
     // one can reinterpret_cast a T* to a ClonePtr<T> when needed.
--- a/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
@@ -208,8 +208,12 @@ public:
         reset(otherp);
     }
 
+    /** <b>(Deprecated)</b> Same as `get()`. Use get() instead; it is more like 
+    the API for `std::unique_ptr`. **/
     DEPRECATED_14("use get() instead")
     const T* getPtr()  const  { return get(); }
+    /** <b>(Deprecated)</b> Same as `upd()`. Use upd() instead; it is a better 
+    match for `get()` modeled after the API for `std::unique_ptr`. **/
     DEPRECATED_14("use upd() instead")
     T* updPtr() { return upd(); }
 private:
--- a/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
+++ b/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
@@ -28,6 +28,7 @@
 #include "lodepng.h"
 
 #include <cstdlib>
+#include <cmath>
 #include <string>
 #include <algorithm>
 #include <set>
@@ -98,16 +99,22 @@ static void shutdown();
 #ifdef _WIN32
     #include <io.h>
     #define READ _read
+    #define WRITEFUNC _write
 #else
     #include <unistd.h>
     #define READ read
+    #define WRITEFUNC write
+#endif
+
+#ifdef _MSC_VER
+#pragma warning(disable:4996) // don't warn about strerror, sprintf, etc.
 #endif
 
 // gcc 4.4.3 complains bitterly if you don't check the return
 // status from the write() system call. This avoids those
 // warnings and maybe, someday, will catch an error.
 #define WRITE(pipeno, buf, len) \
-   {int status=write((pipeno), (buf), (len)); \
+   {int status=WRITEFUNC((pipeno), (buf), (len)); \
     SimTK_ERRCHK4_ALWAYS(status!=-1, "simbody-visualizer",  \
     "An attempt to write() %d bytes to pipe %d failed with errno=%d (%s).", \
     (len),(pipeno),errno,strerror(errno));}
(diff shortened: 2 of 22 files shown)


SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
```diff
@@ -202,7 +207,11 @@ template <class T> class ClonePtr {
         other.reset(p);
         reset(otherp);
     }
-     
+
+    DEPRECATED_14("use get() instead")
+    const T* getPtr()  const  { return get(); }
+    DEPRECATED_14("use upd() instead")
+    T* updPtr() { return upd(); }
```
chrisdembia: In doxygen, there will be no indication that the method is deprecated. Instead, it will look like the method is undocumented but will be around indefinitely.

What about using the macro to **also** insert a doxygen comment for the method? In that case, even when C++14 rolls around, we will still want to use our macro. If you think that is desirable (to automatically insert doxygen comments to duplicate the compiler warning), then we could also remove the `_14` from the macro name.

Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
```diff
@@ -98,16 +99,22 @@ static void shutdown();
...
+#ifdef _MSC_VER
+#pragma warning(disable:4996) // don't warn about strerror, sprintf, etc.
 #endif
 
 // gcc 4.4.3 complains bitterly if you don't check the return
 // status from the write() system call. This avoids those
 // warnings and maybe, someday, will catch an error.
 #define WRITE(pipeno, buf, len) \
-   {int status=write((pipeno), (buf), (len)); \
+   {int status=WRITEFUNC((pipeno), (buf), (len)); \
```
chrisdembia: What do you think about `#undef`ing this at the end of the file?

chrisdembia:
So `_CRT_SECURE_NO_DEPRECATE` didn't work out? [OpenSim has this in its root CMakeLists.txt](https://github.com/opensim-org/opensim-core/blob/master/CMakeLists.txt#L271):

```
if(WIN32)
add_definitions(-D_CRT_SECURE_NO_DEPRECATE)
else(WIN32)
endif(WIN32)
```

SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
sherm1: > What about using the macro to also insert a doxygen comment for the method?

What would you think about instead hiding deprecated things from doxygen? Is there a good reason to keep documenting them if we don't want people to use them?

I did consider a permanent `SimTK_DEPRECATED` also. Another reason for that might be if we want to alter the reported message. Currently I am inserting "DEPRECATED:" before the supplied message because Microsoft just outputs `Warning C4996` which doesn't exactly scream "deprecated". But ... I'm hoping for nicer behavior when they implement the standard. And I'm not a big fan of macros. It made me very happy to get rid of the `OVERRIDE_11` macro.

Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
sherm1: > What do you think about #undefing this at the end of the file?

Not needed because it isn't a header file. Everything it defines is lost at the end of the compilation unit.

sherm1:
> So _CRT_SECURE_NO_DEPRECATE didn't work out?

Surprisingly, I couldn't get that to work reliably. Its behavior is very touchy. I *think* it would work OK for internal code if added to CMake, but we have header files that use the disfavored methods. When those are included in user programs *they* will get stupid warnings if they don't also manage to get that odd macro working. I feel that is an awkward imposition on users and that we ought to take care of our own warnings. `#define`-ing that symbol (or the related one `_CRT_SECURE_NO_WARNINGS`) at the top of header files doesn't work. So the only thing I could think of that is reliable is the `#pragma warning(push/pop)` stuff.

It will be interesting to see what happens with OpenSim once Simbody is no longer globally disabling this warning.

SimTKcommon/include/SimTKcommon/internal/ClonePtr.h
chrisdembia: > What would you think about instead hiding deprecated things from doxygen? Is there a good reason to keep documenting them if we don't want people to use them?

I think hiding a method from doxygen, if it exists, can be very frustrating for users. For many, I think doxygen represents the API. If they can use a method but can't find evidence that it exists in the library  (assuming they are unwilling to look at the source code), they might get very confused, re-download Simbody, reinstall it, etc.

Getting rid of macros is nice, especially for `OVERRIDE_11`, which did not add additional functionality. However, Simbody does uses macros in many places where they *are* useful, and I think this could be such a situation.

I still prefer using the macro to insert doxygen documentation (I don't like public methods without doxygen documentation and I don't like hiding methods from doxygen). Perhaps there could be multiple variants: a temporary `DEPRECATED_14` macro that intends to hold us over until c++14 support, and a `SimTK_DEPRECATED` that will stick around and that inserts doxygen documentation.

I'm also probably biased by my excitement in being able to insert doxygen documentation via macros for the OpenSim Properties. Anyway, having the DEPRECATED_14 macro is an improvement over the previous situation.

By the way, Doxygen has a way to mark a method as deprecated: http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmddeprecated.

Ideally, a future version of doxygen would just pick up the C++14 deprecated attribute and insert documentation for it.

sherm1: > By the way, Doxygen has a way to mark a method as deprecated

I tried that and found it unsatisfactory. All it actually does is create a new paragraph with heading "Deprecated", which is not very useful. Better would be something grayed out, marked with a distinctive logo, or something like that.

sherm1: > I think hiding a method from doxygen, if it exists, can be very frustrating for users.

I think that might be exactly what we are trying to achieve with `[[deprecated]]`. We are trying to tell them not to use the routine, while still allowing them to continue using it if necessary. If they are already using it, they don't need the documentation. If they are thinking about using it, they definitely *should not* do so, so again we shouldn't help them with documentation. Under what circumstance is having the documentation a good thing?

Also there are many public methods and classes that are excluded from doxygen already; we do that when we don't want users to make use of them but have to make them public (or protected) for our own use.

chrisdembia: > If they are already using it, they don't need the documentation.

This is not true when inheriting someone else's code (e.g., going through Jack/Tim's predictive code).

Freeglut is deprecated on macs. Say it were deprecated everywhere in the world, but I still wanted to add the antialiasing and thus needed the documentation on the deprecated methods. (This is a little contrived, since the antialiasing isn't specific to freeglut).

> If they are thinking about using it, they definitely should not do so, so again we shouldn't help them with documentation.

My opinion is we should help them with documentation that says they shouldn't use the method.

The reality of deprecating is that users may have lots of code that depends on deprecated stuff and it may be a difficult task to move to the new methods. They may be on a tight research deadline (well, in this case they shouldn't have moved to the next version of a dependency), and might need to use the deprecated methods for a while and thus would want the documentation.

Having documentation could be helpful throughout the process of moving to the non-deprecated methods. Especially for more complicated changes, like OpenSim moving from `virtual void connectToModel()` to `virtual void extendConnectToModel()`, which presents a subtle opportunity to introduce bugs (if the user forgets to rename `Super::connectToModel()` to `Super::extendConnectToModel()`). I think the doxygen deprecated warning could be very strongly worded. In cases where it's not simply "use this other method instead", the documentation might be necessary to understand how the functionality changed between versions.

sherm1: > This is not true when inheriting someone else's code 

Good point!

sherm1:
If we're going to include these in doxygen, I would like to make sure they are marked "deprecated" in the brief, so that it is clear even if all you see is the quick method summary at the top of the doxygen page. (Doxygen's `@deprecated` doesn't accomplish that.) I've been putting `(Deprecated)` at the beginning of the brief manually. It would be difficult to do that with a macro, especially since the `[[deprecated]]` attribute has odd placement rules (like it must come after `class` or `enum` to deprecate those, but before functions and variables). It would probably be easy enough to handle this with a convention, like:
```c++
/** (Deprecated) The rest of the brief.
What you should do instead of using this method. 
The original methoddocumentation. **/
DEPRECATED_14("what to do instead")
void myFunc();
```
and 
```c++
/** (Deprecated) The rest of the brief. 
What you should do instead of using this class.
The original class documentation. **/
class DEPRECATED_14("what to do instead") MyClass {
};
```
What do you think?

chrisdembia:
> What do you think?

I like it! I agree strongly that the "(Deprecated)" should appear in the brief.

sherm1:
OK, I added back in the doxygen for the deprecated methods. Not sure if this is worthwhile, but it seemed like a good idea to have the **`(Deprecated)`** come out bold in the brief. The only way I found that worked was using html tags like this:
```c++
    /** <b>(Deprecated)</b> Same as `get()`. Use get() instead; it is more like 
    the API for `std::unique_ptr`. **/
    DEPRECATED_14("use get() instead")
    const T* getPtr()  const  { return get(); }
```
which produced this:
image

and also looks good in the detail documentation. Is this a good way to do it?

chrisdembia:
I considered suggesting the bolding but then I thought it might draw undue attention to the deprecated methods in the list of methods at the top. Looks fine though. My eyes actually usually avoid more congested areas, so I probably would avoid the deprecated methods.

sherm1:
> I considered suggesting the bolding but then I thought it might draw undue attention to the deprecated methods in the list of methods at the top.

I think we should move the deprecated methods and classes to the bottom of the list whenever possible. No need to make them unduly prominent!

chrisdembia:
> I think we should move the deprecated methods and classes to the bottom of the list whenever possible. No need to make them unduly prominent!

I agree!

Don't let me hold you up from merging.

sherm1:
OK, thanks for reviewing Chris.

PR #408, opened 2015-07-01
Updated doxygen for SimbodyMatterSubsystem::getNumBodies()

mjhmilla:
Doxygen update in SimbodyMatterSubsystem function getNumBodies():

Sherm and I discussed an update to the doxygen for the function SimbodyMatterSubsystem::getNumBodies() which I found confusing. Now the documentation includes an example so that it is clear that getNumBodies() counts ground as the 1st body, and continues from there. Thus getNumBodies() will return 2 when called on a planar pendulum. It is also mentioned that this counting convention differs from the MobilizedBodyIndex (where Ground is at MobilizedBodyIndex(0)) to prevent any confusion.

### DIFF
--- a/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
+++ b/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
@@ -206,7 +206,7 @@ geometry that can be used to visualize this multibody system. **/
 bool getShowDefaultGeometry() const;
 
 /** The number of bodies includes all mobilized bodies \e including Ground,
-which is the 1st mobilized body, at the 0th MobilizedBodyIndex. (Note: if 
+which is the first mobilized body, at MobilizedBodyIndex 0. (Note: if 
 special particle handling were implemented, the count here would \e not 
 include particles.) Bodies and their inboard mobilizers have the same index 
 since they are grouped together as a MobilizedBody. MobilizedBody numbering 
@@ -217,11 +217,11 @@ you could think of its mobilizer as the Weld joint that attaches it to the
 universe), but otherwise every mobilized body has a unique body and mobilizer. 
 
 Example: 
-  \li Here we use a model of a planar pendulum with 1 degree-of-freedom.
-  \li The command getNumBodies() will return 2.
-  \li Why? Ground counts as a mobilized body, and so does the pendulum.
-  \li But note that Ground is at MobilizedBodyIndex(0), and the pendulum 
-      is at the MobilizedBodyIndex(1)**/
+  - Here we use a model of a planar pendulum with 1 degree-of-freedom.
+  - The command getNumBodies() will return 2  
+  - Why? Ground counts as a mobilized body, and so does the pendulum. 
+  - Note that Ground is at MobilizedBodyIndex(0), and the pendulum is at 
+    the MobilizedBodyIndex(1)**/
 int getNumBodies() const;
 
 /** This is the total number of defined constraints, each of which may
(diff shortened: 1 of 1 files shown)


Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
```diff
@@ -206,15 +206,22 @@ geometry that can be used to visualize this multibody system. **/
...
-which is the 0th mobilized body. (Note: if special particle handling were
-implmemented, the count here would \e not include particles.) Bodies and their
-inboard mobilizers have the same index since they are grouped together as a 
-MobilizedBody. MobilizedBody numbering (using unique integer type
-MobilizedBodyIndex) starts with Ground at MobilizedBodyIndex(0) with a regular
-labeling such that children have higher indices than their parents. Ground
-does not have a mobilizer (or I suppose you could think of its mobilizer as 
-the Weld joint that attaches it to the universe), but otherwise 
-every mobilized body has a unique body and mobilizer. **/
+which is the 1st mobilized body, at the 0th MobilizedBodyIndex. (Note: if
```
sherm1: "first" would be better here to avoid using the numeric value that might look like an index. And then "at MobilizedBodyIndex 0 for more contrast".

Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
```diff
@@ -206,15 +206,22 @@ geometry that can be used to visualize this multibody system. **/
...
+include particles.) Bodies and their inboard mobilizers have the same index 
+since they are grouped together as a MobilizedBody. MobilizedBody numbering 
+(using unique integer type MobilizedBodyIndex) starts with Ground at 
+MobilizedBodyIndex(0) with a regular labeling such that children have higher 
+indices than their parents. Ground does not have a mobilizer (or I suppose 
+you could think of its mobilizer as the Weld joint that attaches it to the 
+universe), but otherwise every mobilized body has a unique body and mobilizer. 
+
+Example: 
+  \li Here we use a model of a planar pendulum with 1 degree-of-freedom.
```
sherm1: (This is very nice!) Should use this format:
```
Example:
- Here we use ...
- Blah blah
```
which will format the same way but looks better in the header.

sherm1:
Thanks, Matt! Please make minor mods, push again, and leave a message here when ready (GitHub doesn't notify for subsequent pushes to the same PR).

aymanhab:
@sherm1 Is it worth mentioning that OpenSim does its own accounting of Bodies now that Ground doesn't count as Body on OpenSim master (while it used to in the past?)

mjhmilla:
Hey Sherm - I just updated the doxygen to match your comments.

sherm1:
> Is it worth mentioning that OpenSim does its own accounting of Bodies now that Ground doesn't count as Body on OpenSim master

I think this would add to the confusion for a Simbody user (and this is very deep in the Simbody API). OpenSim users will hopefully not get here.

Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
```diff
@@ -206,15 +206,22 @@ geometry that can be used to visualize this multibody system. **/
...
+indices than their parents. Ground does not have a mobilizer (or I suppose 
+you could think of its mobilizer as the Weld joint that attaches it to the 
+universe), but otherwise every mobilized body has a unique body and mobilizer. 
+
+Example: 
+  - Here we use a model of a planar pendulum with 1 degree-of-freedom.
+  - The command getNumBodies() will return 2  
+  - Why? Ground counts as a mobilized body, and so does the pendulum. 
+  - Note that Ground is at MobilizedBodyIndex(0), and the pendulum is at 
+    the MobilizedBodyIndex(1)**/
```
sherm1: Looks great, thanks Matt. Sorry, just noticed: two of the bullets above are missing periods. Please add those and then I promise I'll merge after that.

sherm1:
Thanks, Matt!

PR #410, opened 2015-07-12
Improve Simbody's smart pointers for consistency and compatibility

sherm1:
Improve smart pointers for consistency with each other and with C++11's smart pointers.

Added NOEXCEPT_11 define to allow "noexcept" specification to compile with VS2013.

Moved namespace-scope comparison functions and swap() overloads to SimTK namespace.

Fixes issue #328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,18 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 3.6 (in development)
 --------------------
 * Added C++11 features to the `SimTK::Array_` container including `std::initializer_list` construction, move construction, move assignment, and `emplace` methods.
+* Prevented copy construction of Array_<T> from Array_<T2> unless T2 is *implicitly*
+  convertible to T. Previously this was allowed if there was any conversion possible
+  even if it was explicit. Array_ was being too relaxed about this, causing hidden 
+  copies to occur. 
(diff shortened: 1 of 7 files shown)


sherm1:
OpenSim team: this PR makes some changes to Simbody 3.6 (master branch) smart pointers that will affect OpenSim, hopefully for the better! Mainly you will notice the deprecated warnings I mentioned in opensim-org/opensim-core#528, because of misuse of `SimTK::ReferencePtr`. The behavior hasn't changed except for the warnings, but there are actual bugs and leaks revealed by them so they have to be fixed carefully.

There are also some behavioral changes: 
- `ClonePtr`'s relational operators like `operator==()` were incorrectly delegating to the contained *object* rather than the contained *pointer*; I think I previously removed any OpenSim dependency on that so it should have no effect.
- All smart pointers now have move constructors and move assignment so many copies can be avoided, especially when growing an `std::vector` or other container with smart pointers in it. 
- Maybe not obviously, the correct move semantics for `ReferencePtr` is to actually move the pointer, not zero it out as is the copy semantics. This is clear if you think about the desired behavior of std::vector<ReferencePtr> when it is growing and has to be reallocated (which gets done by move construction).

/cc @aseth1 @aymanhab @klshrinidhi @chrisdembia @carmichaelong @msdemers @tkuchida

PR #412, opened 2015-07-21
Implemented Parallel Forces, Changes to ParallelExecutor Class

thomasklau:
Passes all tests


sherm1:
Thanks for submitting this PR, Thomas! I have a few preliminary comments:
- I'm having trouble reviewing it because there are a very large number of changes that were just whitespace cleanup (deleting ws at eol mostly). I can get GitHub to ignore those but then it won't let me make line comments. Please submit the ws cleanup separately from the substantive changes to facilitate review.
- A Simbody handle class can contain only a single pointer to the private implementation (PIMPL pattern). So the new mutex can't go in the `State` class but could go in `StateImpl`.

I'll give this a much closer look once the ws stuff is out of the way.

PR #416, opened 2015-07-27
Make `SimTK::Function_` objects cloneable and assignable

sherm1:
Some minor improvements to the `SimTK::Function_<T>` classes:
- Added `clone()` method to `Function_` base class and implemented it for Simbody-defined concrete Function classes.
- Made concrete Function members non-const to permit assignment.
- Modified `Function_<T>::Step` to allow changing its parameters after construction.

@chrisdembia please review.


chrisdembia:
This looks good. Thanks for doing this.

PR #417, opened 2015-07-28
Add `NullOnCopyUniquePtr` and regression tests

sherm1:
Added smart pointer `SimTK::NullOnCopyUniquePtr` which is the same as `std::unique_ptr` except that it adds copy construction and copy assignment that leave the  target null. This can be used for owned heap-allocated objects that are "local" in the sense that they should not be copied along with their owner. That allows use of compiler-generated default copy construction and copy assignment for the owning class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 
 3.6 (in development)
 --------------------
+* Added smart pointer NullOnCopyUniquePtr which is the same as std::unique_ptr
+  except that it adds copy construction and copy assignment that leave the
+  target null. This can be used for owned heap-allocated objects that are
+  "local" in the sense that they should not be copied along with their owner.
+  That allows use of compiler-generated default copy construction and 
+  copy assignment for the owning class. 
 * Added clone() method to `SimTK::Function_` base class and implemented it for
   Simbody-defined concrete Function classes. Made concrete Function members
   non-const to permit assignment, and modified Function_<T>::Step to allow
--- a/SimTKcommon/include/SimTKcommon/basics.h
+++ b/SimTKcommon/include/SimTKcommon/basics.h
@@ -44,6 +44,7 @@
 #include "SimTKcommon/internal/ClonePtr.h"
 #include "SimTKcommon/internal/CloneOnWritePtr.h"
 #include "SimTKcommon/internal/ReferencePtr.h"
+#include "SimTKcommon/internal/NullOnCopyUniquePtr.h"
 #include "SimTKcommon/internal/String.h"
 #include "SimTKcommon/internal/Serialize.h"
 #include "SimTKcommon/internal/Fortran.h"
--- a/SimTKcommon/include/SimTKcommon/internal/NullOnCopyUniquePtr.h
+++ b/SimTKcommon/include/SimTKcommon/internal/NullOnCopyUniquePtr.h
@@ -1,0 +1,136 @@
+#ifndef SimTK_SimTKCOMMON_NULL_ON_COPY_UNIQUE_PTR_H_
+#define SimTK_SimTKCOMMON_NULL_ON_COPY_UNIQUE_PTR_H_
+
+/* -------------------------------------------------------------------------- *
+ *                       Simbody(tm): SimTKcommon                             *
+ * -------------------------------------------------------------------------- *
+ * This is part of the SimTK biosimulation toolkit originating from           *
+ * Simbios, the NIH National Center for Physics-Based Simulation of           *
+ * Biological Structures at Stanford, funded under the NIH Roadmap for        *
+ * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
+ *                                                                            *
+ * Portions copyright (c) 2015 Stanford University and the Authors.           *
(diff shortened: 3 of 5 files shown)


sherm1:
@klshrinidhi, @chrisdembia please review.

One problem with this is that because it inherits constructors from `std::unique_ptr` it won't compile in Visual Studio 2013. Working around that would be a pain though, so I'm thinking this means Simbody 3.6 must require VS2015. OpenSim 4.0 already does so I think that is probably OK. gcc and clang don't seem to have a problem with this. What do you think?

sherm1:
This is related to opensim-org/opensim-core#528.

klshrinidhi:
I only looked at the smart pointer class. It looks great.

sherm1:
@klshrinidhi, another question: do you think I'm specializing `std::hash` correctly? For functions like `std::swap` everyone recommends *not* specializing in the `std` namespace (so ours are in `SimTK::`). But `std::hash` is a class, so lookup is different and I *think* it is better to specialize in `std`.

sherm1:
@chrisdembia, I attempted to get AppVeyor to build this using VS2015 but got this:
image

Can you tell me the magic incantation? Does this indicate an out-of-date CMake?

klshrinidhi:
I think the partial specialization of std::hash is correct. Following are the requirements I could find:
* DefaultConstructible.
* CopyConstructible.
* CopyAssignable.
* Destructible.
* Expression `h(k)` must return a `size_t` where:
  * `h` is an object of `std::hash<NullOnCopyUniquePtr<...>>` .
  * k is an object of `NullOnCopyUniquePtr<...>` and it can be `const` or not. 
* Swappable.
* Provide type aliases:
  * result_type -- size_t.
  * argument_type -- NullOnCopyUniquePtr<...>.

Operator() inherited from base class will work fine because the argument can be upcasted implicitly.
Type aliases may be needed to be strict but it is probably okay too.

sherm1:
> I think the partial specialization of std::hash is correct. 

Thanks, Shrinidhi. Merging `NullOnCopyUniquePtr` now.

sherm1:
@chrisdembia, I figured out how to get VS2015 working on AppVeyor -- I read the instructions!

chrisdembia:
@sherm1 I should have told you this but right now, appveyor only supports VS2015 on its 1-thread machine, which now causes simbody to take nearly an hour to build. We cannot move to VS2015 yet on OpenSim; the build would take way too long. I was going to wait until appveyor supports VS2015 on its faster machines.

sherm1:
Oh, then this merge will cause OpenSim AppVeyor builds to fail. I'll see if I can find a VS2013-friendly workaround.

chrisdembia:
> Oh, then this merge will cause OpenSim AppVeyor builds to fail. I'll see if I can find a VS2013-friendly workaround.

Are you completely sure about this? Is it worthwhile to try to build OpenSim first before going down that route?

sherm1:
> Are you completely sure about this? 

Actually I'm checking with Feodor first to see if we can get fast VS2015 builds -- that seems more productive!

chrisdembia:
wait wait! check your email first!

aymanhab:
Sherm, can you explain why we need to migrate for this specific change? We're using unique ptr in other headers already and building with 2013.

sherm1:
> Sherm, can you explain why we need to migrate for this specific change? 

Yes. It isn't the use of `std::unique_ptr` but the derivation of a new class *from* `std::unique_ptr`. In C++11 it is possible to inherit constructors from a base class but VS2013 doesn't support that feature. The new smart pointer inherits `std::unique_ptr`'s constructors so won't compile on VS2013. The alternative of trying to mimic those constructors is daunting (cue horrifying list of templatized constructors [here](http://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr)) and difficult to maintain since there will be more in C++17.

This is part of trying to sort out the problems in issue opensim-org/opensim-core#528.

chrisdembia:
@sherm1 can I raise the option of using GitHub's "revert" option for this PR now (and also restoring the deleted branch)? The PR is already reviewed, so when AppVeyor supports VS2015 we can re-merge it. Also, if we are considering reverting, we should revert while it's still easy (while nothing else has been merged into master). This pointer is great but it is not urgent; having a functioning OpenSim Appveyor build is more important than fixing the ReferencePtr misuse I think.

We should also update the Simbody README for the move to vs2015. I'll make the PR for this eventually.

sherm1:
> can I raise the option of using GitHub's "revert" option for this PR now 

I won't merge anything else until we hear from Feodor so it will be easy to revert.

aymanhab:
Thanks @sherm1 for the explanation.

aymanhab:
:+1: @chrisdembia for offering this option as I hate to be forced into migration "immediately" (rather than ndown the road) into a newly released version of VisualStudio with many other players in the ecosystem that we need to make sure work (e.g. SWIG, Java, Python, Matlab etc.). I don't know who else is on Windows/VS2013 but would be good to get some input.

sherm1:
Will roll forward rather than revert (PR #419). Should allow new code to be used in OpenSim but with VS2013.

sherm1:
P.S. Feodor says it will be a week before VS2015 is installed on the fast machines.

PR #419, opened 2015-07-28
Switch back to build with VS2013

sherm1:
Added some more constructors to NullOnCopyUniquePtr rather than inheriting them, so we can continue to build with VS2013.

This should allow the new smart pointer from PR #417 to continue to work,  but with VS2013.

/cc @aymanhab @klshrinidhi @chrisdembia


sherm1:
I'm not going to wait for review on this since the build is broken.

PR #421, opened 2015-08-03
[travis] Use Travis-CI's container-based infrastructure

chrisdembia:
I think Travis-CI is trying to get people to move to their container-based infrastructure; build times should be faster there (guaranteed 2 processors and 4 GB RAM compared to load-balanced 1.5 processors and 3 GB RAM).

This PR does *not* contain ccache'ing of the binary objects.


chrisdembia:
The travis tests actually run much slower with this change.

PR #414, opened 2015-07-23
Implemented Parallel Forces, Changes to ParallelExecutor Class

thomasklau:

PR #423, opened 2015-08-04
Fix ThreadLocal so it doesn't try to destruct thread objects from the main thread

### LINKED ISSUE #422: SimTK::ThreadLocal crashes on thread-destruction cleanup
The `ThreadLocal<T>` class tries to maintain several file-scope static variables [here](https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/ThreadLocal.h#L34), and also registers a destructor with pthreads via `pthread_key_create()`. That destructor ([cleanUpThreadLocalStorage()](https://github.com/simbody/simbody/blob/master/SimTKcommon/include/SimTKcommon/internal/ThreadLocal.h#L38)) references those file-scope statics. However, since that destructor doesn't get invoked until pthread end-of-program cleanup, the file-scope statics can get destructed first, causing `cleanUpThreadLocalStorage()` to throw an exception. This was observed in PR #414 when running on Windows (see discussion there).

/cc @thomasklau, @chrisdembia

sherm1:
Fixes #422. Addresses a problem discovered during review of PR #414.

@thomasklau, @chrisdembia please review. I marked the class as deprecated in the class documentation but I didn't make it generate warnings. That should wait until Thomas reimplements it in terms of `thread_local` and then we can replace all uses of it in Simbody to avoid those warnings.

I got rid of all the unnecessary "static" infrastructure the class had before.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/ThreadLocal.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ThreadLocal.h
@@ -9,9 +9,9 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
  *                                                                            *
- * Portions copyright (c) 2008-12 Stanford University and the Authors.        *
+ * Portions copyright (c) 2008-15 Stanford University and the Authors.        *
  * Authors: Peter Eastman                                                     *
- * Contributors:                                                              *
+ * Contributors: Michael Sherman                                              *
  *                                                                            *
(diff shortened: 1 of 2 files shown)


chrisdembia:
I'll let @thomasklau review this. I don't believe I am equipped.

sherm1:
> I'll let @thomasklau review this. 

Any objections to/comments on this, @thomasklau?

thomasklau:
Looks good to me.

thomasklau:
We should be replacing it soon in the next PR also.

sherm1:
OK, thanks Thomas.

PR #426, opened 2015-08-13
Add ResetOnCopy<T> and ReinitOnCopy<T> mixins

sherm1:
Add ResetOnCopy<T> and ReinitOnCopy<T> mixins. These are used on data members to provide resetting to default values on copy construction or copy assignment, or reinitialization to specified values.
These can be used on built-in types, pointers, containers, smart pointers, Simbody resource index objects, and almost anything else. (I don't know how to make them work with C array types though.)

Also:
- Added a clear() method to unique index types.
- Added templates that define all relational operators when given only operator==() and operator<().

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
@@ -121,17 +121,18 @@ public:
 
     /** Move constructor is simply a pass-through to the move constructor of
     the contained object so behaves normally. **/
-    ResetOnCopy(ResetOnCopy&& source) : Super(std::move(source)) {} // default
+    ResetOnCopy(ResetOnCopy&& source) 
+    :   Super(static_cast<Super&&>(source)) {} // default
 
     /** Copy assignment reinitializes this object to its default-constructed
     condition; the source argument is ignored. **/
     ResetOnCopy& operator=(const ResetOnCopy& ignored) 
-    {   Super::operator=(ignored); return *this; }
+    {   Super::operator=(static_cast<const Super&>(ignored)); return *this; }
 
     /** Move assignment is simply a pass-through to the move assignment of the
     contained object so behaves normally. **/
     ResetOnCopy& operator=(ResetOnCopy&& source) 
-    {   Super::operator=(std::move(source)); return *this; }
+    {   Super::operator=(static_cast<Super&&>(source)); return *this; }
 
     /** Assignment from an object of type `T` uses `T`'s copy assignment
     operator if there is a suitable copy assignment operator available. **/
--- a/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
@@ -46,15 +46,15 @@ class ResetOnCopyHelper {};
 
 /** Ensures that a data member of type `T` is automatically reset to its default
 value upon copy construction or copy assignment.\ This allows the
-compiler-generated defaults for these to be used.
+compiler-generated default copy methods to be used.
 
 Here are some usage examples:
 @code
 class Thing {
-    // Able to use default constructors and assignments.
+    // Able to use default or initializing constructors, but note that any 
+    // initial values shown here disappear when this is copied; all members will
+    // be default-initialized in the copy.
 
-    // Note that any initial values shown here disappear when this
-    // is copied; all members will be default-initialized in the copy.
     ResetOnCopy<int>                     m_defint;
     ResetOnCopy<char>                    m_charZ = 'z';
     ResetOnCopy<string>                  m_defstr;
@@ -63,6 +63,11 @@ class Thing {
     ResetOnCopy<short>                   m_shArr[3] = {9,8,7};
     ResetOnCopy<SubsystemIndex>          m_subIx{5};
     ResetOnCopy<std::vector<string>>     m_vstr {"one", "two", "three"};
+
+    // Caution: not a smart pointer; will be cleared but not deleted.
+    ResetOnCopy<const int*>              m_ptr;
+
+    // Must use a smart pointer like this for automatic delete on copy.
     ResetOnCopy<unique_ptr<Array_<int>>> m_up{new Array_<int>({1,2,3})};
 };
 @endcode
@@ -86,15 +91,16 @@ assignment will use the default constructor. `ResetOnCopy<T>` adds
 default-constructed condition.
 
 @tparam  T  Template type that is a numeric, character, or pointer built-in
-            type, or a class type that is DefaultConstructible and has an 
-            accessible destructor. Enum and array types are not allowed.
+            type, or a class type that is DefaultConstructible and  
+            `Destructible`. Enum and array types are not allowed.
 
-@see ReinitOnCopy if you want to reset to a non-default value.
+@see ReinitOnCopy if you want to reset to a non-default value or use an enum.
 **/
 template <class T> 
 class ResetOnCopy : public ResetOnCopyHelper<T, std::is_scalar<T>::value> {
     using Super = ResetOnCopyHelper<T, std::is_scalar<T>::value>;
 
+    /** @cond **/ // These confuse doxygen.
     // TODO: should be possible to specialize this for arrays.
     static_assert(!std::is_array<T>::value, 
         "ResetOnCopy<T> does not allow T to be an array. Use an array "
@@ -112,6 +118,7 @@ class ResetOnCopy : public ResetOnCopyHelper<T, std::is_scalar<T>::value> {
 
     static_assert(std::is_destructible<T>::value,
         "ResetOnCopy<T> requires type T to have an accessible destructor.");
+    /** @endcond **/
 
 public:
     using Super::Super;
--- a/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
+++ b/SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
@@ -92,11 +92,26 @@ default-constructed condition.
 @see ReinitOnCopy if you want to reset to a non-default value.
 **/
 template <class T> 
-class ResetOnCopy 
-:   public ResetOnCopyHelper<T, 
-            std::is_scalar<typename std::remove_all_extents<T>::type>::value> {
-    using Super = ResetOnCopyHelper<T, 
-            std::is_scalar<typename std::remove_all_extents<T>::type>::value>;
+class ResetOnCopy : public ResetOnCopyHelper<T, std::is_scalar<T>::value> {
+    using Super = ResetOnCopyHelper<T, std::is_scalar<T>::value>;
+
+    // TODO: should be possible to specialize this for arrays.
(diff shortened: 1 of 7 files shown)


sherm1:
Some usage examples from the regression tests:

*Reset*OnCopy
```c++
// Member initializations here are "one time"; if this is copied the copy
// had all members default constructed instead. No overhead.
class UsesResetOnCopy {
public:
    ResetOnCopy<int>                     defint;  // gets zero-initialized
    ResetOnCopy<Body*>                   bodyptr; // gets nullptr-initialized
    ResetOnCopy<char>                    charZ = 'z';
    ResetOnCopy<string>                  defstr;  // gets default-initialized
    ResetOnCopy<string>                  strHelloC = "hello"; 
    ResetOnCopy<short>                   shArr[3] = {9,8,7};
    ResetOnCopy<SubsystemIndex>          subIx{5};
    ResetOnCopy<std::vector<string>>     vstr {"one", "two", "three"};
    ResetOnCopy<unique_ptr<Array_<int>>> up{new Array_<int>({1,2,3})}; // won't leak
};
```
*Reinit*OnCopy
```c++
// Member initializations here are remembered; copy construction using this as
// the source will result in the copy being restored to these values even if the source
// was changed. Overhead: an extra object of type T holds the initial value.
class UsesReinitOnCopy {
public:
    enum Color {Red, Green, Blue};

    // Default construction is not allowed for ReinitOnCopy.
    ReinitOnCopy<char>                   charZ = 'z';
    ReinitOnCopy<string>                 strHelloC = "hello";
    ReinitOnCopy<short>                  shArr[3] = {9,8,7};
    ReinitOnCopy<SubsystemIndex>         subIx{5};
    ReinitOnCopy<std::vector<string>>    vstr {"one", "two", "three"};
    ReinitOnCopy<Color>                  color{Blue};
};
```

For both classes move construction and assignment behaves normally. All methods are inherited from type T (including constructors and assignment operators) and there are default conversions to these members can be used as though they were T's rather than `ResetOnCopy<T>`'s.

sherm1:
(please see earlier comments)
/cc @aseth1, @aymanhab, @klshrinidhi, @chrisdembia, @tkuchida, @msdemers, @thomasklau

chrisdembia:
Cool stuff.

1. Do you plan to remove `NullOnCopyUniquePtr`?
2. What is the reason for different behavior between copy construction and copy assignment?
3. I was planning to use these classes for Simbody handles. The OpenSim Model dynamically allocates a few Simbody handles (MultibodySystem, SimbodyMatterSubsystem, Force::Gravity, ...) and I don't see a `reset()` or `clear()` on these classes; at least sometimes there is a `clearHandle()`. We can alternatively dynamically allocate the handles and use them with `ResetOnCopy`; just seems a little redundant though.

sherm1:
> Do you plan to remove NullOnCopyUniquePtr?

Yes. These can also be used instead of ReferencePtr, but I am not sure which is better.

> What is the reason for different behavior between copy construction and copy assignment?

I'm not sure what you mean. They both perform reset or reinit. Please clarify.

> I was planning to use these classes for Simbody handles. 

I'm planning to add clear() methods to all handles if these classes look promising.

chrisdembia:
> I'm not sure what you mean. They both perform reset or reinit. Please clarify.

The doxygen comment fro `ResetOnCopy` says:
```
For class types `T`, copy construction will use the 
default constructor, and copy assignment will invoke `T::reset()` or, if there 
is no such method then `T::clear()`.
```

Isn't this different behavior? The copy assignment could also use the default constructor.

sherm1:
The test failure (only on OSX) appears to be a bug in the Apple std::string library's implementation of move assignment. This program demonstrates the problem:
```c++
#include <string>
#include <iostream>
#include <utility>
using std::cout; using std::endl;

int main() {
    std::string s2("goodbye");
    cout << "s2=" << s2 << endl;

    s2 = s2;
    cout << "s2=s2 -> s2=" << s2 << endl;
    s2 = std::move(s2);
    cout << "s2=move(s2) -> s2=" << s2 << endl;
}
```
gcc, Linux clang, and VS2015 agree that the self-move shouldn't change the string's value. On OSX only (at least for clang 3.5 there), the string gets wiped out by the self move. I'll have to fix the test for now; does anyone have a later OSX compiler to check if this has been fixed?

chrisdembia:
> Apple std::string library's implementation of move assignment.

Perhaps it's a difference between libstdc++ and LLVM's libc++.

chrisdembia:
Output on clang 3.6 (Apple LLVM version 6.1.0) on OSX 10.10.3:

```
s2=goodbye
s2=s2 -> s2=goodbye
s2=move(s2) -> s2=
```

chrisdembia:
On Ubuntu, using clang 3.6 and libc++ instead of libstdc++, the bug still exists. So I think it's libc++.

sherm1:
> Isn't this different behavior? The copy assignment could also use the default constructor.

Ah, I see. It can't use the default constructor for `this` because the object already exists. If it default constructs a value it would have to copy assign that value to `this`, but `ResetOnCopy<T>` can't assume that T has a copy assignment operator (std::unique_ptr, for example, does not). 

But now that you've said that, I'm wondering if it would be reasonable to invoke the object's destructor and then use "placement new" to default construct it again in place?

sherm1:
> On Ubuntu, using clang 3.6 and libc++ instead of libstdc++, the bug still exists. So I think it's libc++.

Thanks. I wonder how to file a bug report for that.

chrisdembia:
Hmm...maybe [here](https://llvm.org/bugs/)?

klshrinidhi:
I looked into the question:
**Is it allowed and safe to explicitly call the destructor of an object and default construct the object in its place ?**
Short answer -- Yes.
Long answer -- Like I mentioned in the meeting today, I think the problem is more with objects created on the stack and not so much with objects created on the heap because we have control over lifetime of heap objects. I wanted to check if calling the destructor explicitly releases memory for stack objects, and looks like it doesn't. For heap objects, `delete` operator does two things -- destroy and deallocate memory, and so calling destructor explicitly for heap objects will do only that. For stack objects, compiler inserts code to destroy and deallocate when objects go out of scope. If there is *some* valid object in that location when end of scope is reached, there should be no problem.

Note:- I have not tested this with real code.

sherm1:
> I looked into the question: Is it allowed and safe to explicitly call the destructor of an object and default construct the object in its place?

Thanks, Shrinidhi! That's encouraging. Should work the same for heap and stack objects because the code executing in the copy assignment operator will be the same code. I'm going to try this and see if I can pass all my test cases.

sherm1:
> Hmm...maybe here?

Thanks. Filed LLVM bugreport here: https://llvm.org/bugs/show_bug.cgi?id=24453

I switched to `std::vector` from `std::string` for my test case and found it failed on gcc and clang everywhere. So I switched to `SimTK::Array_` where at least I know the move constructor works right!

sherm1:
I tried the Shrinidhi-approved destruct/construct approach in ResetOnCopy<T> and it seems to work fine on all platforms! Here is what copy assignment looks like now:
```c++
    // Here T is of unknown class type; built-ins are handled more easily elsewhere.
    ResetOnCopyHelper& operator=(const ResetOnCopyHelper&) {
        T* thisT = static_cast<T*>(this); // upcast to base
        thisT->~T();        // destruct the current base object
        new (thisT) T();    // reconstruct it with default constructor
        return *this;
    }
```
So now this class should work on anything T that has a destructor and default constructor -- no need for `reset()` or `clear()` methods. Very cool!

sherm1:
This PR is ready for review if any of you brave souls would like to take a crack at it.

chrisdembia:
> thisT->~T();

Are you sure this won't call `transpose()` when `T = SimTK::Matrix`, etc.?

sherm1:
> Are you sure this won't call transpose() when T = SimTK::Matrix, etc.?

Good question, but I think this is safe. The method you are thinking of is actually named `T::operator~()` not `T::~T()`.

sherm1:
The clang people are very responsive! I already got lots of comments, explaining that the standard library is intentionally noncommittal about self-move behavior:

Comment # 2 on bug 24453 from Marshall Clow (home) 
In general there is no requirement on self-move assignment for the standard library. In particular, there is no requirement that it be a no-op.

Howard Hinnant has written a good explanation here:
http://stackoverflow.com/questions/13127455/what-does-the-standard-library-guarantee-about-self-move-assignment

chrisdembia:
> Good question, but I think this is safe. The method you are thinking of is actually named T::operator~() not T::~T().

Oh okay. I guess `this->~T()` is different from `~this->T()`.

sherm1:
> Oh okay. I guess this->~T() is different from ~this->T().

Yes, but I'm adding some more tests to prove it.

SimTKcommon/tests/TestCloneOnWritePtr.cpp
```diff
@@ -426,6 +432,287 @@ void testNullOnCopyUniquePtr() {
...
+    // self copy of underlying std::vector should do nothing
+    myvec.updT() = myvec.getT();
+    SimTK_TEST(myvec.size()==3 && myvec[2]==3 && &myvec[2]==addr2);
+
+    // But, self-copy assignment should clear (cuz reset-on-copy)
+    myvec = myvec;
+    SimTK_TEST(myvec.empty());
+
+    // Test with a matrix.
+    ResetOnCopy<Mat<2,3>> mat23(1,2,3,
```
sherm1: Matrix tests start here, @chrisdembia.

sherm1:
Did some cleanup and doxygen tweaks. I'm done with these now. Any further comments before I merge this? I haven't removed NullOnCopyUniquePtr but will when it isn't being used any more. (`ResetOnCopy<unique_ptr<T>>` replaces it.)

chrisdembia:
No further comments from me.

chrisdembia:
Might be nice to have an alias `using NullOnCopy<T> = ResetOnCopy<unique_ptr<T>>`.

sherm1:
> Might be nice to have an alias using NullOnCopy<T> = ResetOnCopy<unique_ptr<T>>.

I'm reluctant to hide the object ownership consideration. Having it visible could avoid trouble later. Hopefully we'll get used to thinking:
```c++
    ResetOnCopy<double*>           m_thing; // <<<--- RED FLAG, possible leak
    ResetOnCopy<unique_ptr<double> m_thing; // all good here
```

SimTKcommon/include/SimTKcommon/internal/ResetOnCopy.h
```diff
@@ -0,0 +1,287 @@
...
+    const T& getT() const {return m_value;}
+    T&       updT()       {return m_value;}
+
+private:
+    T   m_value{}; // value initialize; i.e., set to zero
+};
+
+
+/** ResetOnCopy helper class specialization for any type `T` that is not a
+built-in ("scalar") type and that has a reset() or clear() method. This will
```
chrisdembia: I don't think this comment is correct anymore.

sherm1: Thanks Chris. I added a fix to pending PR #427.

PR #427, opened 2015-08-14
Remove ill-conceived NullOnCopyUniquePtr class 

sherm1:
This backs out the `NullOnCopyUniquePtr<T>` class that was added in PRs #417 and #419. It is not needed now since PR #426 provided `ResetOnCopy<unique_ptr<T>>` which does the same thing in a better way.

Don't merge this until OpenSim switches to the new class.


sherm1:
@chrisdembia  nuked OpenSim's use of `NullOnCopyUniquePtr` in some or all of opensim-org/opensim-core#595, opensim-org/opensim-core#586, opensim-org/opensim-core#585. So I'm merging this now.

Issue #370, opened 2015-05-12
Delay mass matrix inversion until Acceleration stage


sherm1:
Currently the inverse mass matrix operator is generated in Dynamics stage if it hasn't been requested earlier. This causes trouble in OpenSim for models for which no mass properties have been provided -- see opensim-org/opensim-core#429. OpenSim needs to realize Dynamics stage in order to compute forces for muscle coloring; this should probably be OK as long as no attempt to calculate accelerations is made.

Normally the inverse mass matrix operator is not used until Stage::Acceleration so I think it would be fine to defer its calculation until then. It is already lazy-evaluated and actually available any time after Stage::Position.

sherm1:
Note that the obscure quantity "MobilizerCentrifugalForces" is calculated unconditionally at Dynamics stage [here](https://github.com/simbody/simbody/blob/master/Simbody/src/RigidBodyNode.cpp#L218), and depends on the Articulated Body Inertias having been calculated. I don't think this is used for anything; consider eliminating it rather than moving it to Acceleration.

sherm1:
Fixed by PR #435 a few months ago.

PR #428, opened 2015-08-21
Allow kinematics evaluation separate from position and velocity stages

### LINKED ISSUE #185: Need separate PositionKinematics and VelocityKinematics stages
Currently Simbody's computation `Stage` granularity lumps all position-dependent quantities together in `Stage::Position`, and all velocity-dependent quantities together in `Stage::Velocity`. It would be useful to separate out the kinematic quantities because they are dependent only on generalized coordinates q (position kinematics) and generalized speeds u (velocity kinematics). They do not depend on time or the presence or absence of `Constraint`s or `Motion`s, and hence the calculations should not be invalidated when time advances or when constraints are added or removed. On the other hand, other Position-level quantities like constraint & motion errors and witness functions can depend on time also.

This will provide a significant speedup when using a semi-explicit or semi-implicit method where time and velocities are updated prior to updating positions. Currently the time update invalidates the position kinematics unnecessarily. This will also permit efficient changes to the set of constraints.

One tricky consequence of this change is that it breaks the strict ordering of stages if `Stage::Time` is left where it is. It may be better to think of time as a special case so that later-stage cache entries can say whether they are invalidated by time changes without imposing a strict ordering.

See also issue #188; these need to be addressed together.

### LINKED ISSUE #188: tree kinematics cache entries not invalidated properly
Simbody uses lazy cache entries to hold the kinematics-only computations at `Stage::Position` (q-dependent) and `Stage::Velocity` (u-dependent). You can see [here](https://github.com/simbody/simbody/blob/master/Simbody/src/SimbodyMatterSubsystemRep.cpp#L388) and [here](https://github.com/simbody/simbody/blob/master/Simbody/src/SimbodyMatterSubsystemRep.cpp#L420) that they are invalidated only when the next earlier stage is invalidated (that is, `Stage::Time` for the tree position cache and `Stage::Position` for the tree velocity cache. They should *also* get invalidated whenever a q or u (resp.) changes.

The reason this hasn't caused trouble is that `realizePosition()` and `realizeVelocity()` unconditionally recalculate the tree cache entries; they really should only recalculate if needed but that would only work if these guys were getting invalidated correctly.

See also issue #185; these need to be addressed together.

sherm1:
This PR adds some flexibility to make state cache entries have more subtle dependencies than just realization stages, then uses that new capability to allow early evaluation and time independence of position and velocity kinematics (that is, mapping generalized coordinates to spatial pose and generalized speeds to spatial velocities).

Now changing time does not invalidate kinematics, and only `Stage::Instance` is required for position kinematics. Velocity kinematics requires `Stage::Instance` plus already-realized position kinematics. New (optional) methods `realizePositionKinematics()` and `realizeVelocityKinematics()` can be used to get these quantities without having to wait until `Stage::Position` or `Stage::Velocity` are done. 

Also added move construction and move assignment for State objects.

Fixes #185. Fixes #188. Will address related #370 (articulated body inertias) in a separate PR.

### DIFF
--- a/SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
+++ b/SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
@@ -34,6 +34,16 @@
 #include <cstdarg>
 
 namespace SimTK {
+ 
+/** This is the type to use for Stage and state variable version numbers that
+get incremented whenever a state value changes. Whenever time or any state 
+variable is modified, we increment the stage version for any stage that gets
+invalidated. We also increment a separate version number for the state variable
+that changes, so that cache entries can have finer-grained dependencies than
+just on whole stages. -1 means "unintialized". 0 is never used as 
+a %StageVersion, but is allowed as a cache value which is guaranteed never to 
+look valid. **/
+typedef long long StageVersion;
 
 /** This class is basically a glorified enumerated type, type-safe and range
 checked but permitting convenient (if limited) arithmetic. Constants look like 
@@ -254,7 +264,8 @@ public:
 class CacheEntryOutOfDate : public Base {
 public:
     CacheEntryOutOfDate(const char* fn, int ln,
-        Stage currentStage, Stage dependsOn, int dependsOnVersion, int lastCalculatedVersion) 
+        Stage currentStage, Stage dependsOn, 
+        StageVersion dependsOnVersion, StageVersion lastCalculatedVersion) 
     :   Base(fn,ln)
     {
         setMessage("State Cache entry was out of date at Stage " + currentStage.getName() 
--- a/SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
+++ b/SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
@@ -40,7 +40,7 @@ get incremented whenever a state value changes. Whenever time or any state
 variable is modified, we increment the stage version for any stage that gets
 invalidated. We also increment a separate version number for the state variable
 that changes, so that cache entries can have finer-grained dependencies than
-just on whole stages. -1 means "unintialized". 0 is never used as 
+just on whole stages. -1 means "uninitialized". 0 is never used as 
 a %StageVersion, but is allowed as a cache value which is guaranteed never to 
 look valid. **/
 typedef long long StageVersion;
--- a/SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
+++ b/SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
@@ -9,7 +9,7 @@
  * Biological Structures at Stanford, funded under the NIH Roadmap for        *
  * Medical Research, grant U54 GM072970. See https://simtk.org.               *
  *                                                                            *
- * Portions copyright (c) 2005-14 Stanford University and the Authors.        *
+ * Portions copyright (c) 2005-15 Stanford University and the Authors.        *
  * Authors: Michael Sherman                                                   *
  * Contributors: Peter Eastman                                                *
  *                                                                            *
@@ -38,6 +38,7 @@ done in a separate internal class. **/
 #include <cassert>
 #include <algorithm>
 #include <mutex>
+#include <array>
 
 namespace SimTK {
 
@@ -153,13 +154,6 @@ SimTK_DEFINE_UNIQUE_INDEX_TYPE(SystemMultiplierIndex);
 /// Unique integer type for Subsystem-local multiplier indexing
 /// @see SystemMultiplierIndex
 SimTK_DEFINE_UNIQUE_INDEX_TYPE(MultiplierIndex);
-
-/// This is the type to use for Stage version numbers. Whenever any state 
-/// variable is modified, we increment the stage version for the stage(s) that
-/// depend on it. -1 means "unintialized". 0 is never used as a stage version, 
-/// but is allowed as a cache value which is guaranteed never to look valid. 
-typedef int StageVersion;
-
 
 /** This object is intended to contain all state information for a 
 SimTK::System, except topological information which is stored in the %System 
@@ -287,6 +281,9 @@ State();
 /// state variables either, except those associated with the Topology stage.
 State(const State&);
 
+/// The move constructor is very fast. The source object is left empty.
+State(State&& source);
+
 /// Copy assignment has deep copy semantics; that is, \c this %State will 
 /// contain a \e copy of the source, \e not a reference into it. This makes the 
 /// current %State contain a copy of the state information in the source %State,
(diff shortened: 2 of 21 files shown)


sherm1:
If anyone wants to review this, please do! Otherwise I'll merge it soon.

SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
```diff
@@ -34,6 +34,16 @@
 #include <cstdarg>
 
 namespace SimTK {
+ 
+/** This is the type to use for Stage and state variable version numbers that
+get incremented whenever a state value changes. Whenever time or any state 
+variable is modified, we increment the stage version for any stage that gets
+invalidated. We also increment a separate version number for the state variable
+that changes, so that cache entries can have finer-grained dependencies than
+just on whole stages. -1 means "unintialized". 0 is never used as
```
tkuchida: Typo: unintialized

SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
```diff
@@ -1089,6 +1118,41 @@ was created. This has no effect on the realization level.
...
+/** (Advanced) Return a reference to the discrete variable information for a 
+particular discrete variable. **/
+SimTK_FORCE_INLINE const DiscreteVarInfo& 
+getDiscreteVarInfo(SubsystemIndex, DiscreteVariableIndex) const;
+
+/** (Advanced) Return a reference to the per-subsystem information in the
+state. **/
+SimTK_FORCE_INLINE const PerSubsystemInfo& 
+getPerSubsystemInfo(SubsystemIndex) const;
+
```
chrisdembia: How do these all show up in doxygen? Ideally they are towards the bottom  or in an advanced group.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
```diff
@@ -158,50 +233,63 @@ class CacheEntryInfo {
     CacheEntryInfo& deepAssign(const CacheEntryInfo& src) {
         assert(src.isReasonable());
```
tkuchida: Perhaps `isReasonable()` could have a more descriptive name? I'm unfamiliar with this method and wasn't able to find its implementation [by searching the GitHub repo](https://github.com/simbody/simbody/search?utf8=%E2%9C%93&q=isReasonable).

sherm1: `isReasonable()` is a private method defined right there in that internal class (CacheEntryInfo). Not for external consumption. It is a Debug-mode-only bug catcher.

SimTKcommon/Simulation/include/SimTKcommon/internal/Stage.h
sherm1: Typo fixed -- thanks.

Simbody/src/SimbodyMatterSubsystemRep.cpp
```diff
@@ -1095,6 +1088,114 @@ int SimbodyMatterSubsystemRep::realizeSubsystemPositionImpl(const State& s) cons
...
+    SBTreePositionCache&    tpc = stateDigest.updTreePositionCache();
+
+    // realize tree positions (kinematics)
+    // This includes all local cross-mobilizer kinematics (M in F, B in P)
+    // and all global kinematics relative to Ground (G).
+
+    // Any body which is using quaternions should calculate the quaternion
+    // constraint here and put it in the appropriate slot of qErr.
+    // Set generalized coordinates: sweep from base to tips.
+    for (int i=0 ; i<(int)rbNodeLevels.size() ; i++)
```
tkuchida: `++i`?

Simbody/src/SimbodyMatterSubsystemRep.cpp
```diff
@@ -1250,6 +1335,141 @@ int SimbodyMatterSubsystemRep::realizeSubsystemVelocityImpl(const State& s) cons
...
+checkValidityOfTreeVelocityCache(const State& state,
+                                 const PerSubsystemInfo& subInfo,
+                                 const CacheEntryInfo&   ce) const {
+    const Stage current   = subInfo.getCurrentStage();
+    const Stage dependsOn = ce.getDependsOnStage();
+    assert(dependsOn == Stage::Instance);
+
+    // We don't need to check explicitly whether Instance stage has been
+    // realized because we're going to check that position kinematics has been
+    // realized and that can't happen unless Instance stages was realized first.
```
tkuchida: stages -> stage

SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
sherm1: > How do these all show up in doxygen? Ideally they are towards the bottom or in an advanced group.

Good idea -- I made an advanced group and moved them there. There is still a lot of leftover ugliness in the State doxygen. I don't want to address it now because I have to merge these changes into 4.0 where many other State changes are happening. But at least with an advanced group I'm not making it worse. Thanks.

tkuchida:
Just curious -- why does the caller need to know the number of extra dependencies in the cache (e.g., [here](https://github.com/simbody/simbody/pull/428/files#diff-b341fd521502a66d1d757c06a662fd52R729)? Doesn't the cache know this about itself? I guess I missed something.

Simbody/src/SimbodyMatterSubsystemRep.cpp
sherm1: Typo fixed.

sherm1:
> Just curious -- why does the caller need to know the number of extra dependencies in the cache (e.g., here? Doesn't the cache know this about itself? 

Good question. I made this intentionally redundant because the new mechanism is a house of cards. There is no way to independently verify the length of a C++ array so I wanted to get confirmation that the caller is consistent with the allocation of that cache entry. It would be very easy to make a mistake in that call. If I weren't worried about performance I would use a sized container like `std::vector` (still redundant) but I was having trouble getting the compiler to inline code where I wanted it. I've measured these cache-access methods and they get called **a lot** and have a measurable performance effect if they don't get inlined.

Simbody/src/SimbodyMatterSubsystemRep.cpp
sherm1: > ++i?

That would be ideal, but sometimes discretion is the better part of valor. That is very old code and there are many almost-identical doubly-nested loops written identically in that code (the spacing isn't to-spec either). I did notice those and considered hunting them down and cleaning them all up but it seemed like a distraction from the current project. I'm going to leave those for another day.

tkuchida:
> I made this intentionally redundant because the new mechanism is a house of cards.

Good idea. Thanks for the clarification. `</review>`

sherm1:
>  `</review>`

Thank you! Much appreciated, @tkuchida . You too, @chrisdembia.

PR #432, opened 2015-09-02
Allow state cache entries to specify explicit prerequisites (including tree kinematics)

sherm1:
This is a complete do-over of PR #428 (that fixed issues #185 and #188). The method I used there turned out not to be general enough to handle the needs of issue #370 and anticipated rigid contact needs. It was also very difficult to use correctly. So I tossed that out and implemented what I think is a much better method. (This still doesn't deal with issue #370's articulated body inertias but that will now be a small follow-up PR.)

In this implementation, I have added the concept of "prerequisites" for cache entries, and each cache entry has a boolean `isUpToDateWithPrerequisites` (initially false). Prerequisites can be any state variable (q,u,z or discrete) and any "earlier" cache entry. Now every state variable and cache entry maintains a `DependentsList` that records all the cache entries that have declared that state variable or cache entry as a prerequisite. Whenever a state variable is modified, or a cache entry is invalidated, all the dependents are notified (recursively) so that all dependents get their `isUpToDateWithPrerequisites` flags set to false. When a cache entry is explicitly marked up to date, the flag gets set true.

The new scheme has better performance properties than the old one. In particular it is optimized for speed of access, with costs pushed to the point where variables are updated rather than when cache entries are accessed, which happens *much* more often. There is much less code involved in checking whether a cache entry is up to date, and I used some force-inlines and examined the VS2015 generated assembly code to make sure that `getCacheEntry()` is entirely inlined.

This was much trickier to implement that the previous hack. I added a lot of testing and debugging infrastructure: "advanced" methods in State that are useful for testing; detailed value version checking that only occurs in Debug builds; and a bunch of test cases for esoteric behavior. 

This code ought to be broken up now into separate files but I don't want to do that yet because I have to merge this with my Simbody 4.0 branch which also has State changes.

If anyone is willing to take another look at some or all of this for review, please do.

(@klshrinidhi, in case you're feeling bad about DataTable code that got implemented beautifully and then tossed: I had to completely flush PR #428's code *after* it got reviewed and merged :cry:. Oh well. I like the new code better and I did learn some things from the earlier attempt.)

### DIFF
--- a/SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
+++ b/SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
@@ -159,6 +159,10 @@ class StateImpl;
 class PerSubsystemInfo;
 class DiscreteVarInfo;
 class CacheEntryInfo;
+class DependentList;
+
+using CacheEntryKey = std::pair<SubsystemIndex,CacheEntryIndex>;
+using DiscreteVarKey = std::pair<SubsystemIndex,DiscreteVariableIndex>;
 
 /** This object is intended to contain all state information for a 
 SimTK::System, except topological information which is stored in the %System 
@@ -666,7 +670,7 @@ State -- don't delete the object after this call!
 @see allocateLazyCacheEntry(), isCacheValueRealized(), markCacheValueRealized() **/
 inline CacheEntryIndex 
 allocateCacheEntry(SubsystemIndex, Stage earliest, Stage latest,
-                   AbstractValue* value, unsigned numExtras=0) const;
+                   AbstractValue* value) const;
 
 /** This is an abbreviation for allocation of a cache entry whose earliest
 and latest Stages are the same. That is, this cache entry is guaranteed to
@@ -674,8 +678,8 @@ be valid if its Subsystem has advanced to the supplied Stage or later, and
 is guaranteed to be invalid below that Stage. **/
 inline CacheEntryIndex 
 allocateCacheEntry(SubsystemIndex sx, Stage stage, 
-                   AbstractValue* value, unsigned numExtras=0) const
-{   return allocateCacheEntry(sx, stage, stage, value, numExtras); }
+                   AbstractValue* value) const
+{   return allocateCacheEntry(sx, stage, stage, value); }
 
 /** This is an abbreviation for allocation of a lazy cache entry. The 
 \a earliest stage at which this \e can be evaluated is provided; but there is no
@@ -687,10 +691,21 @@ invalidated in the State.
 @see allocateCacheEntry(), isCacheValueRealized(), markCacheValueRealized() **/
 inline CacheEntryIndex 
 allocateLazyCacheEntry(SubsystemIndex sx, Stage earliest, 
-                       AbstractValue* value, unsigned numExtras=0) const {
-    return allocateCacheEntry(sx, earliest, Stage::Infinity, 
-                              value, numExtras);
+                       AbstractValue* value) const {
+    return allocateCacheEntry(sx, earliest, Stage::Infinity, value);
 }
+
+/** (Advanced) Allocate a cache entry with prerequisites other than just
+reaching a particular computation stage. Possible prerequisites are continuous
+variables q, u, and z; any discrete variable; and any "upstream" cache entry
+whose `earliest` stage is no later than this one's `earliest` stage. **/
+inline CacheEntryIndex
+allocateCacheEntryWithPrerequisites
+   (SubsystemIndex, Stage earliest, Stage latest,
+    bool q, bool u, bool z,
+    const Array_<DiscreteVarKey>& discreteVars,
+    const Array_<CacheEntryKey>& cacheEntries,
+    AbstractValue* value);
 
 /** At what stage was this State when this cache entry was allocated?
 The answer must be Stage::Empty, Stage::Topology, or Stage::Model. **/
@@ -725,18 +740,6 @@ without getting an exception thrown.
 @see allocateCacheEntry(), markCacheValueRealized(), getCacheEntry() **/
 inline bool isCacheValueRealized(SubsystemIndex, CacheEntryIndex) const;
 
-/** (Advanced) Check whether cache entry value is up to date, dealing with
-extra dependencies. The number of extras must match the number provided
-when the cache entry was allocated. If the cache value is up to date, its
-stage version is returned for the convenience of higher-level computations
-that might have a dependency on this one. Otherwise the stage version is
-returned with an invalid value.
-@see markCacheValueRealizedWithExtras() **/
-inline bool isCacheValueRealizedWithExtras
-   (SubsystemIndex, CacheEntryIndex,
-    unsigned numExtras, const StageVersion* extraVersions,
-    StageVersion& cacheValueVersion) const;
-
 /** Mark the value of a particular cache entry as up to date after it has
 been recalculated. This %State's current stage must be at least the
 \a earliest stage as supplied when this cache entry was allocated, and
@@ -751,18 +754,10 @@ a state variable on which it depends.
 @see allocateCacheEntry(), isCacheValueRealized(), getCacheEntry() **/
 inline void markCacheValueRealized(SubsystemIndex, CacheEntryIndex) const;
 
-/** (Advanced) Mark a cache entry up to date after recalculating it and
-include version information for extra dependencies. The number of extras must 
-match the number provided when the cache entry was allocated. 
-@see isCacheValueRealizedWithExtras() **/
-inline void markCacheValueRealizedWithExtras
-   (SubsystemIndex, CacheEntryIndex,
-    unsigned numExtras, const StageVersion* extraVersions) const;
-
 /** (Advanced) Normally cache entries are invalidated automatically, however 
 this method allows manual invalidation of the value of a particular cache entry.
-After a cache entry has been marked invalid here, isCacheValueRealized() or
-isCacheValueRealizedWithExtras() will return false.
+After a cache entry has been marked invalid here, isCacheValueRealized() will 
+return false.
 @see isCacheValueRealized(), markCacheValueRealized() **/
 inline void markCacheValueNotRealized(SubsystemIndex, CacheEntryIndex) const;
 /**@}**/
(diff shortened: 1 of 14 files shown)


SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
```diff
@@ -1122,35 +1117,56 @@ was created. This has no effect on the realization level.
...
-inline StageVersion getUStageVersion() const;
+inline StageVersion getUValueVersion() const;
 /** (Advanced) Return a StageVersion for z, meaning an integer that is
 incremented whenever any z is changed (or more precisely whenever z or y is
 returned with writable access). Will be 1 or greater if z has been
 allocated. **/
-inline StageVersion getZStageVersion() const;
-
-/** (Advanced) Return a reference to the cache entry information for a 
+inline StageVersion getZValueVersion() const;
```
chrisdembia: Why do the methods say `ValueVersion` but they return a `StageVersion` (and the comment says `StageVersion`)?

chrisdembia:
I like this idea a lot.

Seems like you have two names for similar things: prerequisites and dependents. Could you use only one of the two names? I like dependents but it sounds more generic than prerequisites; I don't know which I like more.

SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
sherm1: I'm now using the same data type (`StageVersion` == `long long`) both to mark stages (the historical use for that type) and the values of prerequisites. The method name now makes sense but the return type is weird. Any ideas for fixing it? I could make a ValueVersion type and pretend it is different from StageVersion; I could get rid of the typedef altogether and use `long long` instead; I could switch to some kind of noncommittal type name like `GenerationNumber` or `VersionIdentifier` or ???

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
```diff
@@ -35,6 +35,100 @@ contain any user-visible objects. **/
...
+depends-on stage becoming invalid. For this to work, the depends-on stage
+of a downstream cache entry must be the same as the highest depends-on stage of 
+any of it upstream cache prerequisites. That way it is guaranteed to be 
+implicitly invalidated whenever any of its upstream prerequisites are.
+
+The dependents are expected to know their prerequisites so that they can
+remove themselves from any DependentLists they are on when deleted; and so 
+that they can be registered on the appropriate DependentLists when copied.
+*/
+class DependentList {
```
chrisdembia: In your PR description you suggested this class was named `DependentsList`. `DependentList` is slightly confusing since "dependent" can be either a noun or an adjective (while "dependents" is a noun). What about "PrereqList"?

sherm1:
> Seems like you have two names for similar things: prerequisites and dependents. 

Actually those are opposites. A "prerequisite" is an independent thing that might change. A "dependent" is some downstream computation that must be re-evaluated whenever one of its prerequisites changes.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
```diff
@@ -126,130 +234,115 @@ class DiscreteVarInfo {
...
+between, then validity is determined as follows:
+  - compare the saved depends-on stage version with the curent version; if they
+    don't match then the cache entry is not valid
+  - then, the cache entry is valid if its up-to-date-with-prerequisites
+    flag is set and invalid otherwise.
+
+Cache entries also have an allocation stage (Topology, Model, or Instance)
+and must be de-allocated if that stage is invalidated.
+*/
+class SimTK_SimTKCOMMON_EXPORT CacheEntryInfo {
```
chrisdembia: Why export? Isn't this an internal type? DiscreteVarInfo is not exported.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
```diff
@@ -260,36 +353,102 @@ class CacheEntryInfo {
...
-    AbstractValue& updValue() {assert(m_value); return *m_value;}
+    // Merely handing out the cache entry's value with write access does not
+    // trigger invalidation of dependents. (Maybe it should, but currently it
+    // gets done often with no intent to modify, esp. by SBStateDigest.)
+    // So be sure that the cache entry gets invalidated first either by an
+    // explicit prerequisite change notification, or because the depends-on
+    // stage got invalidated.
+    AbstractValue& updValue(const StateImpl& stateImpl) {
+       assert(m_value); 
+       return *m_value;
```
chrisdembia: Doesn't this violate that references can't be null?

chrisdembia: Feel free to ignore this comment; I don't think it's so important.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
sherm1: > DependentList is slightly confusing since "dependent" can be either a noun or an adjective (while "dependents" is a noun).

How about `class ListOfDependents`? (It is definitely not a list of prerequisites; rather, anything that can be a prerequisite owns one of these `ListOfDependents` objects so it can invalidate them explicitly when it changes.)

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
sherm1: > Why export? Isn't this an internal type? DiscreteVarInfo is not exported.

It's because of several routines here that are not inlined, like `registerWithPrerequisites()`. Those are in the DLL. DiscreteVarInfo is entirely inline. I could move the export to each of the non-inlined methods but it seems easier this way. I've currently got worse problems with the force-inline stuff on gcc & clang (different problems on each).

SimTKcommon/tests/StateTest.cpp
```diff
@@ -98,81 +97,258 @@ using namespace SimTK;
...
 
-    // Although the cache entry *could* be valid at this point,
+    // Allocate a cache entry at Instance stage that has Time as depends-on
+    // and also has a cross-subsystem dependency on discrete variable 
+    // dvx0ModelPos.
+    const CacheEntryIndex cx1InstanceTime = 
+    s.allocateCacheEntryWithPrerequisites(Sub1, Stage::Time, Stage::Velocity,
+        false, false, true, // depends on z
+        {DiscreteVarKey(Sub0,dvx0ModelPos)}, 
+        {CacheEntryKey(Sub0,cx0TopoModel)},
```
chrisdembia: Just tried using an initializer for an `std::vector<std::pair<int, std::string>>` and it worked, so you could probably do `{ {Sub0, dvx0ModelPos}, {Sub0, cx0TopoModel} }`. Though what you've written is clearer I think, and the std::pair initializer makes the code somewhat cryptic.

klshrinidhi:
Thanks. But I don't agree.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
sherm1: > Doesn't this violate that references can't be null?

I think this is OK. `m_value` is a pointer. In Debug the pointer is being checked to see that it is non-null. Only then does it get dereferenced. It's true that in Release builds you might get a null reference but lots of bad things can happen in Release builds as a result of various assertions and validity checks being left out; that's a concession to performance. Code really only needs to follow *all* the rules in Debug builds.

chrisdembia:
> Actually those are opposites. A "prerequisite" is an independent thing that might change. A "dependent" is some downstream computation that must be re-evaluated whenever one of its prerequisites changes.

Oops. Maybe something like dependent and dependee could make the relationship clearer for people like me.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
chrisdembia: +1

SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
chrisdembia: I think "generation" connotes a heavy expensive thing. I like introducing ValueVersion and pretending it's different.

sherm1:
> Thanks. But I don't agree.

With what?

chrisdembia:
> Actually those are opposites. A "prerequisite" is an independent thing that might change. A "dependent" is some downstream computation that must be re-evaluated whenever one of its prerequisites changes.

I think I was confusing "dependent" with "dependency". Could you replace "prerequisite" with "dependency"?

chrisdembia:
Oy, that thing happened where the words lost their meaning for me.

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
```diff
@@ -35,6 +35,100 @@ contain any user-visible objects. **/
...
+  - upstream cache entries. 
+
+When the prerequisite's value changes or is explicitly invalidated, the
+notePrerequisiteChange() method is invoked, causing the invalidate() method of 
+each of the dependent cache entries to be called.
+
+CAUTION: upstream cache entries may be invalidated implicitly by their
+depends-on stage becoming invalid. For this to work, the depends-on stage
+of a downstream cache entry must be the same as the highest depends-on stage of 
+any of it upstream cache prerequisites. That way it is guaranteed to be
```
tkuchida: of it -> of its

sherm1:
> I think I was confusing "dependent" with "dependency". Could you replace "prerequisite" with "dependency"?

One of the dictionary definition for "dependency" is "a dependent or subordinate thing, especially a country or province controlled by another". So it isn't a good word to use as an antonym to "dependent"! Prerequisite: "a thing that is required as a prior condition for something else to happen or exist" -- exactly the right meaning! "Dependent" isn't as perfect. Words with some of the flavor I'm trying to capture with "dependent": client, customer, follower; I don't think they are as good because we happily say "this cache entry depends on that state variable". Other ideas?

SimTKcommon/Simulation/include/SimTKcommon/internal/StateImpl.h
sherm1: > of it -> of its

Thanks, Tom. Fixed in [this commit](https://github.com/simbody/simbody/commit/c12503798a9a353c083f1db4ca851da78d13fcad) along with another problem I noticed while fixing it.

tkuchida:
> Other ideas?

Consumer? I like `class ListOfDependents` more, though.

chrisdembia:
I don't know what's wrong with me but I'm really struggling with the word "dependency" / "dependent". I like the idea of consumer or something like that. I know this is silly but it sounds more "nouny" to me.

chrisdembia:
"input" and "output" is also very clear.

SimTKcommon/Simulation/include/SimTKcommon/internal/State.h
sherm1: > I like introducing ValueVersion and pretending it's different.

Done. Looks a lot better -- thanks!

sherm1:
I went with `class ListOfDependents`. So there are still prerequisites and dependents but I think "list of dependents" is much clearer than "dependent list" which can be parsed numerous ways. Also, only "prerequisite" is used in the user-directed part of the API (that is, some cache entries can specify prerequisites). The other term is only used internally.

sherm1:
When the builds succeed I'm going to merge this unless someone says not to.

PR #433, opened 2015-09-04
Changes to allow swig 3.0.5 to process certain Simbody headers.

chrisdembia:
The changes I made here are based on discussions in https://github.com/opensim-org/opensim-core/issues/592.

I chose to remove the template class typedefs necessary, but I left the ones that I didn't need to change (with perhaps 1 or 2 exceptions). Also, I chose *not* to remove any typedefs for Mat.h; instead I put the typedefs in the public section. I justify this because Vec.h has similar typedefs that are public. By making only the necessary changes, I hope to have minimized the chance of introducing a bug.

Also, for computation within enums, I chose to use `#ifndef SWIG` rather than resorting to a `constexpr` function in order to make sure the C++ code remains clear and that I don't introduce a bug.

I've tested OpenSim wrapping against these changes, and both the python and java wrapping compile (and the python tests and testContext both pass). You can see my changes to OpenSim here: https://github.com/opensim-org/opensim-core/tree/swig-simtk

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/common.h
+++ b/SimTKcommon/include/SimTKcommon/internal/common.h
@@ -782,7 +782,9 @@ template<> struct Is64BitHelper<false>
 platform, meaning that the size of a pointer is the same as the size of a 
 long long; otherwise it will be FalseType and we have a 32-bit platform meaning
 that the size of a pointer is the same as an int. **/
-static const bool Is64BitPlatform = sizeof(size_t) > sizeof(int);
+// We use a constexpr function to avoid a bug in SWIG.
+constexpr bool detect64BitPlatform() { return (sizeof(size_t) > sizeof(int)); }
+static const bool Is64BitPlatform = detect64BitPlatform();
 typedef Is64BitHelper<Is64BitPlatform>::Result Is64BitPlatformType;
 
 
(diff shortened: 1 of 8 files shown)


SimTKcommon/include/SimTKcommon/internal/common.h
```diff
@@ -782,7 +782,9 @@ template<> struct Is64BitHelper<false>
 platform, meaning that the size of a pointer is the same as the size of a 
 long long; otherwise it will be FalseType and we have a 32-bit platform meaning
 that the size of a pointer is the same as an int. **/
-static const bool Is64BitPlatform = sizeof(size_t) > sizeof(int);
+// We use a constexpr function to avoid a bug in SWIG.
+constexpr bool detect64BitPlatform() { return (sizeof(size_t) > sizeof(int)); }
+static const bool Is64BitPlatform = detect64BitPlatform();
```
chrisdembia: I chose to leave `Is64BitPlatform` as a bool variable to preserve backwards compatibility.

chrisdembia:
@aymanhab

sherm1:
This all looks good, Chris -- thanks! Merging.

chrisdembia:
Thank you Sherm.

PR #434, opened 2015-09-05
Add in missing overrides.

chrisdembia:
I used clang-modernize to do this.

@sherm1 do you want me to remove third party code from the changes (IpOpt, tinyxml, is CPodes third party?)?

### DIFF
--- a/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
+++ b/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
@@ -190,12 +190,12 @@ public:
     const Vec3& getPoint() const {return point;}
 
     // virtuals
-    DecorativePointRep* cloneDecorativeGeometryRep() const {
+    DecorativePointRep* cloneDecorativeGeometryRep() const override {
         DecorativePointRep* DGRep = new DecorativePointRep(*this);
         return DGRep; 
     }
 
-    void implementGeometry(DecorativeGeometryImplementation& geometry) const {
+    void implementGeometry(DecorativeGeometryImplementation& geometry) const override {
         geometry.implementPointGeometry(getMyPointHandle());
     }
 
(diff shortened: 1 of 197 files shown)


sherm1:
Wow! I had no idea there were still so many missing overrides. I'm very happy to see that code modernized. I think doing it for the 3rd party stuff is fine too. If we replace any of that we'll do it wholesale anyway rather than try to do it line by line.

Merging with 4.0 might be a little awkward though -- I'll have to try it.

chrisdembia:
> Merging with 4.0 might be a little awkward though -- I'll have to try it.

I can submit a separate PR for the 4.0 branch. You shouldn't merge manually. Instead, we should run the automated process on both branches (I think).

chrisdembia:
I just did the same thing for develop-4.0 and I have a branch in my fork containing the changes (override-4.0). I get 13 conflicting files when trying to merge override-4.0 with this branch (override). If you were to merge override-4.0 into develop-4.0, then you'll still have about 13 files with conflicts, but one side of the diff won't have overrides. If you merge override-4.0 into override (or vice versa), both sides of the diff have overrides. So if I submit a PR for override-4.0 it might make the merge slightly less awkward but it doesn't get rid of the conflicts. 

The reason I haven't opened a PR for override-4.0 is that I don't know if the slight benefit of a smoother merge is worth introducing so many changed files into the history.

sherm1:
> he reason I haven't opened a PR for override-4.0 is that I don't know if the slight benefit of a smoother merge is worth introducing so many changed files into the history.

I agree. I'm sure the merges will be easy anyway since the decisions to be made will all be whether or not to add "override". But if it isn't hard to rerun this then I would like to get a big pending merge done first (that already includes override fixes that I did as I spotted them) and do the zillion-file override fix after that.

chrisdembia:
Sure. Let me know if you want me to redo this process after your merge.

sherm1:
> Let me know if you want me to redo this process after your merge.

Chris, this would be a good time to retry this against Simbody master. It is up to date with all changes successfully merged into dev4.0 branch (up to date in GitHub). Then I will attempt to merge the modernized commit to dev4.0 branch manually (in a dummy repo) before merging the PR to master and abort if it's a disaster. But I don't think it will be too bad.

Best to nuke this PR and submit a new one I think.

PR #435, opened 2015-09-09
Defer calculation of articulated body inertias until Acceleration stage

sherm1:
This is a follow-on to PR #432 that uses the new facility introduced there to fix the problem described in issue #370, and opensim-org/opensim-core#429. Now a model with zero mass properties can be realized to `Stage::Dynamics` without causing a singular mass matrix error; that error will not occur unless the model is realized to `Stage::Acceleration` where the mass matrix *must* be inverted to calculate accelerations.

Three more cache entries now use the "prerequisites" feature: 
- CompositeBodyInertias (depends on position kinematics from PR #432; never valid unless explicitly requested), 
- ArticulatedBodyInertias (also depends on position kinematics; automatically valid at `Stage::Acceleration`), and 
- ArticulatedBodyVelocity (depends on velocity kinematics from PR #432; automatically valid at `Stage::Acceleration`. 

[Technical note: So in addition to the last two being deferred to `Stage::Acceleration`, all three are now correctly understood as time-independent; changing time does not invalidate them. Any of them can be realized any time after `Stage::Instance`.]

Reviewers: you can probably ignore the changes to `RigidBodyNode_...` files. Those were very ugly and I couldn't resist cleaning them up. The real changes there are just to pass in the right cache entries for the data that got moved around. It might make sense to start with the test cases I added to the now inadequately-named TestMassMatrix regression.

/cc @chrisdembia, @tkuchida, @aymanhab, @aseth1

### DIFF
--- a/SimTKcommon/include/SimTKcommon/Testing.h
+++ b/SimTKcommon/include/SimTKcommon/Testing.h
@@ -560,6 +560,7 @@ private:
      "Test values should NOT have been numerically equivalent at tolerance=%g.",(tol));}
 
 #ifndef SimTK_TEST_SUPPRESS_EXPECTED_THROW
+
 /// Test that the supplied statement throws an std::exception of some kind.
 #define SimTK_TEST_MUST_THROW(stmt)             \
     do {int threw=0; try {stmt;}                \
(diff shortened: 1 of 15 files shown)


sherm1:
Please let me know if you are planning to review this. If I don't hear anything I'm going to "self review" and merge it tonight.

chrisdembia:
I don't plan on reviewing.

aymanhab:
I'm not planning on reviewing.

PR #437, opened 2015-09-09
[README] Noted remedy if VS installed without C++

tkuchida:
It's surprising that Visual Studio 2015 does not support C++ by default. I selected *Typical* during the install, which is what I would expect most people to do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ Windows using Visual Studio
 
 All needed library dependencies are provided with the Simbody installation on Windows, including linear algebra and visualization dependencies.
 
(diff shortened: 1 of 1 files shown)


chrisdembia:
+1. 

http://blogs.msdn.com/b/vcblog/archive/2015/07/24/setup-changes-in-visual-studio-2015-affecting-c-developers.aspx

>  C++ is also very large because of the size of the libraries that we ship across many architectures and their matching PDBs. C++ also requires the installation of a Windows SDK. Many non-C++ developers don't need any of this very large payload and so the overall VS install experience is significantly improved by this change. ... We continue to invest heavily in C++.

tkuchida:
> http://blogs.msdn.com/b/vcblog/archive/2015/07/24/setup-changes-in-visual-studio-2015-affecting-c-developers.aspx

I found that blog, too, and agreed with the comments made by Richie Hindle: "Either force people to make the choice, or install everything for those that don't choose (making it obvious that that's what [sic] happening, by showing the tree with everything checked)."

sherm1:
I made the same installation mistake -- thanks, Tom!

PR #440, opened 2015-10-11
Added the possibility to compile simbody with MinGW for Windows

Gjacquenot:
These commits enable compilation with various versions of MinGW for Windows.

Since, simbody is shipped with Blas and Lapack libraries compiled with specific versions of MinGW, some checks had to be added to verify compatibility.

If it is possible to compile simbody, configuration and compilation run smoothly.
Otherwise, user is asked to provide its own version of Blas and Lapack, or to change to its version of MinGW.

Versions tested:

- MinGW 32 with gcc 4.7.2 and dwarf exception mechanism
- MinGW 64 with gcc 4.9.2 and SJLJ exception mechanism
- MinGW 64 with gcc 5.2.0 and SEH exception mechanism with Blas and Lapack compiled manually


sherm1:
Hi, Guilliaume. Unfortunately the important changes in this PR got mixed in with a bunch of cosmetic changes to hundreds of files. Please submit a PR with just the MinGW changes in it.

We probably would not want to make the trailing blank-removal changes to the master branch now because we would then have a daunting merge process with the 4.0 development branch. Once those branches are merged we could consider it.

chrisdembia:
@Gjacquenot: it seems you are very familiar with git, but you can use `git cherry-pick` to isolate the necesasry changes.

Gjacquenot:
Thanks for the tip

I just have to delete two commits with a `git rebase -i` command

- https://github.com/Gjacquenot/simbody/commit/88e7f4827fee253f1979b8c348acbca184d71991
- https://github.com/Gjacquenot/simbody/commit/49525a942a23d8150c0ee669a1032db681d52018

Unfortunately, I don't have the tools to do it right away.

PR #442, opened 2015-10-16
SimTK_ASSERT was defined twice

tkuchida:


sherm1:
:+1: I'm surprised the compiler even let that slip by. Good catch!

PR #449, opened 2015-11-03
Remove need for link_directories()

chrisdembia:
This PR addresses an issue raised by @elen4 about removing the requirement that users call `link_directories(${Simbody_LIB_DIR})` in their CMake project. This is done using imported targets that are different within Simbody's CMake versus in the SimbodyConfig.cmake.

This PR requires a little clean-up before it's ready for review; I'm opening the PR to test the changes.


chrisdembia:
This PR is now ready for review. I'd appreciate if @elen4 and @Gjacquenot could take a peek. 

- Within Simbody, fake (IMPORTED) targets `lapack`, `blas`, and `pthread` are created that tell other Simbody targets how to link to `lapack`, etc.
- Projects using Simbody would also like to link to these targets, so we create IMPORTED targets in SimbodyConfig.cmke as well, but these IMPORTED targets point to the location of the libraries in the installation rather than in the simbody source repo.
- The `lapack` and `blas` targets are only created if not BUILD(ing)_USING_OTHER_LAPACK.
- The `pthread` target is only created if using MSVC, since we expect MinGW users to have their own pthread.
- I removed FindSimbody.cmake since its content was outdated/incorrect, and the SimbodyConfig.cmake provides all the same functionality and more.
- I tested this by building an example (ChainExample). When trying to run ChainExample, I get an error that libraries cannot be found. However, that's not new behavior; it's just that I didn't edit my PATH.
- I did not test with MinGW.
- I did not test with other lapack/blas (BUILD_USING_OTHER_LAPACK), but I'm fairly confident there would be no issues.

Gjacquenot:
I will test it with MinGW.
I don't think it will impact the compilation with MinGW.

Since this PR talks about compilation, I would like to share some compilation issue I have faced:
I have tried to compile simbody with static version of Lapack and Blas with something like that `-DBUILD_USING_OTHER_LAPACK:PATH=C:\liblapack.a;C:\libblas.a;`

Unfortunalety, I have faced multi definition problems at link time, since lapack and blas are used multiple times: for SimTKcommon and SimTKmath. (`dlamch` for example)

I think the code should use the `PRIVATE` keyword with `target_link_libraries` to prevent this error.

I guess each time a dynamic linking was done with Simbody and Lapack...

Gjacquenot:
I confirm that your branch at commit ad2a0e6ee9f92631370f5480ee18182705f4419f compiles successfully with MinGW

elen4:
Hi Chris, nice trick! :smile:
I tested with MSVC14 (of course) 32 bit and CMake 3.3.1, compiling OpenSim master (without copying Simbody libraries). I tested with "scale -PS" after updating the path and it worked. :+1:

chrisdembia:
Thank you both for testing! :smile: . 

@sherm1, can you merge if you're fine with the changes (e.g., deleting FindSimbody.cmake)?

sherm1:
> @sherm1, can you merge if you're fine with the changes (e.g., deleting FindSimbody.cmake)?

Looks great -- thanks for doing this Chris and for reviewing it Elena and Guillaume. Chris, does this need any change to `README.md`?

chrisdembia:
> Chris, does this need any change to README.md?

Just checked...nope.

chrisdembia:
Thank you!

PR #451, opened 2015-11-04
Added a check on MSVC compiler with C++ 2011 features

Gjacquenot:
This PR adds a CMake check on the MSVC compilation. If C++ 2011 feature like `constexpr` is not supported, we stop configuration. 

I have used MSVC2013 and I only discovered at compilation time that this compiler did not support all C++ 2011 features. This is why I propose this PR

### DIFF
--- a/Platform/CMakeLists.txt
+++ b/Platform/CMakeLists.txt
@@ -19,6 +19,25 @@ if(WIN32 AND NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
  return()
 endif()
 
+# With MSVC, we check that the compiler supports C++ 2011.
+# We can check the value of MSVC_VERSION with a test like this:
+#   if(MSVC_VERSION LESS 1800 OR MSVC_VERSION EQUAL 1800)
+#   endif()
+# However, there is a special version of MSVC 2013 (CTP) that supports C++ 2011
+# features, so we make a test to see if compiler supports C++ 2011.
+if(MSVC)
+    include(CheckCXXSourceCompiles)
+    CHECK_CXX_SOURCE_COMPILES("constexpr int f() { return 0; }\nint main(){f();return 0;}" MSVC_SUPPORTS_CONSTEXPR)
+    if(NOT MSVC_SUPPORTS_CONSTEXPR)
+        message(FATAL_ERROR "\nMSVC does not support C++ 2011 features, for "
+                            "example 'constexpr'. Update to at leaset MSVC 2015 "
+                            "or use a MinGW version on Windows.\nLook at the "
+                            "README.md for more information\nIf you have the"
+                            " 'Visual C++ Compiler Nov 2013 CTP (CTP_Nov2013)'"
+                            " comment this test and configure normally.")
+    endif()
+endif()
+
 if(MINGW)
     # Caution: No cross-compiling is taken into account:
     # If we use a 32 bit compiler, we will get a 32 executable.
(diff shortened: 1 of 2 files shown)


chrisdembia:
Newer versions of cmake have support for conveniently detecting C++ language feature support: https://cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html:

```
target_compile_features(mylib PRIVATE cxx_constexpr)
```
Is there a reason not to use this?

Platform/CMakeLists.txt
```diff
@@ -19,6 +19,25 @@ if(WIN32 AND NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
  return()
 endif()
 
+# With MSVC, we check that the compiler supports C++ 2011.
+# We can check the value of MSVC_VERSION with a test like this:
+#   if(MSVC_VERSION LESS 1800 OR MSVC_VERSION EQUAL 1800)
+#   endif()
+# However, there is a special version of MSVC 2013 (CTP) that supports C++ 2011
+# features, so we make a test to see if compiler supports C++ 2011.
```
chrisdembia: I really like having CMake check the generator (MSVC) version. My personal opinion is to uncomment the above block (and give a FATAL_ERROR if the MSVC version is too low) and comment out the block below.

Platform/CMakeLists.txt
```diff
@@ -19,6 +19,25 @@ if(WIN32 AND NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
...
+if(MSVC)
+    include(CheckCXXSourceCompiles)
+    CHECK_CXX_SOURCE_COMPILES("constexpr int f() { return 0; }\nint main(){f();return 0;}" MSVC_SUPPORTS_CONSTEXPR)
+    if(NOT MSVC_SUPPORTS_CONSTEXPR)
+        message(FATAL_ERROR "\nMSVC does not support C++ 2011 features, for "
+                            "example 'constexpr'. Update to at leaset MSVC 2015 "
```
chrisdembia: leaset -> least

chrisdembia:
Thanks for the update. Another comment...perhaps this stuff belongs somewhere near the top of the top-level CMakeLists? Before making more changes, perhaps wait to see if @sherm1 has other feedback.

sherm1:
> perhaps this stuff belongs somewhere near the top of the top-level CMakeLists? ... see if @sherm1 has other feedback

- I prefer the simple test of MSVC version number (as is there now). I think it is too messy to check individual features and would be hard to maintain as new code gets added. 
- I would like to see this code in the top-level CMakeLists.txt rather than under `Platform/`, as we do for the CMake version for example. That way we specify the toolchain requirements as soon as possible and have them collected in one place.

Platform/CMakeLists.txt
```diff
@@ -19,6 +19,33 @@ if(WIN32 AND NOT CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
...
+    #       if(NOT MSVC_SUPPORTS_CONSTEXPR)
+    #           message(FATAL_ERROR "MSVC does not support C++ 2011 features")
+    #       endif()
+    #   - the value of MSVC_VERSION
+    #       if(MSVC_VERSION LESS 1800 OR MSVC_VERSION EQUAL 1800)
+    #       endif()
+    #
+    # If you use a special version of MSVC 2013 (CTP) that supports C++ 2011
+    # features, so you should comment the following block.
+    if(MSVC_VERSION LESS 1800 OR MSVC_VERSION EQUAL 1800)
```
sherm1: This is good enough -- the comment above about other methods can be removed.

Gjacquenot:
@sherm1 and @chrisdembia : It can be merged now.

sherm1:
Thanks, @Gjacquenot!

PR #452, opened 2015-11-05
Fixed a bug detecting if file is DLL

Gjacquenot:
Before this fix, CMake get_filename_component ext function returns 0.dll
for "lapack-3.5.0.dll", which is not the expected result

This fix proposes to test the end of the filename to see if file is a dll

Below is a CMakeLists.txt that exposes the problem

    cmake_minimum_required(VERSION 2.8.8)
    project(test)
    set(LIBF "lapack-3.5.0.dll")
    
    get_filename_component(LIBF_SUFFIX ${LIBF} EXT)
    if(LIBF_SUFFIX STREQUAL ".dll")
        MESSAGE(STATUS "File is a DLL")
    else()
        MESSAGE(STATUS "File was not detected as a DLL -> PROBLEM with get_filename_component")
    endif()
    
    string(REGEX MATCH "[.][dD][lL][lL]$" isFileADll ${LIBF})
    IF(isFileADll)
        MESSAGE(STATUS "File is a DLL -> OK")
    ELSE()
        MESSAGE(STATUS "File was not detected as a DLL -> It should have been detected such")
    ENDIF()
    
    set(LIBF "lapack-3.5.0.DLL")
    string(REGEX MATCH "[.][dD][lL][lL]$" isFileADll ${LIBF})
    IF(isFileADll)
        MESSAGE(STATUS "File is a DLL -> OK")
    ELSE()
        MESSAGE(STATUS "File was not detected as a DLL -> It should have been detected such")
    ENDIF()

### DIFF
--- a/Platform/CMakeLists.txt
+++ b/Platform/CMakeLists.txt
@@ -137,7 +137,6 @@ set_target_properties(PlatformFiles
 
 foreach(LIBF ${LIB_FILES})
     get_filename_component(LIBF_ROOT ${LIBF} NAME)
-    get_filename_component(LIBF_SUFFIX ${LIBF} EXT)
     set(COPIED_LIBF "${BUILD_BINARY_DIR}/${CMAKE_CFG_INTDIR}/${LIBF_ROOT}")
     if(MINGW)
         set(LIBF_SRC ${LIBF})
@@ -147,7 +146,8 @@ foreach(LIBF ${LIB_FILES})
         file(TO_NATIVE_PATH "${COPIED_LIBF}" LIBF_DEST)
     endif()
 
-    if(LIBF_SUFFIX STREQUAL ".dll")
+    string(REGEX MATCH "[.][dD][lL][lL]$" isFileADll ${LIBF})
+    if(isFileADll)
         # Only the runtime libraries (.dll) are needed next to the
         # executables/tests; import libraries (.lib) are linked through
         # target_link_libraries.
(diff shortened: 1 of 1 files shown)


Gjacquenot:
@sherm1 : This PR fixed a bug I have encountered while giving my Lapack DLL, with a name like 'lapack.3.5.0.dll'. This file was not recognized as DLL and the associated copying target was not generated, making configuration crashes

chrisdembia:
Previously, without MinGW, Simbody would not try to copy over user-supplied lapack/blas (I *believe*). However, it seems that with MinGW, the user-supplied lapack/blas is copied. Is that desired? Perhaps we should unify behavior between MSVC and MinGW, but that could be a separate PR.

This looks good to me.

Gjacquenot:
@chrisdembia : you are right: with MinGW, I want the user-supplied lapack/blas to be copied in the build directory and also installed. This is done to ease the deployment of simbody. If we decide to package simbody, these libraries will be in the installer. With MSVC and user-supplied lapack/blas, I think we should do the same...

By the way, do you know if it is possible to have Appveyor CI with MinGW? It would be really nice to have CI with MinGW, since MSVC 2015 seems really restrivtive to me.

If Appveyor does not provide a MinGW compiler, and maybe we can download a file and unzip the compiler.

Gjacquenot:
@chrisdembia : be careful about mingw configuration ,  you may have to supply a lapack binary

Gjacquenot:
@sherm1 I think this PR can be merged.

sherm1:
Looks good, Guillaume. Merging.

PR #455, opened 2015-11-15
Fix typo in ASM gmx_atomic

j-rivero:
Thanks to Frederic Bonnard <frediz@linux.vnet.ibm.com> in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=768155


sherm1:
Thanks, José! Do you know if Frederic was actually trying to build on PowerPC? Or did he just notice this by inspection?

sherm1:
Oh, I see from the bug report that he actually tried it. I'm amazed that was the only thing that broke!

j-rivero:
It is the good part of contributing to the major Linux distributions, you are getting [a whole range of multiplatform testing](https://buildd.debian.org/status/package.php?p=simbody) and specialized developers sending patches.

Issue #456, opened 2015-11-20
Should add @Gjacquenot's elastic foundation theory test case


sherm1:
@Gjacquenot wrote a test comparing a Hertz sphere with an equivalent Elastic Foundation sphere and measuring the dependency of force with depth. I'm filing this issue with the discussion to date copied in so we have a place to track it.

sherm1:
From Guillaume Jacquenot 11/20/2015:
I have made some convergence check on the problem of a sphere in contact with an infinite plane with ElasticFoundation.
The result was quite surprising to me, since I was expecting a convergence like for a Hertz problem (Each time we double the penetration depth, the force is multiplied by 2**1.5=2.82
But theoretical background is different...
But after checking it is ok.
Doubling the penetration depth of a sphere on a plane result in multiplying the force normal by 4.
I dont know if this should be in simbody repository?

```c++
#include "SimTKsimbody.h"

using namespace SimTK;
using namespace std;

void testForcesIn();
void testForcesIn()
{
    const double radius = 1.0;
    const double initialPenetration = 0.0005;
    const int maxLevel = 10;
    const Real stiffness = 1e9;
    const Real dissipation = 0.0, us = 0.0, ud = 0.0, uv = 0.0, vt = 0.0;
    vector<double> F;
    for (int i=0;i<maxLevel;++i)
    {
        const double penetration = initialPenetration * pow(2.0,(double)i);
        MultibodySystem system;
        SimbodyMatterSubsystem matter(system);
        GeneralContactSubsystem contacts(system);
        GeneralForceSubsystem forces(system);

        PolygonalMesh sphereMesh;
        sphereMesh = PolygonalMesh::createSphereMesh(radius,6);
        const Body::Rigid body(MassProperties(1.0, Vec3(0), Inertia(1)));
        const ContactSetIndex setIndex = contacts.createContactSet();
        const MobilizedBody::Translation mesh(matter.updGround(), Transform(), body, Transform());
        contacts.addBody(setIndex, mesh, ContactGeometry::TriangleMesh(sphereMesh), Transform());
        contacts.addBody(setIndex, matter.updGround(), ContactGeometry::HalfSpace(), Transform(Rotation(-0.5*Pi, ZAxis), Vec3(0.0,penetration-radius,0.0))); // y < 0
        ElasticFoundationForce ef(forces, contacts, setIndex);
        ef.setBodyParameters(ContactSurfaceIndex(0), stiffness, dissipation, us, ud, uv);
[...]

PR #457, opened 2015-11-23
Added test cases for elastic foundation theory

Gjacquenot:
I have added two test cases for the elastic foundation problem of a sphere on a plane (old and new formulation). For this problem, an analytical solution can be computed: the resulting force corresponds to the product of the spherical cap volume of intersection with the stiffness.

For the different penetration values, I have set a tolerance in %. For some penetrations, we are very close to the analytical solution. (Less than 0.5%). For other, tolerance may be higher (15%)

 - For the old formulation, I have understood that we declare the contact property taking into account the properties of the material. It is fine.

 - For the new formulation, we declare material properties and the code computes global stiffness. I would have liked to have the possibility to declare an infinitely rigid ground. I had to split the stiffness between the material, which for this test case is not ideal.
Moreover, I did not understand why with this formulation, I had to declare a thickness. In the old formulation, I did not have to do so...

### DIFF
--- a/Simbody/tests/TestElasticFoundationForce.cpp
+++ b/Simbody/tests/TestElasticFoundationForce.cpp
@@ -49,7 +49,7 @@ void testForces() {
 
     // Create a triangle mesh in the shape of a pyramid, with the
     // square base having area 1 (split into two triangles).
-    
+
     vector<Vec3> vertices;
     vertices.push_back(Vec3(0, 0, 0));
     vertices.push_back(Vec3(1, 0, 0));
@@ -57,14 +57,14 @@ void testForces() {
     vertices.push_back(Vec3(0, 0, 1));
     vertices.push_back(Vec3(0.5, 1, 0.5));
     vector<int> faceIndices;
-    int faces[6][3] = {{0, 1, 2}, {0, 2, 3}, {1, 0, 4}, 
+    int faces[6][3] = {{0, 1, 2}, {0, 2, 3}, {1, 0, 4},
                        {2, 1, 4}, {3, 2, 4}, {0, 3, 4}};
     for (int i = 0; i < 6; i++)
         for (int j = 0; j < 3; j++)
             faceIndices.push_back(faces[i][j]);
 
     // Create the mobilized bodies and configure the contact model.
-    
+
     Body::Rigid body(MassProperties(1.0, Vec3(0), Inertia(1)));
     ContactSetIndex setIndex = contacts.createContactSet();
     MobilizedBody::Translation mesh(matter.updGround(), Transform(), body, Transform());
@@ -76,10 +76,10 @@ void testForces() {
     ef.setTransitionVelocity(vt);
     ASSERT(ef.getTransitionVelocity() == vt);
     State state = system.realizeTopology();
-    
-    // Position the pyramid at a variety of positions and check the normal 
+
+    // Position the pyramid at a variety of positions and check the normal
     // force.
-    
+
     for (Real depth = -0.1; depth < 0.1; depth += 0.01) {
         mesh.setQToFitTranslation(state, Vec3(0, -depth, 0));
         system.realize(state, Stage::Dynamics);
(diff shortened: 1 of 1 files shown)


sherm1:
Thanks very much, Guillaume -- this is great. I'll make a few minor comments on the code. FIrst, answers to your comments above:
> I would have liked to have the possibility to declare an infinitely rigid ground.

Infinite might not be allowed but I think you could achieve the same effect by making Ground much, much stiffer than the contacting ball (like several orders of magnitude). Then the combining rules would make the overall stiffness that of the softer material, essentially treating Ground as infinite.

> I did not understand why with this formulation, I had to declare a thickness. 

The reason is that the EF theory assumes a thin compliant region over an otherwise-rigid body and the stiffness is supposed to increase as you approach the rigid body. The small-penetrations pressure formula looks like
```
          (1-v)E       d
p(d) = -------------- ---     E=Young's modulus, v=Poisson's ratio
       (1+v)(1-2*v)    h
```
This is what the new Simbody model is trying to support so it separates the stiffness from h. Setting h=1 allows you to ignore that as you did.

For larger penetrations into nonlinear materials (such as for knee cartilage), h is more significant:
```
          (1-v)E
p(d) = -------------- ln(1 - d/h)
       (1+v)(1-2*v)
```
We don't have that model implemented but you can see that it would be important to know h separately from the stiffness in that case. We are getting some pressure from our biomechanics users for the latter formula, but they also want to be able to vary h across the mesh. [This paper](http://www.sciencedirect.com/science/article/pii/S1350453304001225) shows how these are used for knee modeling: 
> Bei & Fregly, Multibody dynamic simulation of knee contact mechanics, Medical Engr. & Physics 26:777-789.

Anyway, that's the logic behind specifying h; I'm not sure we are doing anything very useful with it at the moment.

sherm1:
One odd thing I noticed when running this test was that if I further refined the sphere mesh (you had it at 6 I tried up to 8) it made the small-penetration results *much* better but some of the large-penetration results got a little worse, enough to cause a tolerance violation. I would have thought having more elements would always improve things.

sherm1:
@Gjacquenot, you should add yourself to the Authors list in the block at the top of this file.

Simbody/tests/TestElasticFoundationForce.cpp
```diff
@@ -130,9 +130,146 @@ void testForces() {
...
+ *        where
+ *          r is the radis of the sphere
+ *          h the height of the cap. In our case, the penetration
+ *          depth
+ * @note If we want to go further, we can observe that doubling
+ *       the penetration depth results in multiplying the normal
+ *       effort by 4.
+ *       This is different from Hertz theory, where doubling the
+ *       penetration depth results in multiplying
+ *       the normal effort by 2^(3/2)~2.68
```
sherm1: Great comment.

Simbody/tests/TestElasticFoundationForce.cpp
```diff
@@ -130,9 +130,146 @@ void testForces() {
...
+ *
+ */
+void testEffSphereOnPlaneOldFormulation(bool verbose = false);
+void testEffSphereOnPlaneOldFormulation(bool verbose)
```
sherm1: You don't need to do this in two lines--you can put the default argument in the definition.

Simbody/tests/TestElasticFoundationForce.cpp
```diff
@@ -130,9 +130,146 @@ void testForces() {
...
+    // Define some tolerances for each level in %
+    const Real tolerances[6]= {0.15, 0.07, 0.03, 0.02, 0.01, 0.02};
+    for (int i=0;i<maxLevel;++i)
+    {
+        // For each level, penetration is double
+        const Real penetration = initialPenetration * pow(2.0,(Real)i);
+        // Creation of the classical problem
+        MultibodySystem system;
+        SimbodyMatterSubsystem matter(system);
+        GeneralContactSubsystem contacts(system);
```
sherm1: Delete this line -- not needed in this case.

sherm1:
Done reviewing. Ready to merge after minor changes.

Gjacquenot:
I think it is ok now!

sherm1:
Do you not want to be on the Authors list for that file?

sherm1:
This resolves issue #456. Thanks, @Gjacquenot!

PR #458, opened 2016-01-01
Remove usage of NOEXCEPT_11.

chrisdembia:
Leaving in the definition of NOEXCEPT_11 in case users are using it.


sherm1:
Great! I'm very glad to see those gone. Thanks, Chris.

Issue #305, opened 2014-12-19
Need a general way to serialize a State


sherm1:
Simbody's `State` object has the nice feature that it unambiguously holds everything changeable about a given `MultibodySystem`. So if we could serialize it (meaning write it to a string or file), and then read it back later, we could put the system back into the bit-identical condition it was in at the time it was serialized. OpenSim has an immediate application for this -- see opensim-org/opensim-core#149.

I would like to see an Xml serializer and deserializer first, since Simbody has built-in tools for Xml and OpenSim uses Xml. This would make the serialized states easy to process externally and allow humans to examine and change them. We could add additional types later.

The reason this is a somewhat difficult project is that the `State` contains not just time and continuous variables (which are easily serialized) but also arbitrary objects stored as discrete state variables. Those need an abstraction that would allow users to provide (de)serializers for whatever they put in there.

sherm1:
I looked at boost serialization (old school) and the very nice modernized [cereal](http://uscilab.github.io/cereal) package. Cereal looks great for an application where you have control over the whole file format, but not so good for serializing just a single element in someone else's larger file. It is not well designed for incorporating into a library without interfering with the library user's ability to use cereal themselves.

So I propose to work instead with in-memory `SimTK::Xml::Element` objects, which are easily incorporated anywhere in an Xml file. I would add methods something like:
```c++
class SimTK::State {
    /** Return an XML element containing a precise representation of
    the contents of this State, suitable for serialization. **/
    Xml::Element toXmlElement() const;

    /** Set this State to the contents of the given XML element, which
    should have come from deserializing an elemement that was
    originally produced by toXmlElement(). **/
    void fromXmlElement(const Xml::Element&);
};
```
The trickier part is that `AbstractValue` objects (used for DiscreteStateVariable entries in State) have to be augmented with 
- toXmlElement() that delegates to the concrete `Value<T>` which will use a unique tag to identify T and then delegates to T; and 
- fromXmlElement() that uses that tag as a key for a factory that returns a `Value<T>`. Users storing a new type T must specialize corresponding methods, and register with the factory prior to deserialization.

Does that sound workable?
/cc @chrisdembia @aymanhab @aseth1 @klshrinidhi

chrisdembia:
I don't know how hard that is to implement; sounds like OpenSim's Object all over again. Sounds fine to me though.

chrisdembia:
[...]

PR #459, opened 2016-01-20
Minor fixes in support for upcoming state serialization

sherm1:
A few miscellaneous changes needed for issue #305:

- Modify default floating point-to-String conversions to use lossless number of digits.
- Make lack of type-to-String conversion a runtime rather than compile time error to avoid overly strict requirements on types that never actually get serialized.
- Remove some unwarranted code from existing serialization collection.


chrisdembia:
Do these changes warrant a note in the changelog?

sherm1:
> Do these changes warrant a note in the changelog?

Yes, thanks. Fixed.

PR #460, opened 2016-01-21
Convert SimTK::Xml from a class to a namespace

sherm1:
This is needed to untangle the `Xml` and `Array_` classes which will be mutually dependent at the completion of State serialization (issue #305). This is the change that was anticipated by OpenSim PR opensim-org/opensim-core#736.

This change is backwards compatible for people who were following the recommended approach of using the `Xml::Document` typedef rather than the raw `Xml` class. Otherwise it is necessary to change uses of `SimTK::Xml` to `SimTK::Xml::Document`.

There are also some as-yet-unused new serialization methods at the end of `Xml.h`; please ignore for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,12 @@ to ParallelExecutor, mutex state lock.
   digits by default. An attempt to use the default type-to-String conversion 
   when no stream insertion operator is available is now a runtime rather than
   compile time error. [PR #459](https://github.com/simbody/simbody/pull/459).
+* Upgraded SimTK::Xml from a class to a namespace; local classes like Xml::Element
+  were promoted to namespace level within Xml. This was necessary to untangle
+  Array_ and Xml classes which can be mutually dependent. This change is backwards
+  compatible for those following the recommended use of the previously-available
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,12 @@ to ParallelExecutor, mutex state lock.
   digits by default. An attempt to use the default type-to-String conversion 
   when no stream insertion operator is available is now a runtime rather than
   compile time error. [PR #459](https://github.com/simbody/simbody/pull/459).
+* Upgraded SimTK::Xml from a class to a namespace; local classes like Xml::Element
+  were promoted to namespace level within Xml. This was necessary to untangle
+  Array_ and Xml classes which can be mutually dependent. This change is backwards
+  compatible for those following the recommended use of the previously-available
+  Xml::Document typedef instead of Xml directly. Otherwise you will have to change
+  uses of SimTK::Xml to SimTK::Xml::Document. [PR #460](https://github.com/simbody/simbody/pull/460)
```
chrisdembia: In OpenSim, you made a change to the `using` statements; is that necessary or not?

sherm1: Yes, thanks -- fixed.

PR #469, opened 2016-02-03
Add State::isConsistent() method.

chrisdembia:
This method compares if two states have the same number of subsystems, state variables, constraints, etc. I use such a method in OpenSim's StatesTrajectory class ([see PR](https://github.com/opensim-org/opensim-core/pull/730)).

Notes:

1. The function checks that the number of multipliers matches in both states, even though NUDotErr should be the same as NMultipliers. I did this in case the two numbers diverge in the future for some reason.
2. The test is only executed in release mode, since the test relies on the fact that the stage checks do not occur in release.

Questions:

1. Where should the implementation of this method go? It felt wrong to put its implementation in State; should I move the implementation to StateImpl? OR should this method be a static method in State that takes two states, or a free function?

2. Where should I put the function in State.h? I chose to place it between `clear()` and `setNumSubsystems)`; this is how it looks in doxygen:

 image

### DIFF
--- a/SimTKcommon/tests/StateTest.cpp
+++ b/SimTKcommon/tests/StateTest.cpp
@@ -532,10 +532,10 @@ int main() {
     }
 
 
-    // TODO SimTK_START_TEST("StateTest");
+    SimTK_START_TEST("StateTest");
         //SimTK_SUBTEST(testLowestModified);
         SimTK_SUBTEST(testCacheValidity);
         SimTK_SUBTEST(testMisc);
         SimTK_SUBTEST(testConsistent);
-    // TODO SimTK_END_TEST();
+    SimTK_END_TEST();
 }
--- a/SimTKcommon/tests/StateTest.cpp
+++ b/SimTKcommon/tests/StateTest.cpp
@@ -411,6 +411,97 @@ void testMisc() {
     //cout << "after clear(), State s=" << s;
 }
 
+void testConsistent() {
+    State sA;
+    State sB;
+
+    SimTK_TEST(sA.isConsistent(sB)); SimTK_TEST(sB.isConsistent(sA));
+
+    sA.updQ().resize(5);
+    SimTK_TEST(!sA.isConsistent(sB)); SimTK_TEST(!sB.isConsistent(sA));
+    sB.updQ().resize(5);
+    SimTK_TEST(sA.isConsistent(sB)); SimTK_TEST(sB.isConsistent(sA));
+
+    sA.updU().resize(3);
+    SimTK_TEST(!sA.isConsistent(sB)); SimTK_TEST(!sB.isConsistent(sA));
+    sB.updU().resize(3);
+    SimTK_TEST(sA.isConsistent(sB)); SimTK_TEST(sB.isConsistent(sA));
+
+    sA.updZ().resize(8);
+    SimTK_TEST(!sA.isConsistent(sB)); SimTK_TEST(!sB.isConsistent(sA));
+    sB.updZ().resize(8);
+    SimTK_TEST(sA.isConsistent(sB)); SimTK_TEST(sB.isConsistent(sA));
+
+    sA.updQErr().resize(2);
+    SimTK_TEST(!sA.isConsistent(sB)); SimTK_TEST(!sB.isConsistent(sA));
+    sB.updQErr().resize(2);
+    SimTK_TEST(sA.isConsistent(sB)); SimTK_TEST(sB.isConsistent(sA));
+
+    sA.updUErr().resize(7);
+    SimTK_TEST(!sA.isConsistent(sB)); SimTK_TEST(!sB.isConsistent(sA));
+    sB.updUErr().resize(7);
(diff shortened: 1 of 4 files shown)


sherm1:
> Where should the implementation of this method go? It felt wrong to put its implementation in State ...

I like it as part of `State` the way you have it. It is reasonable to do it that way since it only makes use of State API; it doesn't access anything internal so doesn't need to be in `StateImpl`.

> Where should I put the function in State.h? I chose to place it between clear() and setNumSubsystems) ...

I think that's fine. Ideally the whole State doxygen would be reorganized to move the bookkeeping stuff to the bottom in its own `@name` group, but right now I think it is among friends there.

sherm1:
> The test is only executed in release mode, since the test relies on the fact that the stage checks do not occur in release.

I don't think that's a good idea since the code is *allowed* to do those checks in Release; it just doesn't always do so. But it is easily fixed since you can just advance the State to any stage you like -- it doesn't inherently know whether any actual computations have been performed. Simbody's realize(Dynamics) method, for example, performs calculation on state `s` and then invokes `s.advanceSystemToStage(Stage::Dynamics)`. 

Simbody's `StateTest` regression in SimTKcommon/tests does a lot of this. Look [here](https://github.com/simbody/simbody/blob/master/SimTKcommon/tests/StateTest.cpp#L101) for its code that advances each subsystem and then the whole state to a given stage.

chrisdembia:
Oh neat. How about if I run the test once without realizing at all (only in release mode), and then run the test again in both release and debug modes when realized to Instance?

sherm1:
> How about if I run the test once without realizing at all (only in release mode), and then run the test again in both release and debug modes when realized to Instance?

Why do you want to run the un-realized version in Release? It might be more useful to run it in Debug to verify that exceptions are thrown when they should be. But I think it would be OK just to realize it properly and make sure it works. But the documentation should be clear (if it isn't already) about how realized the states must be before you can call the method. 

An alternative would be to find the lowest stage of the two states, and then only do verifications that are allowed up to that stage.

chrisdembia:
Okay I'll test that an exception gets thrown in Debug, and I'll document that the state must be realized to instance.

chrisdembia:
I tried running the test in debug and ran into the issue that matrix views can't be resized (in debug):

```
Test failed due to exception: SimTK Exception thrown at MatrixHelperRep.h:336: 
  Operation 'resize()' allowed only for owners, not views
```

However, my test exploits the ability to resize the matrix views (somehow...) in release mode. This is necessary because I'm trying to avoid using an actual system (to set the number of Q's, U's, etc.). I tried avoiding the resizing by just assigning a new Vector (`sA.updQ() = SimTK::Vector(5, 0.0)`), but this gave a different exception (related to an invalid number of rows in `copyAssign()`). Can you think of other workarounds?

sherm1:
> The test exploits the ability to resize the matrix views (somehow...) in release mode ... Can you think of other workarounds?

Oh yeah, I saw that but forgot to mention it. You shouldn't use resize there. (I'm so happy that Debug mode is enforcing things as it should!) Instead just make the calls that actually allocate q's, u's, etc. Check out [`testMisc()`](https://github.com/simbody/simbody/blob/master/SimTKcommon/tests/StateTest.cpp#L355) in the StateTest regression; it does lots of those allocations and no System is necessary.

chrisdembia:
I've updated this PR; ready for review again.

SimTKcommon/tests/StateTest.cpp
```diff
@@ -411,6 +411,113 @@ void testMisc() {
...
+        int subsysIdx, int size) {
+    state.invalidateAll(stage);
+    (state.*allocfun)(SubsystemIndex(subsysIdx), Vector(size));
+    advanceStage(state, stage);
+}
+// For QErr, UErr, ZErr, and EventTriggers.
+template <typename IDX, typename ...ARGS>
+void allocate(State& state, Stage stage,
+        IDX (State::*allocfun)(SubsystemIndex, ARGS...) const,
+        int subsysIdx, ARGS... args) {
```
sherm1: Nice! I like these helper functions. Great use of C++11 goodies to shrink code!

SimTKcommon/tests/StateTest.cpp
```diff
@@ -425,9 +532,10 @@ int main() {
     }
 
 
-    SimTK_START_TEST("StateTest");
+    // TODO SimTK_START_TEST("StateTest");
```
sherm1: Why is that TODO now?

sherm1:
Sorry for the delayed review, Chris -- I've been busy! This looks ready to merge except possibly for the TODOs in the test program outer block.

chrisdembia:
> This looks ready to merge except possibly for the TODOs in the test program outer block.

Oops, I forgot to remove those. Removing those lines allows gdb/lldb to catch the exceptions and give me a backtrace. I've now uncommented those lines.

chrisdembia:
Thank you

PR #472, opened 2016-03-03
Add some compiler checks to avoid having compilation crash

Gjacquenot:
This PR adds some compiler checks to avoid having compilation crash.
These checks stops cmake configuration if compilers are too old.
The minimum version number are taken from the Readme.md

This PR comes from a compilation crash that occured with GCC 4.7.2 on Debian

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -506,14 +506,6 @@ include(Dart)
 set(DART_TESTING_TIMEOUT 7200)
 
 if(BUILD_TESTING)
-
-    #if(UNIX AND NOT CYGWIN AND NOT APPLE)
-    #  if(NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE MATCHES Debug)
-    #    add_definitions(-fprofile-arcs -ftest-coverage)
-    #    link_libraries(gcov)
-    #  endif()
-    #endif()
-
     #
     # Testing
     #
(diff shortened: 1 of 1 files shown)


chrisdembia:
Awesome. Could you make the error messages include the version required? To avoid duplicating the version number, it'd be good to create a separate variable to hold the minimum required version.

Also, consider using `VERSION_LESS` instead of `LESS`; see https://cmake.org/cmake/help/v3.0/command/if.html

CMakeLists.txt
```diff
@@ -318,13 +331,13 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
...
-        "-fno-strict-aliasing -fno-tree-vrp -fno-guess-branch-probability")
-      endif()
-    endif()
+    # # We know Gcc 4.4.3 on Ubuntu 10 is buggy.
+    # if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
+    #     if(GCC_VERSION VERSION_EQUAL 4.4)
+    #         set(GCC_OPT_DISABLE
+    #             "-fno-strict-aliasing -fno-tree-vrp -fno-guess-branch-probability")
+    #     endif()
+    # endif()
```
sherm1: Please just delete this code -- no point in leaving it here.

CMakeLists.txt
```diff
@@ -499,8 +512,8 @@ if(BUILD_TESTING)
     #  if(NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE MATCHES Debug)
     #    add_definitions(-fprofile-arcs -ftest-coverage)
     #    link_libraries(gcov)
-    #  endif(NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE MATCHES Debug)
-    #endif(UNIX AND NOT CYGWIN AND NOT APPLE)
+    #  endif()
+    #endif()
```
sherm1: Should this block be deleted rather than fixed?

Gjacquenot:
@chrisdembia , @sherm1 : These 3 new commits take into account your remarks. If you are ok with these, this PR can now be merged.

sherm1:
Looks great -- thanks, @Gjacquenot!

PR #473, opened 2016-03-04
Use RPATH on OSX to find simbody-visualizer, etc.

chrisdembia:
On OSX, we can avoid requiring users to set DYLD_LIBRARY_PATH by baking, into executables, the relative path from the executables to the Simbody libraries.

This is helpful in OpenSim also when using the Simbody visualizer; OpenSim users won't need to modify their DYLD_LIBRARY_PATH to point to Simbody's libraries.

Even if you relocate the Simbody installation (intact), the Simbody visualizer and examples will properly find the libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -299,6 +299,7 @@ if(APPLE AND NOT (${CMAKE_VERSION} VERSION_LESS 2.8.12))
     # into directories that are not already searched by the linker.
     list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES
         "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" isSystemDir)
+    # CMake sets isSystemDir to -1 if the lib dir is NOT a system lib dir.
     if("${isSystemDir}" STREQUAL "-1")
         # This variable is used later on to toggle if RPATH should be set for
         # specific targets.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,10 +21,6 @@ cmake_minimum_required(VERSION ${SIMBODY_REQUIRED_CMAKE_VERSION})
 if(COMMAND cmake_policy)
     cmake_policy(SET CMP0003 NEW)
     cmake_policy(SET CMP0005 NEW)
-    if(NOT (${CMAKE_VERSION} VERSION_LESS 3.0))
-        # MACOSX_RPATH enabled by default; policy introduced with cmake 3.0.
-        cmake_policy(SET CMP0042 NEW)
-    endif()
 endif(COMMAND cmake_policy)
 
 project(Simbody)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -271,6 +267,46 @@ else()
...
+    # OSX. This helps executables find the libraries they depend on without
+    # having to set the DYLD_LIBRARY_PATH environment variable.
+
+    # Consider a library libfoo.dyld and an executable bar.
+    # On OSX, libraries have an "install name" that, when linking, is copied
+    # into the target (e.g., bar). The "install name" can be the full path to
+    # libfoo.dylib, in which case bar will have no trouble finding libfoo.dylib
+    # at runtime (since it has the full path to it). This doesn't work if you
+    # want to be able to relocate your project. Therefore, it's possible to use
+    # special tokens in the "install name" that are evaluated separately. The
```
sherm1: Great comment -- thanks!

CMakeLists.txt
```diff
@@ -271,6 +267,46 @@ else()
...
+
+    # Set the install name of libraries to contain "@rpath".
+    # This allows clients of our libraries to point to them however they wish.
+    set(CMAKE_MACOSX_RPATH ON)
+
+    # We only need to set RPATH in executables if the libraries are installed
+    # into directories that are not already searched by the linker.
+    list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES
+        "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}" isSystemDir)
+    if("${isSystemDir}" STREQUAL "-1")
```
sherm1: Why "-1" here? A comment would be good.

sherm1:
Chris, this looks great. Does it really work? Have you tried moving around the visualizer and some examples?

CMakeLists.txt
chrisdembia: Addressed.

chrisdembia:
Yes it works. That is, if I move around the whole Simbody installation, the simbody-visualizer and examples still find the libraries. There might be a misunderstanding though. If I move the executables but do *not* move the libraries, then the executables will not find the libraries. We could achieve that functionality instead if you prefer, by setting the RPATH in the executables (the `INSTALL_RPATH` target property) to the absolute path of the libraries instead of the relative paths.

A minor caveat is that when copying the whole simbody installation to another location, the examples symlink in `share/doc/simbody/examples/bin` gets removed and the libraries are copied into that location, putting them in the wrong location relative to the libraries; the original copy of the examples in `lib/simbody/examples` still find the libraries though.

chrisdembia:
Ready for review again.

sherm1:
- Does this change need an update to (a) README, and/or (b) CHANGELOG?
- GitHub says there are conflicts now that need to be resolved.

chrisdembia:
I resolved the conflicts, and I updated the README and the CHANGELOG.

My change to the README is vague, because I'm not sure I can provide a precise explanation that is brief. See OpenSim's [more precise explanation](https://github.com/opensim-org/opensim-core#set-environment-variables-1) to see what I mean.

sherm1:
Great -- thanks, Chris.

PR #475, opened 2016-03-10
Implement VisualizerGeometry::implementMeshFileGeometry on Simbody side

aymanhab:
Some background. The cleanest fix in my opinion to this issue is to beef up the protocol to  support passing MeshFile by name and have the Visualizer process load the file,  cache it and never load it again, unfortunately that shifts a significant responsibility to the visualizer and makes it hard to report/handle errors gracefully (since it's a different process). 
Implemented fix is a compromise where the caller to construct DecorativeMeshFile does the search to make sure the file exists and can be opened, then DecorativeMeshFile is instantiated with the full path. DecorativeMeshFileRep lazily loads the file into a PolygonalMesh object, caches it and sends it across the protocol to the visualizer. 
A straightforward approach to do the loading inside implementDecorativeMeshFile call proved expensive since it causes the file to be loaded anew with every frame (even if it's 'Fixed' geometry). 

Please let me know what you think or if you have suggestions for improvement, before spending much time on the code/implementation. Thanks much.

### DIFF
--- a/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
+++ b/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
@@ -567,20 +567,21 @@ class DecorativeMeshFileRep : public DecorativeGeometryRep {
 public:
 // no default constructor
 explicit DecorativeMeshFileRep(const std::string& meshFileName) 
-:   meshFile(meshFileName) {
-    pmesh = nullptr;
+:   meshFile(meshFileName),
+    isInitialized(false) {
 }
 
 const std::string& getMeshFile() const {
     return  meshFile;
 }
-
+// This is logically const but is non-const here to facilitate caching of the mesh 
+// once file is opened.
 const PolygonalMesh& getMesh() {
-    if (pmesh==nullptr) {
-        pmesh.reset(new PolygonalMesh());
-        pmesh->loadFile(meshFile);
+    if (!isInitialized) {
+        isInitialized = true; // We tried to load.
+        pmesh.loadFile(meshFile);
     }
-    return *pmesh;
+    return pmesh;
 }
 // virtuals
 DecorativeGeometryRep* cloneDecorativeGeometryRep() const override {
@@ -595,7 +596,9 @@ void implementGeometry(DecorativeGeometryImplementation& geometry) const overrid
 SimTK_DOWNCAST(DecorativeMeshFileRep, DecorativeGeometryRep);
 private:
 std::string meshFile;
-std::shared_ptr<PolygonalMesh> pmesh;
+PolygonalMesh pmesh;
+bool isInitialized; // Whether we attempted to load the file into pmesh;
+
 // This is just a static downcast since the DecorativeGeometry handle class 
 // is not virtual.
 
--- a/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
+++ b/SimTKcommon/Geometry/src/DecorativeGeometryRep.h
@@ -568,12 +568,20 @@ public:
 // no default constructor
 explicit DecorativeMeshFileRep(const std::string& meshFileName) 
 :   meshFile(meshFileName) {
+    pmesh = nullptr;
 }
 
 const std::string& getMeshFile() const {
     return  meshFile;
 }
 
+const PolygonalMesh& getMesh() {
+    if (pmesh==nullptr) {
+        pmesh.reset(new PolygonalMesh());
+        pmesh->loadFile(meshFile);
+    }
+    return *pmesh;
+}
 // virtuals
 DecorativeGeometryRep* cloneDecorativeGeometryRep() const override {
     DecorativeMeshFileRep* DGRep = new DecorativeMeshFileRep(*this);
@@ -587,7 +595,7 @@ void implementGeometry(DecorativeGeometryImplementation& geometry) const overrid
 SimTK_DOWNCAST(DecorativeMeshFileRep, DecorativeGeometryRep);
 private:
 std::string meshFile;
-
+std::shared_ptr<PolygonalMesh> pmesh;
 // This is just a static downcast since the DecorativeGeometry handle class 
 // is not virtual.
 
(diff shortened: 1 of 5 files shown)


sherm1:
The approach sounds good, Ayman -- I'll comment on the code.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -587,7 +595,7 @@ void implementGeometry(DecorativeGeometryImplementation& geometry) const overrid
 SimTK_DOWNCAST(DecorativeMeshFileRep, DecorativeGeometryRep);
 private:
 std::string meshFile;
-
+std::shared_ptr<PolygonalMesh> pmesh;
```
sherm1: Since PolygonalMesh has shallow copy semantics I don't think this should be a pointer at all. Can't it just be a PolygonalMesh as in DecorativeMesh? You could have a separate "isInitialized" flag for the lazy load.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
```diff
@@ -568,12 +568,20 @@ class DecorativeMeshFileRep : public DecorativeGeometryRep {
...
 explicit DecorativeMeshFileRep(const std::string& meshFileName) 
 :   meshFile(meshFileName) {
+    pmesh = nullptr;
 }
 
 const std::string& getMeshFile() const {
     return  meshFile;
 }
 
+const PolygonalMesh& getMesh() {
```
sherm1: This method will throw an error if the file extension is not recognized. That could be a surprise to someone who doesn't understand that the file is lazy-loaded. This behavior should be documented at the top level since it will be user-visible. Please consider whether it would make sense to validate the file extension when it is first seen rather than waiting until lazy loading -- that could avoid some surprises. But that wouldn't get rid of all possible errors since the file might not be present or readable.

sherm1: Also do you think this method should be const? Its semantics are const but it has a mutable write-once cache. It might be cleaner to use mutable for them. OTOH this works and the top level is const where it matters.

sherm1:
Thanks, Ayman -- looks good pending a few minor things. General comment -- please check the DecorativeMesh and DecorativeMeshFile doxygen to make sure they are still right. Each should mention the other and have an `@see`. The useful individual methods should have a short doxygen comment.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
aymanhab: It is true that PolygonalMesh works as is but it doesn't  or I couldn't find a method to tell if it has been populated/initialized. If you believe a separate flag (isInitialized) is cleaner  I'll add that and remove the pointer.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
aymanhab: For the use case from OpenSim::Mesh, all the path searches, extension checks, and permission checks are performed before the DecorativeMeshFile is instantiated so the control path is relatively safe in that case (unless parsing the file fails or the file is invalid) but as an API that can be used by others in arbitrary contexts, I agree some validation and documentation would be needed. I'll add safety checks and doxygen docs.

I wasn't sure which one is cleaner (const and mutable or neither) I went with neither since mutable seems to turn red flags but I'll do what's consistent with Simbody conventions, so let me know your preference.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
sherm1: > It is true that PolygonalMesh works as is but it doesn't or I couldn't find a method to tell if it has been populated/initialized. If you believe a separate flag (isInitialized) is cleaner I'll add that and remove the pointer.

You could use `PolygonalMesh::getNumVertices()` and assume that if the count is zero then it has not been initialized. That would only be wrong if the mesh file actually contained zero vertices which I suppose it could but that is pathological. In that case it would get reopened repeatedly, but I think that's fine as long as there is a comment pointing it out. A separate flag would be a little tidier.  

Using a pointer here is really just an awkward way of having a separate flag. Pointer==null is being used as the "uninitialized" flag.

SimTKcommon/Geometry/src/DecorativeGeometryRep.h
sherm1: > I wasn't sure which one is cleaner (const and mutable or neither)

I'm happy with either as long as there is a comment explaining the intent. You could leave it as-is and add a comment like `/* This is logically const but non-const here to facilitate caching of the mesh once opened. */`. Or, change it to const and write `/* Although this is const, it saves the mesh in a mutable cache the first time it is called. */`.

aymanhab: Awesome, thanks much.

aymanhab:
Changes done per review, Doxygen changes made and reviewed. Ready for next round.

sherm1:
Looks ready to me, Ayman. An obvious question: have you tested this? I don't think it needs its own regression test but I would like to know that it has been seen to work in real life!

sherm1:
Would be good to test this both for a few cases that work (check that transform gets applied properly, etc.) and cases where the file name is not found or has an unrecognized format, just to make sure the error message gets generated properly and shows up somewhere.

aymanhab:
Thanks @sherm1 Indeed I tested that code (along with a corresponding piece of code on the OpenSim side that generates DecorativeMeshFile(s)) and it worked with no issues. The OpenSim code is on the branch 
https://github.com/opensim-org/opensim-core/compare/implement_MeshFileGeometry

Since the OpenSim side validates that the file exists before creating DecorativeMeshFile, I had to manually remove the file after the check was done in the debugger and it did throw an exception that showed in the console and aborted. It may be better to enclose the loadFile within a try.\/catch to be safe in case the caller didn't do the checks. Let me know your preference.

sherm1:
> It may be better to enclose the loadFile within a try/catch to be safe in case the caller didn't do the checks.

Thanks, Ayman. I don't think that's necessary because the error would have to be reported via exception anyway (no status return). Will merge.

aymanhab:
Awesome, thanks much @sherm1

Issue #477, opened 2016-03-31
Conflict when used with Qt


mitkof6:
When Simbody is included with Qt, there is a conflict with the definitions. Apparently the problem is the definition of qInfo:

Qt:
`#define qInfo QMessageLogger(QT_MESSAGELOG_FILE, QT_MESSAGELOG_LINE, QT_MESSAGELOG_FUNC).info`

Simbody::StateImpl.h
`Array_<ContinuousVarInfo>      qInfo, uInfo, zInfo;`

`2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C2059: syntax error: 'nullptr'
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C3484: syntax error: expected '->' before the return type
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C3613: missing return type after '->' ('int' assumed)
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C3646: 'info': unknown override specifier
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C3551: if a trailing return type is used then the leading return type shall be the single type-specifier 'auto' (not 'SimTK::Array_<SimTK::ContinuousVarInfo,unsigned int>')
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C2071: 'SimTK::PerSubsystemInfo::uInfo': illegal storage class
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(797): error C2071: 'SimTK::PerSubsystemInfo::zInfo': illegal storage class
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(690): error C2660: 'SimTK::PerSubsystemInfo::QMessageLogger': function does not take 3 arguments
2>d:\lib\simbody\install\master_3.6_x64_vs2015\include\SimTKcommon/internal/StateImpl.h(690): error C2228: left of '.info' must have class/struct/union
[...]

PR #478, opened 2016-04-01
Qt conflict with definition of qInfo small

mitkof6:
Regarding issue [#477](https://github.com/simbody/simbody/issues/477) I have renamed **qInfo  -> q_info** in StateImpl.h and State.cpp. Also, for some reason on windows when I use Qt and Simbody there is the header rpcndr.h, which defines small as:

`#define small char`

which is in conflict with SimTKlapack.h. I have renamed **small  -> small1**. 

The tests pass, and I have also tested with Qt.


sherm1:
Dimitar, it looks like some kind of automated formatting tool was on and made a bunch of changes to the code that have nothing to do with the topic of this PR. Can you redo it with just the relevant changes please?

PR #479, opened 2016-04-03
Qt conflict with definition of qInfo and small

mitkof6:
Without auto formatting.

Regarding issue [#477](https://github.com/simbody/simbody/issues/477) I have renamed **qInfo  -> q_info** in StateImpl.h and State.cpp. Also, for some reason on windows when I use Qt and Simbody there is the header rpcndr.h, which defines small as:

`#define small char`

which is in conflict with SimTKlapack.h. I have renamed **small  -> small1**. 

The tests pass, and I have also tested with Qt. The Qt team said that will work on a definition that will undo some of their macros.

### DIFF
--- a/SimTKcommon/include/SimTKlapack.h
+++ b/SimTKcommon/include/SimTKlapack.h
@@ -1279,7 +1279,7 @@ extern int disnan( SimTK_D_INPUT_(din) );
 
 extern int dlaisnan( SimTK_D_INPUT_(din1), SimTK_D_INPUT_(din2) );
 
-extern void dlabad_(SimTK_D_INOUT_(small1), SimTK_D_INOUT_(large));
+extern void dlabad_(SimTK_D_INOUT_(small1), SimTK_D_INOUT_(large));//small -> small1 conflict with definition when Qt is used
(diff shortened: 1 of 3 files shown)


SimTKcommon/include/SimTKlapack.h
```diff
@@ -1279,7 +1279,7 @@ extern int disnan( SimTK_D_INPUT_(din) );
 
 extern int dlaisnan( SimTK_D_INPUT_(din1), SimTK_D_INPUT_(din2) );
 
-extern void dlabad_(SimTK_D_INOUT_(small), SimTK_D_INOUT_(large));
+extern void dlabad_(SimTK_D_INOUT_(small1), SimTK_D_INOUT_(large));
```
sherm1: Needs a comment so no one will accidentally change it back.

sherm1:
Looks good -- ready to merge after adding a comment about the small->small1 change.

mitkof6:
@sherm1 I have added the comments.

sherm1:
Thanks, Dimitar!

Issue #482, opened 2016-04-13
Compiler version check for gcc incorrecly fails


elen4:
I tried to compile Simbody on an Ubuntu 14.04 LTS machine, with gcc 4.8.4 installed, as checked with ``gcc --version``. Although the ``SIMBODY_REQUIRED_GCC_VERSION`` is 4.8.1, the check fails and issues a fatal error, because the version retrieved with the ``gcc -dumpversion`` command is in major.minor fomat, i.e., 4.8 ( https://bugs.launchpad.net/ubuntu/+source/gcc-4.8/+bug/1360404).
Setting ``SIMBODY_REQUIRED_GCC_VERSION`` to 4.8 in the main CMakeLists.txt fixes this issue for me.

chrisdembia:
Thank you for reporting this.

Maybe we should try using the variable `CMAKE_CXX_COMPILER_VERSION` instead of using `gcc -dumpversion`

http://stackoverflow.com/questions/14933172/how-can-i-add-a-minimum-compiler-version-requisite

Could you quickly test if this also fixes the problem for you?

elen4:
Hi Chris, yes, this works for me as I get the full major.minor.patch version. I am using CMake 2.8.12.2, but I think that this variable was introduced in CMake 2.8.9, so I guess you would need to bump up the minimum required version, since it currently is 2.8.8 for non-MSVC compilers

chrisdembia:
Ah okay good point.

chrisdembia:
Thanks for checking.

Gjacquenot:
I have created a pull request to fix this.
I have not updated the CMake version. Maybe @chrisdembia , you want it to be updated?
We would move from 2.8.8 -> 2.8.10 for CMake

chrisdembia:
Ubuntu 14.04 provides cmake 2.8.12.2, and we already require gcc 4.9.1, which Ubuntu 14.04 does not require. So I suppose it would be okay to change the required CMake version.

chrisdembia:
I know this isn't directly related to this issue, but it would be good for us to clearly state the version(s) of Ubuntu that we support in the README.

elen4:
Isn't the required gcc version 4.8.1? This is stated both in the README and in the main CMakeLists, and Ubuntu 14.04 complies with this

chrisdembia:
Oh wow, good catch. Sorry about that. Yes, then the required version of Ubuntu is 14.04. Thanks, @elen4.

[...]

PR #485, opened 2016-04-18
Fixed bug : Compiler version check for gcc incorrecly fails #482 

Gjacquenot:
This bug was reported here:
#482

We use CMAKE_CXX_COMPILER_VERSION, to fetch compiler version

I dont think it is required to update the version of CMake required to configure the project: from Cmake documentation, it says it is already defined

Below are the documentation found

From https://cmake.org/Wiki/CMake_2.8.8_Docs, it says it is defined but internal
  CMAKE_<LANG>_COMPILER_VERSION
       An internal variable subject to change.

       Compiler version in major[.minor[.patch[.tweak]]] format.  This
       variable is reserved for internal use by CMake and is not guaranteed
       to be set.

From https://cmake.org/Wiki/CMake_2.8.10_Docs, it says it is defined
  CMAKE_<LANG>_COMPILER_VERSION
       Compiler version string.

       Compiler version in major[.minor[.patch[.tweak]]] format.  This
       variable is not guaranteed to be defined for all compilers or
       languages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ This is not a comprehensive list of changes but rather a hand-curated collection
 
 3.6 (in development)
 --------------------
+* Forced GCC to be at least 4.9.0, so that new C++11 features like regular
+  expressions can be used (pr #485).
+* Minimum Ubuntu version supported 14.04 LTS (Trusty), with a [manual update of GCC](http://askubuntu.com/questions/466651/how-do-i-use-the-latest-gcc-on-ubuntu-14-04).
+* Recommended minimum Ubuntu version : 15.04 (Vivid), that is shipped with GCC 4.9.2.
 * Added mixin classes `ResetOnCopy<T>` and `ReinitOnCopy<T>` to force default
   construction or reinitialization of data members on copy construction or copy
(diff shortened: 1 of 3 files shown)


sherm1:
:+1:

chrisdembia:
Please also update the README, which lists the minimum required CMake version. Can you also list the minimum required version of Ubuntu, which is now 15.04 (http://packages.ubuntu.com/search?keywords=gcc&searchon=names&suite=all&section=all)?

The appveyor pr failure is unrelated.

sherm1:
@chrisdembia, do you think this deserves a CHANGELOG update also?

chrisdembia:
Yes I do think the CHANGELOG should be updated. Based on conversation in #482, it sounds like the README should be updated to say gcc 4.9 is the minimum required version, and that Ubuntu 14.04 is the minimum supported version of Ubuntu. @Gjacquenot would you be willing to make these changes?

Gjacquenot:
@sherm1 , @chrisdembia : I think these new commits integrate what we have talked about. For me, it can be merged

chrisdembia:
Looks good to me.

sherm1:
Thanks, @Gjacquenot!

Issue #13, opened 2013-08-19
Differentiator should check if sizes were set for gradient & jacobian functions


sherm1:
(Transferred from SimTK.org. Filed 2010-02-03)

(Tim Dorn ran into this problem)

If a user derives a Differentiator function from GradientFunction or JacobianFunction but neglects to call the appropriate base class constructor, the sizes will be passed as "-1" so this mistake is easy to detect. However, despite extensive error checking in Differentiator it doesn't appear to check those sizes for reasonableness before plowing ahead and encountering mysterious disasters.

Those checks should be added, with a helpful error message mentioning that the base class constructors need to be called explicitly.

Kerl1310:
Looking to get involved in some OS projects and I'm interested in helping out with this. Got several years C++ experience in academia and just started a C++ Development job.

sherm1:
Awesome -- welcome, Kyle. @chrisdembia may have other ideas for where you can contribute.

chrisdembia:
Hey Kyle. Simbody is a great project to get involved in! This issue is definitely of the right size to start. If you're interested in C++11, you could add initializer list constructors to Simbody's math types (Vector_, Matrix_, etc.). An idea that is not very glamorous but that could familiarize you with the codebase would be to try to get rid of warnings. We'd like to get rid of as many as possible (when using -Wall, etc.). There are a few issues labeled as "help wanted": https://github.com/simbody/simbody/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

A fun kind of project would be adding support for other optimization libraries to Simbody. We support IPOPT, CFSQP and some others, but we wouldn't mind supporting more libraries, especially SNOPT (though you might have trouble acquiring SNOPT yourself). But I think this would be better to do once you've done a smaller task like this issue.

sherm1:
Here is a related Differentiator issue: #8 ("optimizer limits should restrict differentiator perturbations") that could be a good project. It is more complicated because it requires some API changes and modifications to Differentiator use by optimizers; some careful test cases would be needed.

jozuenoon:
Hi,
I'm making myself familiar with code base as was proposed by @chrisdembia.
My changes are in forked here https://github.com/jozuenoon/simbody. If this changes are in agreement with your standards let me know. I did just small changes in Vector_ and VectorBase to enable braced-init-list to initialize Vector_ values - hopefully this was intended by @chrisdembia in comment above.

chrisdembia:
[...]

PR #491, opened 2016-05-08
Get rid of some compiler warnings

jozuenoon:
As stated #13 I'm making some warnings removal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,8 +347,6 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
         endif()
     else() # not APPLE
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-strict-aliasing")
     endif()
 
     if(inst_set_to_use)


SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
```diff
@@ -799,9 +799,14 @@ template <class R> class NTraits< conjugate<R> > {
     static TNormalize normalize(const T& t) {return TNormalize(t/abs(t));}
 
     // 1/conj(z) = conj(1/z), for complex z.
-    static TInvert invert(const T& t)    
-    {   const typename NTraits<THerm>::TInvert cmplx(NTraits<THerm>::invert(t.conj()));
-        return reinterpret_cast<const TInvert&>(cmplx); } // recast complex to conjugate it
+    // Here reinterpret_cast is source of [-Wstrict-aliasing] warning.
+    // New code is more explicit and should be faster(???).
+    static TInvert invert(const T& t){
+      return static_cast<TInvert>(TReal(1)/t.conj());
```
sherm1: The point of the above method is to conjugate the complex quotient (1/t.conj) merely by reinterpreting it as being of type `conjugate`, rather than performing the conjugation operation, which costs a floating point operation. That's why it does a reinterpret_cast, which is correct. Then the copy constructor is invoked to return the result with no computation.

In the new version, I believe the complex quotient is converted to a conjugate type *of the same value*, which means it is the conjugate of the result rather than the actual result. (That is, this returns with the imaginary part negated from the correct answer.)

So I think you got rid of the warning by making the code wrong! Unfortunately it appears there wasn't a regression test for this -- there should be and you definitely would have to add some tests if you want to mess with this very touchy and performance-sensitive code. It would be great to see some better tests for these operations.

CMakeLists.txt
```diff
@@ -347,6 +347,7 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
         endif()
     else() # not APPLE
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
```
sherm1: Why did the `-std=c++11` go away? Is that not needed now?

SimTKcommon/BigMatrix/include/SimTKcommon/internal/BigMatrix.h
```diff
@@ -1040,7 +1040,9 @@ std::istream& readVectorFromStreamHelper
     openBracket = (char)ch;
     if      (openBracket=='(') {in.get(); closeBracket = ')';}
     else if (openBracket=='[') {in.get(); closeBracket = ']';}
-    else lookForCloser = false;
+    else {lookForCloser = false;
+          closeBracket = 'e'; // Just to get rid of [-Wmaybe-uninitialized] warning.
+    }
```
sherm1: The problem is the violation of RAII on line 1036. Instead, defer the declarations until line 1040 and write something like:
```c++
char openBracket{(char)ch}, closeBracket{};
```
(That syntax is "value initialization" which is zero for numeric types.)

CMakeLists.txt
chrisdembia: In GCC 6.1, the default c++ standard is c++14; maybe that's why? I think this should still have the `-std=c++11`. If you want to enable all warnings, set `CMAKE_CXX_FLAGS` in the CMake GUI / ccmake / command line (`cmake . -DCMAKE_CXX_FLAGS="-Wall"`).

examples/TheoJansenStrandbeest.cpp
```diff
@@ -415,7 +415,7 @@ void UserInputHandler::handleEvent(State& state, Real accuracy,
     }  
 }
 
-
+/*
```
sherm1: Why did you surround this block comment with `/* */`? That is redundant.

SimTKcommon/Geometry/src/PolygonalMesh.cpp
```diff
@@ -321,8 +321,8 @@ void PolygonalMesh::loadVtpFile(const String& pathname) {
     Xml::Element points   = piece.getRequiredElement("Points");
     const int numPoints = 
         piece.getRequiredAttributeValueAs<int>("NumberOfPoints");
-    const int numPolys  = 
-        piece.getRequiredAttributeValueAs<int>("NumberOfPolys");
+    const unsigned int numPolys  =
```
sherm1: Why were all these ints made unsigned? Where was the warning and why couldn't it have been fixed just with a cast somewhere? Making things unsigned is dangerous because simple arithmetic using unsigned doesn't always work. Are you sure the underlying code doesn't use a `-1` as a marker somewhere? If it has any "count down" for loops those won't work with unsigneds. 

Minor nit: we use `unsigned` rather than `unsigned int`.

SimTKcommon/src/gmx_atomic.h
```diff
@@ -263,9 +263,9 @@ static inline int
 gmx_atomic_fetch_add(gmx_atomic_t *     a,
                      volatile int       i)
 {
-    int __i;
+    //int __i;
```
sherm1: No point leaving this dead code around.

examples/ExampleSimplePlanarMechanism.cpp
```diff
@@ -22,7 +22,7 @@
  * -------------------------------------------------------------------------- */
 #include "Simbody.h"
 using namespace SimTK;
-
+/*
```
sherm1: These make the code uglier for no gain.

examples/ExampleWrapping.cpp
```diff
@@ -205,10 +207,10 @@ int main() {
...
     Vec3 Q(r*cos(uQ)*sin(vQ), r*sin(uQ)*sin(vQ), r*cos(vQ));
 
-    Vec3 r_OP = P-O;
-    Vec3 r_IQ = Q-I;
-    Vec3 tP = r_OP.normalize();
-    Vec3 tQ = r_IQ.normalize();
+//    Vec3 r_OP = P-O;
+//    Vec3 r_IQ = Q-I;
+//    Vec3 tP = r_OP.normalize();
+//    Vec3 tQ = r_IQ.normalize();
```
sherm1: Should nuke this code if it is dead.

sherm1:
Reviewed. Comments:
- The purposes of dealing with warnings should be to make the code better. Some of these changes clearly make it worse, like adding unneeded comment delimiters and leaving dead code around in comments. Merely getting rid of warnings is not an improvement.
- `reinterpret_cast` is in the language for a reason -- it can be very useful when used correctly. Whether you can get `-Wall` to like it isn't clear. We depend on type reinterpretation to accomplish effective computations (such as negation, transposition, and conjugation) without having to do any actual floating point instructions.
- the Travis build didn't work.

I do like the idea of getting rid of warnings whenever possible -- if you can do it without making the code worse, please push some fixes to this PR.

SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
jozuenoon: Thank you for explanation! I'm aware about project approach to just reinterpret underlying values for performance sake. As you mentioned conjugate and complex need to be _same type_ to fix this warning, I believe this warning along with optimizer could introduce strange bugs hard to trace.

examples/TheoJansenStrandbeest.cpp
jozuenoon: Line 428, comment which is ending with "\" is generating warning. Quick fix for clarity or it could be ignored.

jozuenoon:
Sure, thanks for comprehensive comments! Underlying purpose for me is to get familiar with codebase. I will make best fixes as you suggested in comments.

examples/TheoJansenStrandbeest.cpp
sherm1: You could convert the whole comment to the `/* */` kind if that would get rid of the warning. That would be an improvement, I think.

SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
sherm1: Why do you think there would be bugs? The whole point of this very old piece of code was to get the optimizer to produce great code that executes quickly. That is one of the best reasons to use C++ over safer languages like Java, and is why things like `reinterpret_cast` are in the language. In this case the warning is presumably that using `reinterpret_cast` is dangerous, which it is. But that isn't the same as saying it shouldn't be used.

You shouldn't touch this code if you don't fully understand it and have regression tests to make sure it still works correctly. What you could do is locally suppress the warning here so that you could still run `-Wall`. Please note that we build `-Werror` so our CI builds fail if there is even a single warning issued anywhere. If you do suppress the warning please do it over as small an are as possible and provide clear documentation about what you're doing and why.

Issue #175, opened 2014-06-24
Ipopt fails when running concurrently in multiple threads 


cpizzolato:
I've already posted this on the forum on SimTK, but then I realised it would make more sense here.

I'm trying to run multiple instances of OpenSim::InverseKinematicsSolver on different threads, but Simbody throws the following error:

```
.*** Error in `./voslParallel': free(): invalid pointer: 0x00007fea108c98c0 ***
Aborted (core dumped)
```

checking with GDB I get 

```
.*** Error in `/home/ceinms/CEINMS/Builds/vosl/src/voslParallel': double free or corruption (out): 0x00007fffd88bf090 ***

Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff0c0e700 (LWP 64447)]
0x00007ffff4fc9f79 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56   ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
```

backtrace

```
(gdb) backtrace 
#0  0x00007ffff4fc9f79 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff4fcd388 in __GI_abort () at abort.c:89
#2  0x00007ffff50071d4 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7ffff5115a10 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff50134ae in malloc_printerr (ptr=<optimized out>, str=0x7ffff5115b40 "double free or corruption (out)", action=1) at malloc.c:4996
#4  _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:3840
#5  0x00007ffff486a658 in Ipopt::LapackSolverInterface::Factorization(int const*, int const*, bool, int) () from /opt/OpenSim/lib/libSimTKmath.so.3.3
#6  0x00007ffff486aada in Ipopt::LapackSolverInterface::MultiSolve(bool, int const*, int const*, int, double*, bool, int) () from /opt/OpenSim/lib/libSimTKmath.so.3.3
[...]

PR #493, opened 2016-05-11
Remove global variables in IpOpt interface - issue #175

cpizzolato:
Removes the global variables in IpLapackSolverInterface.cpp. Makes `ipiv` a member variable of `IpLapackSolverInterface`. IpOpt can now be used in multiple concurrent threads. I have tested this with OpenSim inverse kinematics and with the Simbody test suit without encountering any problem.


chrisdembia:
Do you know if IpOpt has addressed this issue in newer versions of IpOpt?

cpizzolato:
Hi Chris,

the version of IpOpt currently implemented in simbody is pretty old, from the file headers seems from 2006. Comparing it with the latest version (3.12.5) of IpOpt (http://www.coin-or.org/download/source/Ipopt/) is not really straightforward, since the structure is quite different. I've checked the changelog and they mention some fix regarding thread-safety, although on a different file. Also, the file that I've patched no longer exists in the laterst IpOpt implementation.

chrisdembia:
Okay; thank you for looking into that.

SimTKmath/Optimizers/src/IpOpt/IpLapackSolverInterface.cpp
```diff
@@ -146,6 +146,7 @@ double *afact;
     delete [] a;
     delete [] irn_;
     delete [] jcn_;
+    delete [] ipiv_;
```
sherm1: Should set `ipiv_ = nullptr;` here -- the other three are reassigned but ipiv is not.

SimTKmath/Optimizers/src/IpOpt/IpLapackSolverInterface.hpp
```diff
@@ -160,6 +160,7 @@ namespace Ipopt
     //MUMPS data structure
 //    DMUMPS_STRUC_C mumps_data;
 
+    int *ipiv_;
```
sherm1: For consistency, please move this up next to the other data members (below `negevals_`), with a comment like `/** Array for storing the pivot order after factorization. */`.

sherm1:
Claudio, thank you very much for this PR! I'm done reviewing it -- looks great pending two small changes. Please let me know when those are in so we can merge this.

Question: is this actually sufficient to provide thread safety when using IpOpt or are there other problems remaining?

Regarding this version of IpOpt vs. the current one, I think the broken file here might be one we added. I looked at "blame" and it appears to have been created by Jack Middleton who worked on Simbody a long time ago.

cpizzolato:
Sherm, thanks for the review! I'm going to update the PR with the suggested changes. 

I don't know if this modification solves all the thread safety issues since I tested it only through OpenSim. I'll try to build a test-case using just simbody to be sure there are no other problems and I'll get back to you.

Claudio

chrisdembia:
Note: it seems like any other fixes for thread safety could happen in separate PRs.

sherm1:
Agreed -- there is no question that this PR is an improvement. I was just wondering whether this was the end of the thread un-safety story.

sherm1:
I would be very happy to merge this PR after the few minor changes we discussed.

cpizzolato:
I've created a small test based on the example ConstrainedOptimization.cpp where I run multiple identical threads that perform the optimisation at the same time. With the master branch I get a seg fault, as expected. The patch solves the problem and everything seems fine. 

I can adapt the example so it can be added to the ctest if you think it makes sense. If so, I'll create another PR for it.

cpizzolato:
Hi @sherm1, I've added a simple adhoc test based on the Ipopt test in the regression suite. It's basically the same, but it launches 10 optimizations concurrently and then it collects the results. Using the unsafe Ipopt in the master branch you will get a segmentation fault. Using the fixed Ipopt you get the same optimization results 10 times, as expected. 

Let me know if there's anything else I can do or fix reguarding this PR!

Cheers
Claudio

sherm1:
@cpizzolato, thanks very much for this fix! The test case looks great. Sorry for the reviewing delay.

cpizzolato:
@sherm1 thanks for the merge! I'm glad the test case was up to standards!

PR #497, opened 2016-05-31
Add in override keyword to some RigidBodyNodeSpec classes.

chrisdembia:
This PR suppresses some warnings in clang v3.7 related to missing override keywords for RigidBodyNodeSpec classes.


sherm1:
Great -- thanks, Chris! Somehow I thought we had already caught all of those.

chrisdembia:
Yeah I thought so too. Maybe clang's detection is improving.

PR #501, opened 2016-06-03
Throw detailed exception if can't write T to Xml element.

chrisdembia:
This PR introduces an exception if we are unable to write `T` to an Xml element.


sherm1:
Did you mean to submit this to the master branch?

Also there are a bunch of ugly warnings on the Windows build.

chrisdembia:
Oops. I will resubmit and fix that warning.

PR #502, opened 2016-06-03
Throw meaningful exception if T cannot be converted to Xml element

chrisdembia:
This PR now correctly goes into the the `feature_state_serialization` branch, and hopefully does not produce a warning with MSVC.

Could you also review the exception message? It's tailored for developers, not end-users. Perhaps that's not what we want?


chrisdembia:
Thank you for merging (this PR as well as the others).

PR #503, opened 2016-06-05
Add operator>> for Motion::Level enum

chrisdembia:
This PR adds an `operator>>` for enum `Motion::Level`.

This PR also changes the error-handling mechanism used for `Stage`s `operator>>` to use `failbit` instead of an exception. This is more consistent with the rest of the Simbody code base.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/503)

### DIFF
--- a/Simbody/include/simbody/internal/Motion.h
+++ b/Simbody/include/simbody/internal/Motion.h
@@ -231,6 +231,20 @@ particular implementation object. **/
 explicit Motion(MotionImpl* r) : HandleBase(r) { }
 };
 
+/** This operator is used when deserializing a SimTK::Motion::Level enum
+from an Xml element. **/
+inline std::istream& operator>>(std::istream& input, Motion::Level& level)
+{
+    int i;
+    input >> i;
+    switch (i) {
+    case -1: level = Motion::NoLevel;           return input;
+    case  0: level = Motion::Acceleration;      return input;
+    case  1: level = Motion::Velocity;          return input;
+    case  2: level = Motion::Position;          return input;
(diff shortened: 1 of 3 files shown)


Simbody/include/simbody/internal/Motion.h
```diff
@@ -231,6 +231,20 @@ particular implementation object. **/
...
+from an Xml element. **/
+inline std::istream& operator>>(std::istream& input, Motion::Level& level)
+{
+    int i;
+    input >> i;
+    switch (i) {
+    case -1: level = Motion::NoLevel;           return input;
+    case  0: level = Motion::Acceleration;      return input;
+    case  1: level = Motion::Velocity;          return input;
+    case  2: level = Motion::Position;          return input;
```
sherm1: Consider writing these out as readable names (like you did for Stage).

sherm1:
Reviewed. Just one comment.

chrisdembia:
I have updated this PR to address your comments. I'm not sure you'll like it. I think I got the idea from somewhere else in Simbody, though I can't find it again.

sherm1:
Oh, sorry Chris. I see I was ambiguous. I meant to literally use the *strings* "Acceleration" etc. in the XML files the way you did for `Stage`. Currently there will be 0,1,2 actually present in the XML which won't mean anything to a human reader. (I didn't really have a problem with using the numerical values in the `switch` although your remodeled version looks good too.)

chrisdembia:
Ah good idea. I now use strings for (de)serialization of Motion::Level.

adamkewley:

PR #504, opened 2016-06-06
Add a helper for fromXmlElement implementations; Add serialization for Rotation

chrisdembia:
This PR contains a new helper function that can be used when implementing `fromXmlElement()`. Right now, it's named `fromXmlElementHelperHelper()`; a better name would be `fromXmlElementHelper()` but that's taken for now. This is a variadic template function that takes references to the variables you want to set. It also handles `AbstractValue`s in a uniform way (if the `AbstractValue` is contained in a smart pointer). I don't expect _all_ implementations of `fromXmlElement()` to use `fromXmlElementHelperHelper()`, but I think it can be useful in many cases and reduces the chance of making an error. In particular, this helper does not really help with managing backwards compatibility for multiple versions; I expect the implementor would have to detect the version and then make appropriate calls to the helperhelper.

This PR also adds xml (de)serialization methods for Rotation. 

@sherm1, even if you like these changes, I expect some cleanup will be necessary. I would like to do such cleanup in follow-up PRs to the feature branch. If you like the helperhelper, I will also create one for `toXmlElement()`.

Some questions:
1. What's the best way to handle the template type to `Rotation_`? Right now, I ignore it. Is that okay? Should we use another Xml attribute (`P="Real"`)?
2. Right now, `Rotation_::fromXmlElement()` does _not_ try to make the provided Mat33 a valid rotation matrix. I think this is the right thing to do, but your opinion would be helpful. I think it's important to get the _exact_ same results with the deserialized State, and this requires using all values exactly as they were, without doing any corrections. I tried using `Rotation(mat)` originally to deserialize, and this led to the last 2 or 3 digits being different from the original values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/504)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -158,6 +158,19 @@ of the underlying concrete type.
 inline std::ostream& operator<<(std::ostream& o, const AbstractValue& v) 
 {   o << v.toXmlElement(""); return o; }
 
+// TODO document this method.
+template <template <typename> class OwningSmartPtrType, typename ... TArgs>
+inline void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
+        Xml::element_iterator end,
+        OwningSmartPtrType<AbstractValue>& fieldValue, const std::string& fieldName, 
+        TArgs& ... fields) {
+    assert(eit != end);
+    std::unique_ptr<AbstractValue> vp =
+        AbstractValue::createFromXmlElement(*eit++, fieldName);
+    fieldValue.reset(vp.release());
+    fromXmlElementHelperHelperRecurse(eit, end, fields...);
+}
+
 
 //==============================================================================
 //                               VALUE <T>
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -165,16 +165,17 @@ inline std::ostream& operator<<(std::ostream& o, const AbstractValue& v)
 // (through fromXmlElementHelperHelper()) a smart pointer to the AbstractValue
 // they want deserialized. We heap-allocate a new AbstractValue of the right
 // type, and transfer ownership to the passed-in smart pointer.
-template <template <typename...> class OwningSmartPtrType, typename ... FieldTypes>
+template <template <typename... PtrArgs> class OwningSmartPtrType,
+        typename... PtrArgs, typename... FieldTypes>
 inline void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
         Xml::element_iterator end,
-        std::pair<OwningSmartPtrType<AbstractValue>*, const char*> firstField,
-        std::pair<FieldTypes*, const char*> ... remainingFields) {
+        std::pair<OwningSmartPtrType<AbstractValue, PtrArgs...>*, const char*> firstField,
+        std::pair<FieldTypes*, const char*>... remainingFields) {
     assert(eit != end);
     OwningSmartPtrType<AbstractValue>* value = firstField.first;
     const std::string& name = firstField.second;
     std::unique_ptr<AbstractValue> vp =
-        AbstractValue::createFromXmlElement(*eit++, name);
+            AbstractValue::createFromXmlElement(*eit++, name);
     value->reset(vp.release());
     fromXmlElementHelperHelperRecurse(eit, end, remainingFields...);
 }
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -159,16 +159,18 @@ inline std::ostream& operator<<(std::ostream& o, const AbstractValue& v)
 {   o << v.toXmlElement(""); return o; }
 
 // TODO document this method.
-template <template <typename> class OwningSmartPtrType, typename ... TArgs>
+template <template <typename> class OwningSmartPtrType, typename ... FieldTypes>
 inline void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
         Xml::element_iterator end,
-        OwningSmartPtrType<AbstractValue>& fieldValue, const std::string& fieldName, 
-        TArgs& ... fields) {
+        std::pair<OwningSmartPtrType<AbstractValue>*, const char*> firstField,
+        std::pair<FieldTypes*, const char*> ... remainingFields) {
     assert(eit != end);
+    OwningSmartPtrType<AbstractValue>* value = firstField.first;
+    const std::string& name = firstField.second;
     std::unique_ptr<AbstractValue> vp =
-        AbstractValue::createFromXmlElement(*eit++, fieldName);
-    fieldValue.reset(vp.release());
-    fromXmlElementHelperHelperRecurse(eit, end, fields...);
+        AbstractValue::createFromXmlElement(*eit++, name);
+    value->reset(vp.release());
+    fromXmlElementHelperHelperRecurse(eit, end, remainingFields...);
 }
 
 
--- a/SimTKcommon/include/SimTKcommon/internal/Value.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Value.h
@@ -158,8 +158,14 @@ of the underlying concrete type.
 inline std::ostream& operator<<(std::ostream& o, const AbstractValue& v) 
 {   o << v.toXmlElement(""); return o; }
 
-// TODO document this method.
-template <template <typename> class OwningSmartPtrType, typename ... FieldTypes>
+// This function is invoked by fromXmlElementHelperHelper().
+// Users should not call this method.
+// Since one cannot use fromXmlElementHelper() to deserialize AbstractValues,
+// we must handle them specially. Here, the user passes us
+// (through fromXmlElementHelperHelper()) a smart pointer to the AbstractValue
+// they want deserialized. We heap-allocate a new AbstractValue of the right
+// type, and transfer ownership to the passed-in smart pointer.
+template <template <typename...> class OwningSmartPtrType, typename ... FieldTypes>
 inline void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
         Xml::element_iterator end,
         std::pair<OwningSmartPtrType<AbstractValue>*, const char*> firstField,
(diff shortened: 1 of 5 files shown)


SimTKcommon/include/SimTKcommon/internal/Xml.h
```diff
@@ -1708,10 +1709,76 @@ template <class T> inline
...
+// is more verbose but also more clear:
+//
+// fromXmlElementHelperHelper("Foo", 1, e, "reqName",
+//     std::make_pair{bar, "bar"},
+//     std::make_pair{baz, "baz"}
+//     );
+template <typename T, typename ... TArgs> inline
+void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
+        std::pair<T&, std::string> firstField,
+        std::pair<TArgs&, std::string> ... fields) {
```
sherm1: I like the interface with the pairs!

SimTKcommon/tests/TestStateXml.cpp
```diff
@@ -25,6 +25,46 @@
 
 using namespace SimTK;
 
+/* This function converts the provided instance of `T` to Xml, deserializes
+it from Xml, ensures the deserialized instance is equal to the original
+instance, reserializes the deserialized instance back to Xml and ensures
+the original desrialization and the reserialization are equal.
+
+One might set `checkInMemoryEquality` to false if `value` contains NaNs. */
```
sherm1: This seems awkward, and we do want to check if NaNs go in and out correctly. Could you add explicit isNaN() checks? It would be good to make sure inf and -inf are working too (operator== is fine for them but any serialization that doesn't go properly through Simbody's formatting for floating point numbers won't work right).

SimTKcommon/tests/TestStateXml.cpp
```diff
@@ -25,6 +25,46 @@
...
+template <typename T>
+void testXmlRoundtrip(const T& value, bool checkInMemoryEquality = true,
```
sherm1: I like this function -- that's a very good (stringent) test for (de)serialization.

sherm1:
Cool. I like it. Macho template metaprogramming! Would like to see it use the pair interface though.

SimTKcommon/tests/TestStateXml.cpp
chrisdembia: Ah good point about inf.


> This seems awkward

Because we have to manually say not to check for equality?

> Could you add explicit isNaN() checks?

Could you clarify? For non-scalar types, I want to make sure that specific elements are NaN/not NaN, rather than just checking that there *is* a NaN somewhere in the matrix/vector. Are you suggesting replacing the manual argument with `if (x.isNaN()) { skip operator==() }`?

SimTKcommon/tests/TestStateXml.cpp
chrisdembia: Let me know if you think it should it go somewhere else. It could be useful for multiple test executables.

SimTKcommon/tests/TestStateXml.cpp
```diff
@@ -25,6 +25,46 @@
 
 using namespace SimTK;
 
+/* This function converts the provided instance of `T` to Xml, deserializes
+it from Xml, ensures the deserialized instance is equal to the original
+instance, reserializes the deserialized instance back to Xml and ensures
+the original desrialization and the reserialization are equal.
```
sherm1: typo

SimTKcommon/tests/TestStateXml.cpp
sherm1: Oh, I guess the NaN's still get checked via the string comparison so that's OK. I think what's really needed is a different equality operator (other than `operator==`) that returns true for `NaN==NaN`. Maybe something like `template <class T> bool isSameAs(const T& thing1, const T& thing2)` where the default implementation would just invoke `operator==`. GTest has something like that. Anyway, I'd be happy to say that's out of scope for this PR -- how about filing an issue for it instead?

SimTKcommon/tests/TestStateXml.cpp
sherm1: > Let me know if you think it should it go somewhere else.

I think it's fine where it is for now. We can always generalize it later if warranted.

SimTKcommon/include/SimTKcommon/internal/Xml.h
chrisdembia: Hmmm.. I tried to get it to work again and I ran into some issues. It's still possible but it doesn't look as nice as I had hoped:


```
fromXmlElementHelperHelper("Foo", 1, e, "reqName",
    std::make_pair(std::ref(bar), "bar"),
    std::make_pair(std::ref(baz), "baz"));
```

The issue is that `make_pair` takes its arguments by value.

chrisdembia: Two alternatives:

1. We create our own `make_refpair`: `template <typename F, typename S> make_refpair(F&, S&)`.
2. We use pointers instead of references.

sherm1: Pointers would be fine here I think. That would mean we'd write `std::make_pair<&thing, "name">` right? That looks good to me now. I've gotten a little more relaxed about pointers for output variables since it is the Google style that we're using in Drake. I can see the appeal somewhat  since the `&thing` makes it clear that the variable is going to be modified. I *think* that is still nicer than separate arguments -- what do you think?

SimTKcommon/tests/TestStateXml.cpp
chrisdembia: I have created issue #505 for that. 

I realized that `if (x.isNaN())` is not a good way to remove the need for that extra bool argument because `x` may not have an `isNaN()`. And yes, for now, `NaN`s are checked via the string comparison (it's actually why I added the string comparison).

Let me know if you think there is still something to change about this method for this PR.

SimTKcommon/include/SimTKcommon/internal/Xml.h
chrisdembia: I think the separate arguments presents a greater chance for errors. I like the idea of using pointers. The ideal situation would be:

```fromXmlElementHelperHelper("Foo", 1, e, "reqName",
    {bar, "bar"},
    {baz, "baz"});
```

which I can explore in a subsequent PR.

I've updated this PR to use `std::make_pair(&thing, "name")`.

SimTKcommon/tests/TestStateXml.cpp
chrisdembia: Fixed.

chrisdembia:
I think I've addressed your comments. I'd like to hear what you think about the two questions in the initial post, though.

sherm1:
Sorry, forgot about those:
1. We only use `Rotation_<double>` so I think it is fine just to support that. `float` turned out to be less than useless.
2. I agree that bit-identical is the goal here.

SimTKcommon/include/SimTKcommon/internal/Value.h
```diff
@@ -158,6 +158,21 @@ of the underlying concrete type.
 inline std::ostream& operator<<(std::ostream& o, const AbstractValue& v) 
 {   o << v.toXmlElement(""); return o; }
 
+// TODO document this method.
```
sherm1: better to document this now -- it's unlikely to get revisited later and in any case it would be harder to remember what to say about it then

SimTKcommon/include/SimTKcommon/internal/Xml.h
```diff
@@ -1708,10 +1709,77 @@ template <class T> inline
...
 }
 
 
+inline void fromXmlElementHelperHelperRecurse(Xml::element_iterator eit,
+        Xml::element_iterator end) {
+    assert(eit == end);
+}
+
+/*
+// TODO decide between this and the other interface.
```
sherm1: should this comment block be removed now?

SimTKcommon/include/SimTKcommon/internal/Xml.h
```diff
@@ -1698,6 +1698,7 @@ template <class T> inline
 auto fromXmlElementHelper(T& thing, Xml::Element& elt, 
                           const std::string& requiredTag, bool)
     -> decltype(std::declval<T>().fromXmlElement(elt,requiredTag)) {
+    // TODO std::cout << "DEBUGfromXmlmember" << elt << std::endl;
```
sherm1: remove dead debugging statements?

chrisdembia:
> 1. We only use Rotation_<double> so I think it is fine just to support that. float turned out to be less than useless.

Do you think the code is fine as-is then, or should I provide the definition of the serialization functions for only`Rotation_<double>`?

sherm1:
I think the code is fine now. Assuming I'm understanding it, it will work for any instantiation of `Rotation` but won't document what instantiation was used -- since it will always be `double` that's not likely to cause trouble! Even if it were written out as `float` it would be fine if read in the same way, and could even be read back with `double` although of course it would still be low precision.

sherm1:
gcc & clang builds are failing now

chrisdembia:
Thanks; I have a fix locally. I'll push it when I address the other comments.

SimTKcommon/include/SimTKcommon/internal/Xml.h
chrisdembia: Done.

SimTKcommon/include/SimTKcommon/internal/Xml.h
chrisdembia: Yes; done.

SimTKcommon/include/SimTKcommon/internal/Value.h
chrisdembia: Done.

chrisdembia:
I believe I have addressed all of your comments. I wrote documentation for `fromXmlElementHelperHelper()` (and also viewed the html output). I plan to improve documentation in future PRs. I'm thinking about creating a new doxygen page for all the `(from|to)Xml` functions.

sherm1:
TestXML failed in AppVeyor.

chrisdembia:
Oh..well that's pretty frustrating. MSVC isn't realizing to use the more specific template function.

chrisdembia:
TestXml now passes with MSVC.

chrisdembia:
I moved `testXmlRoundtrip()` to `Testing.h`, as a static member function in `SimTK::Test`. I will need this function in `TestSimbodyStateXml.cpp` as well as in `TestStateXml.cpp`.

adamkewley:

PR #506, opened 2016-06-06
Fix bug in ContinuousVarInfo::toXmlElement.

chrisdembia:
I think were was a copy-paste bug in ContinuousVarInfo. This PR fixes it.


sherm1:
Good catch!

PR #517, opened 2016-07-28
SimbodyMatterSubsystem::calcConstraintAccelerationErrors().

### LINKED ISSUE #514: Public interface for calcConstraintAccelerationErrors()
Right now, there is an internal method `SimbodyMatterSubsystemRep::calcConstraintAccelerationErrors()`; this method is not available to Simbody users. I'd like to add a method `SimbodyMatterSubsystem::calcConstraintAccelerationErrors()`. This would be very similar to the presence of `SimbodyMatterSubsystemRep::calcBodyAccelerationFromUDot()` and `SimbodyMatterSubsystem::calcBodyAccelerationFromUDot()`. For example, the user method would accept non-contiguous input/output Vectors.

I plan to add a test case in `TestConstraints`, based on some of the tests in `TestMassMatrix` and on `testConstraintMatrices()` in `TestConstraints`.

I plan to use this method in OpenSim alongside computing an implicit form of the dynamics.

chrisdembia:
Previously, `SimbodyMatterSubsystemRep` had an operator
`calcConstraintAccelerationErrors()`, but there was no related operator for use
by users. This PR introduces such an operator. The operator is tested
in TestConstraints to ensure it gives results that are consistent with
other methods/operators.

In the doxygen, I stated that the complexity of the operator is O(m+n) This was determined by inspecting the internal documentation for:
 * `SimbodyMatterSubsystemRep::calcConstraintAccelerationErrors()` (O(m+n)), 
 * `SimbodyMatterSubsystemRep::calcBodyAccelerationsFromUDot()` (O(nu+nb)), and 
 * `SimbodyMatterSubsystem::calcQDotDot()` (O(n)).

Fixes #514.

### DIFF
--- a/Simbody/tests/TestConstraints.cpp
+++ b/Simbody/tests/TestConstraints.cpp
@@ -582,6 +582,152 @@ void testConstraintMatrices() {
     delete &system;
 }
 
+// Test the operator SimbodyMatterSubsystem::calcConstraintAccelerationErrors(),
+// which computes pvaerr = G udot - b. For the most part, we just ensure that
+// this operator gives results consistent with other methods.
+void testConstraintAccelerationErrors() {
+    // Create a chain of bodies. Copied from testConstraintMatrices().
+    MultibodySystem& system = createSystem();
+    SimbodyMatterSubsystem& matter = system.updMatterSubsystem();
+    MobilizedBody& first = matter.updMobilizedBody(MobilizedBodyIndex(1));
+    MobilizedBody& second = matter.updMobilizedBody(MobilizedBodyIndex(2));
+    MobilizedBody& fifth = matter.updMobilizedBody(MobilizedBodyIndex(5));
+    MobilizedBody& last = matter.updMobilizedBody(MobilizedBodyIndex(NUM_BODIES));
+
+    // Add a body on a free joint to body 5 then weld it.
+    MobilizedBody::Free extra(fifth, Transform(),
+        MassProperties(1, Vec3(.01,.02,.03), Inertia(1,1.1,1.2)), Transform());
+    Constraint::Weld weld(extra, fifth);
+
+    Constraint::Ball ball(first, last);
+    Constraint::ConstantAcceleration accel2(second, MobilizerUIndex(1), .01);
+    Constraint::ConstantSpeed speed5(fifth, MobilizerUIndex(1), .1);
+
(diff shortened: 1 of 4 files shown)


Simbody/tests/TestConstraints.cpp
```diff
@@ -582,6 +582,152 @@ void testConstraintMatrices() {
...
+        // With udot from forward dynamics, errors should be close to zero.
+        Vector pvaerrFromOperator;
+        matter.calcConstraintAccelerationErrors
+                (stateOperator, udotFromForward, pvaerrFromOperator);
+
+        SimTK_TEST(pvaerrFromOperator.norm() < 1e-10);
+
+        // The same code should have been executed when using forward
+        // dynamics or the operator, so the errors should be exactly the same.
+        SimTK_TEST(pvaerrFromOperator.size() == m);
```
sherm1: This condition is probably too strict. It would be OK if later someone didn't run *exactly* the same code. I would suggest checking to SignificantReal (1e-14 ish) which is what you get with `SimTK_TEST_EQ`.

Simbody/tests/TestConstraints.cpp
```diff
@@ -582,6 +582,152 @@ void testConstraintMatrices() {
...
+
+    // Test noncontiguous input and output vectors (similar to TestMassMatrix).
+    {
+        system.realize(state, SimTK::Stage::Velocity);
+        Matrix MatUdot(3, nu);  // use middle row
+        MatUdot.setToNaN();
+        MatUdot[1] = ~udotFromForward;
+        Matrix Matpvaerr(3, m); // use middle row
+        Matpvaerr.setToNaN();
+        auto TODO = ~MatUdot[1];
```
sherm1: Did you mean to leave this here? Looks like it could be removed.

sherm1:
Wow -- this looks beautiful, Chris. I really like the comprehensive tests! I had a few minor comments but otherwise this is ready to go in.

Simbody/tests/TestConstraints.cpp
chrisdembia: Oops. You weren't supposed to see that!

chrisdembia:
Thank you for the quick review. I have addressed both of your comments.

chrisdembia:
Thank you.

PR #530, opened 2016-10-02
simbody.pc.in: include dir to match cmake config

scpeters:
The cmake config file defines `Simbody_INCLUDE_DIR` as [PACKAGE_SIMBODY_INCLUDE_INSTALL_DIR](https://github.com/simbody/simbody/blob/master/cmake/SimbodyConfig.cmake.in#L38-L39), which is set to [${CMAKE_INSTALL_INCLUDEDIR}/simbody](https://github.com/simbody/simbody/blob/master/CMakeLists.txt#L578), but the pkg-config file sets it to just the include folder.

### DIFF
--- a/cmake/pkgconfig/simbody.pc.in
+++ b/cmake/pkgconfig/simbody.pc.in
@@ -1,6 +1,6 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 libdir=${prefix}/@CMAKE_INSTALL_LIBDIR@
-includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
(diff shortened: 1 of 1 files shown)


chrisdembia:
I think this means the pkgconfig file cannot be used on Windows; see [here](https://github.com/simbody/simbody/blob/master/CMakeLists.txt#L571). Can you use `@SIMBODY_INCLUDE_INSTALL_DIR@` instead of `@CMAKE_INSTALL_INCLUDEDIR@`?

scpeters:
Sure thing. I'll go ahead and squash those commits...

chrisdembia:
Thank you. LGTM.

Issue #520, opened 2016-08-04
CMAES Optimizer get stuck for high dimension problem


mitkof6:
I am trying to perform Anderson F.C., el al., A Dynamic Optimization Solution for Vertical Jumping, and to find a solution through optimization. There is something that me and some students want to test. I took a code written from Carmichael and adapted it so I can have the same criterion as in the original paper. I choose CMAES, because the other optimization algorithms does not find a good solution (local minimum). The source code can be downloaded from [github](https://github.com/mitkof6/skyjump) and it uses OpenSim3.3.

In my implementation I have 24 muscles and I also choose the number of control nodes per muscle (so I can find the muscle control signal). When I have 9 control points I have 24 * 9 = 216 parameters and CMAES, initializes and runs normally. I wanted to add more control points and when I set p > 10 CMAES gets stuck in the initialization, and the objective function is not called at all. Is this a limitation to the algorithm?

Also, because I am lazy sometimes, I find that brute force can give you a good solution and does not get stuck in local minimum. Are you interested in a pure random optimization as an alternative algorithm to compliment Simbody optimization algorithms?

chrisdembia:
Cool project; thank you for sharing.

Dimitar, could you make the title of this issue more specific? Is this a Simbody issue, or a CMAES issue? That is, if you use other CMAES implementations, do they have the same problem? What is p?

mitkof6:
I can't figure out where it gets stuck, but I will have a look again. I have used Simobody's CMAES. 

p is the number of control nodes per muscle, so when I increase p the algorithm gets stuck in the initialization.

mitkof6:
It seems that a bad initialization of the parameter space, and the default standard deviation of CMAES can cause this problem. A proper initialization of the CMAES will not cause problems, even if there are many parameters.

chrisdembia:
Alternatively, it may help to scale the problem better? Perhaps the best fix is to add more documentation about choosing the parameters of the CMAES algorithm.

carmichaelong:
Was it the case that your initial guess had many parameters that were near their bounds and/or your standard deviation was large compared to the range of the bounds of your parameters? Since CMAES has to sample a distribution to generate a population, sometimes it can sample outside of the bounds. In this case, it must *resample* the distribution (or else you'd get some bias in your sampling if you just clamp these guesses). With many parameters, this means it becomes more and more unlikely to sample a valid distribution, which means you can get caught in initialization (or really at any step along the way, but mainly at initialization).

[...]

PR #540, opened 2016-12-09
CMAES treat init_stepsize as Vector 

mitkof6:
#520 

Change init_stepsize so that it can be assigned as a Vector valued property.

### DIFF
--- a/SimTKmath/include/simmath/Optimizer.h
+++ b/SimTKmath/include/simmath/Optimizer.h
@@ -316,8 +316,16 @@ private:
  *
  * - <b>popsize</b> (int; default: depends on number of parameters) The
  *   population size (also known as lambda).
- * - <b>init_stepsize</b> (real; default: 0.3) Initial step size; same for all
- *   parameters (also known as sigma). A warning is emitted if this is not set.
+ * - <b>init_stepsize</b> (Vector; default: 0.3) Initial stddev; After the encoding of variables,
+ *   the initial solution point x0 and the initial standard deviation (step_size) σ0 must be chosen.
+ *   In a practical application, one often wants to start by trying to improve a given solution
+ *   locally. In this case we choose a rather small σ0 (say in [0.001, 0.1], given the x-values
+ *   "live" in [0,10]). Thereby we can also check, whether the initial solution is possibly a local
+ *   optimum. When a global optimum is sought-after on rugged or multimodal landscapes, σ0 should be
+ *   chosen such that the final desirable location (or at least some of its domain of attraction) is
+ *   not far outside of x0 ± 2σ0 in each coordinate. (Remark that in Rn, if each boundary domain is
+ *   in distance σ, then the boundary corner is σ √n away, which poses a slight dilemma for larger n.)
+ *   A warning is emitted if this is not set.
  * - <b>seed</b> (int; default: 0, which uses clock time) Seed for the random
  *   number generator that is used to sample the population from a normal
  *   distribution. See note below.
@@ -393,6 +401,8 @@ public:
     bool setAdvancedIntOption( const char *option, const int value );
     /// Set the value of an advanced option specified by an boolean value.
     bool setAdvancedBoolOption( const char *option, const bool value );
+    /// Set the value of an advanced option specified by an Vector value.
+    bool setAdvancedVectorOption( const char *option, const Vector value );
 
     
     /// Set which numerical differentiation algorithm is to be used for the next
--- a/SimTKmath/include/simmath/Optimizer.h
+++ b/SimTKmath/include/simmath/Optimizer.h
@@ -316,16 +316,21 @@ private:
  *
  * - <b>popsize</b> (int; default: depends on number of parameters) The
  *   population size (also known as lambda).
- * - <b>init_stepsize</b> (Vector; default: 0.3) Initial stddev; After the encoding of variables,
- *   the initial solution point x0 and the initial standard deviation (step_size) σ0 must be chosen.
+ * - <b>init_stepsize</b> (Vector/Real; default: 0.3) Initial stddev; After the encoding of variables,
+ *   the initial solution point x0 and the initial standard deviation (step_size) sigma0 must be chosen.
  *   In a practical application, one often wants to start by trying to improve a given solution
- *   locally. In this case we choose a rather small σ0 (say in [0.001, 0.1], given the x-values
+ *   locally. In this case we choose a rather small sigma0 (say in [0.001, 0.1], given the x-values
  *   "live" in [0,10]). Thereby we can also check, whether the initial solution is possibly a local
- *   optimum. When a global optimum is sought-after on rugged or multimodal landscapes, σ0 should be
+ *   optimum. When a global optimum is sought-after on rugged or multimodal landscapes, sigma0 should be
  *   chosen such that the final desirable location (or at least some of its domain of attraction) is
- *   not far outside of x0 ± 2σ0 in each coordinate. (Remark that in Rn, if each boundary domain is
- *   in distance σ, then the boundary corner is σ √n away, which poses a slight dilemma for larger n.)
+ *   not far outside of x0 ± 2sigma0 in each coordinate. (Remark that in Rn, if each boundary domain is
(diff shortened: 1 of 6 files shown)


SimTKmath/include/simmath/Optimizer.h
```diff
@@ -316,8 +316,16 @@ class SimTK_SIMMATH_EXPORT OptimizerSystem {
  *
  * - <b>popsize</b> (int; default: depends on number of parameters) The
  *   population size (also known as lambda).
- * - <b>init_stepsize</b> (real; default: 0.3) Initial step size; same for all
- *   parameters (also known as sigma). A warning is emitted if this is not set.
+ * - <b>init_stepsize</b> (Vector; default: 0.3) Initial stddev; After the encoding of variables,
+ *   the initial solution point x0 and the initial standard deviation (step_size) σ0 must be chosen.
```
chrisdembia: How does the sigma look in doxygen?

mitkof6:
Thank you @chrisdembia I agree with your suggestions.

chrisdembia:
That looks great; except I think an exception should be thrown if the user sets both the Real and the Vector options. What do you think?

mitkof6:
Its best to inform since the user may cause an unexpected behavior.

chrisdembia:
Pinging @carmichaelong in case he's interested in reviewing this.

carmichaelong:
I'm interested in reviewing, but probably won't get to it until tomorrow or Monday. If it's slowing down anyone's work, feel free to merge without me reviewing.

SimTKmath/tests/CMAESTest.cpp
```diff
@@ -514,7 +515,9 @@ void testMultithreading() {
     Optimizer opt(sys, SimTK::CMAES);
     opt.setConvergenceTolerance(1e-12);
     opt.setMaxIterations(5000);
-    opt.setAdvancedRealOption("init_stepsize", 0.3);
+    Vector initStepSize(N, 0.3);// test vector option
+    opt.setAdvancedVectorOption("init_stepsize", initStepSize);
+    //opt.setAdvancedRealOption("init_stepsize", 0.3);// to test exception if both are set
```
sherm1: Should remove commented-out line, or better would be to leave in a test somewhere to verify that the "Real" option is working correctly.

carmichaelong:
Generally looks good. I'd consider a few changes:
1) This change stemmed from the problem of the initialization step taking an (approaching) infinite time due to resampling. Consider adding this point to the comment, with a focus from a user who is running into this problem (i.e. they may not know the sampling method/step in detail).
2) I'm strongly in favor of the second option from @sherm1's review (i.e. both the "Real" and "Vector" options have a test).
3) Maybe a comment / short discussion in the comment about normalizing variables? This could be discussed in either the "Real" or "Vector" case. This is probably the least important of the three changes.

mitkof6:
Thank you for the comments. I add some comments in the description from:

[https://www.lri.fr/~hansen/cmaes_inmatlab.html#practical]()

mitkof6:
@chrisdembia @carmichaelong @sherm1 I forget to tag you. I have addressed your comments. Please let me know if you need something else.

SimTKmath/include/simmath/Optimizer.h
```diff
@@ -309,27 +309,76 @@ class SimTK_SIMMATH_EXPORT OptimizerSystem {
...
+ *   "live" in [0,10]). Thereby we can also check, whether the initial solution is
+ *   possibly a local optimum. When a global optimum is sought-after on rugged or
+ *   multimodal landscapes, sigma0 should be chosen such that the final desirable
+ *   location (or at least some of its domain of attraction) is not far outside of
+ *   x0 ± 2sigma0 in each coordinate. (Remark that in Rn, if each boundary domain is
+ *   in distance sigma, then the boundary corner is sigma*sqrt(n) away, which poses
+ *   a slight dilemma for larger n.)
+ *
+ *   A warning is emitted if this is not set and the default value is assigned to
+ *   each term of sigma.
```
chrisdembia: Consider a slightly different wording: "default value is used for each variable."

SimTKmath/include/simmath/Optimizer.h
```diff
@@ -309,27 +309,76 @@ class SimTK_SIMMATH_EXPORT OptimizerSystem {
...
- *   not far outside of x0 ± 2sigma0 in each coordinate. (Remark that in Rn, if each boundary domain is
- *   in distance sigma, then the boundary corner is sigma*sqrt(n) away, which poses a slight dilemma
- *   for larger n.)
- *   A warning is emitted if this is not set.
+ * - <b>init_stepsize</b> (Vector/Real; default: 0.3) Initial stddev; After the
+ *   encoding of variables, the initial solution point x0 and the initial standard
+ *   deviation (step_size) sigma0 must be chosen. In a practical application, one
+ *   often wants to start by trying to improve a given solution locally. In this
+ *   case we choose a rather small sigma0 (say in [0.001, 0.1], given the x-values
+ *   "live" in [0,10]). Thereby we can also check, whether the initial solution is
```
chrisdembia: Can remove the comma between "check" and "whether"

SimTKmath/include/simmath/Optimizer.h
```diff
@@ -309,27 +309,76 @@ class SimTK_SIMMATH_EXPORT OptimizerSystem {
...
  *
+ *
+ * Encoding of Variables
+ *
+ * Inappropriate initialization of the algorithm may lead to resampling of the
+ * parameter distribution. Since some parameters may be defined in a bounded
+ * region [a, b]. That is if a user defines the standard deviation and the
+ * initial value of the parameter in a manner, where the resultant distribution
+ * chooses values out of the bounds. To overcome this the problem the user can
+ * formulate the problem as follows:
```
chrisdembia: This section will be useful to users, but this paragraph is unclear to me. Consider the following rewording:

> The CMA algorithm involves sampling a random distribution for the variables (parameters) in the optimization problem. If the sampled values do not lie within the variables' bounds, CMA must resample the distribution until the variables lie within the bounds. This resampling is undesirable, and may prevent the algorithm from functioning properly. There are two ways to avoid excessive resampling:

SimTKmath/include/simmath/Optimizer.h
mitkof6: @chrisdembia ok.

chrisdembia:
Thanks @mitkof6; looks good to me.

SimTKmath/include/simmath/Optimizer.h
```diff
@@ -309,15 +309,80 @@ class SimTK_SIMMATH_EXPORT OptimizerSystem {
...
+ * to avoid excessive resampling:To overcome this the problem the user can
+ * formulate the problem as follows:
+ *
+ * 1) Either define the bounds of the parameter and choose the appropriate stddev
+ * and initial value for each parameter (see init_stepsize)
+ * 2) Or to reformulate the problem, by rescaling the parameter space so that
+ * each parameter map in a region between e.g. [0, 1] and each parameter has an
+ * initial value of 0.5 and stddev of 0.2 (see comments below).
+ *
+ * The specific formulation of a (real) optimization problem has a tremendous
```
carmichaelong: I would add a note somewhere in this section where you got this from (both for a reference for people. and since this is a bit of copy/paste from the website you linked earlier).

mitkof6: @carmichaelong ok.

carmichaelong:
Looks good to me too.

SimTKmath/tests/CMAESTest.cpp
sherm1: Should remove commented-out line, or better would be to leave in a test somewhere to verify that the "Real" option is working correctly.

mitkof6: Hi Sherm, I think I addressed this issue. Please let me know if I am mistaken.

sherm1:
Thanks, Dimitar -- I may have been looking at at earlier diff. Looks great!

mitkof6:
Thank you all for the comments.

PR #546, opened 2017-01-06
polynomial root finder update

stavrosk7:
This is a method which, once you get started, quickly gives a very good approximation to a root of polynomial (and other) equations.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/546)


sherm1:
Closing this since no activity for three years.

PR #548, opened 2017-02-20
Avoid exception in Visualizer::Impl::~Impl().

chrisdembia:
This exception, which occurred when destructing a model whose
simbody-visualizer was already shutdown, was causing OpenSim's MATLAB and
Python bindings to crash.

I imagine that if we are destructing the Visualizer, it is fine if there is difficulty sending the shutdown signal to the simbody-visualizer GUI.


chrisdembia:
Thank you @sherm1 .

PR #551, opened 2017-02-25
Add coverage status of simbody with gcov and codecov.io

Gjacquenot:
This PR adds the coverage status for simbody

It tells us that 50% of Lapack is covered with existing tests.

The current result can be seen here:
https://codecov.io/gh/Gjacquenot/simbody/commits

How it works:
It uses gcov to create coverage data, and continuous integration with Travis sends this data to codecov service. codecov parses the result and proposes a HTML view of the coverage.

I had to change the build directory: it is now in the source directory. Otherwise codecov fail to parse coverage result

Once this PR accepted, one should create a codecov account and update the README.md for the coverage badge (if adresses I have written are incorrect).


sherm1:
Thanks, @Gjacquenot. This looks cool but I'm not sure I understand it. What do you mean by "50% of Lapack" -- Lapack is a separate package and we don't try to test it ourselves.

@chrisdembia, any thoughts on this?

chrisdembia:
Thanks for submitting this PR, @Gjacquenot, and for your interest in improving Simbody!  I think it is fine to add this codecov integration. This will help encourage contributors to fully test their changes. Is this service free? If so, what are the conditions under which the service is free?

> I had to change the build directory: it is now in the source directory. Otherwise codecov fail to parse coverage result

This is fine with me, since you use a subdirectory within the build directory (`simbody-build`).

I'm happy to make a codecov account after this gets merged.


Can we avoid the new BUILD_TYPE? It seems like this is an inappropriate use of "BUILD_TYPE", which is intended to alter compiler flags. Could you instead create a CMake option?  `option(SIMBODY_COVERAGE "<add-description>" OFF)`.

It would also be great to avoid adding the additional build job with travis; could an existing build job be used for testing coverage? Four build jobs is already a lot.

In the README, can you add a space before the badge? Right now, there is no space between the appveyor and codecov badges: 
image

Gjacquenot:
@sherm1 : I am sorry I wanted to mention simbody instead of Lapack. It is a copy paste error. The mistake comes from the fact  that I have created the same PR for Lapack: https://github.com/Reference-LAPACK/lapack

@chrisdembia codecov.io mentions that their services are free for public repositories stored on GitHub, as for travis and appveyor: Every account has unlimited public repositories and one free private repository credit.
I have read the terms: https://codecov.io/site/terms, and see nothing weird about the use we want to do.

I see there are other web services that can do so:

- Coveralls: https://coveralls.io
- Code Climate: https://codeclimate.com

They all offer their services for free for open source services

To benefit from coverage data, we need to activate coverage flags, that are included [here](https://github.com/Gjacquenot/simbody/blob/dev/coverage/cmake/Findcodecov.cmake#L15)

BTW, codecov.io can aggregate covering data from different builds (e.g. builds that use different preprocessing macros for example)

I agree that 5 builds is too much: I propose to switch one of the existing build to Coverage. For time being, gcc is used to generate coverage data, I know we can use clang but not sure it works right now.

I think that using a CMake option does not change anything, it is equivalent to use `BUILD_TYPE=Coverage`.

If you think that the metrics is not representative, we can select which directories to cover or not.

Have you taken a look at the coverage [result](https://codecov.io/gh/Gjacquenot/simbody/tree/c82570d2cb1cc235fdc51bd73abff6316c885cb2)?

chrisdembia:
Thanks for that information. Based on my reading of `Findcodecov.cmake`, it does not seem like a separate BUILD_TYPE is necessary, since the flags are specified using [`set_property()`](https://github.com/Gjacquenot/simbody/blob/dev/coverage/cmake/Findcodecov.cmake#L181), not by adding flags to `CMAKE_CXX_FLAGS_COVERAGE`.

And I would use the clang Debug build to run coverage; simply add `-DSIMBODY_COVERAGE=ON` in this build job to the `cmake` command.

Also, if coverage is not supported for other compilers like MSVC, it would be great if we could give an error when using an unsupported compiler and `SIMBODY_COVERAGE` is `ON`.

chrisdembia:
> If you think that the metrics is not representative, we can select which directories to cover or not.

Looks fine to me.

> Have you taken a look at the coverage result?

Yes, briefly. Neat!

If clang does not work yet, the Debug build could use gcc instead. The reason for having used clang was that clang compiles faster.

CMakeLists.txt
```diff
@@ -407,6 +407,12 @@ if(${CMAKE_CXX_COMPILER_ID} MATCHES "GNU" OR
     set(CMAKE_C_FLAGS_MINSIZEREL ${BUILD_C_FLAGS_MINSIZEREL}
         CACHE STRING "Can't change here -- see BUILD_C..." FORCE)
 
+    set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
+    if(${CMAKE_BUILD_TYPE} STREQUAL "Coverage")
+        message(STATUS "Adding coverage")
```
chrisdembia: Could this message be more descriptive? "Adding ability to assess test coverage using codecov."

Gjacquenot:
@chrisdembia: I have converted BUILD_TYPE Coverage to the option SIMBODY_COVERAGE.

I have tried to use clang for the coverage report, however I had an [error](https://travis-ci.org/Gjacquenot/simbody/jobs/205934090#L1898)

CMakeLists.txt
```diff
@@ -51,6 +56,9 @@ if(MSVC)
                             " 'Visual C++ Compiler Nov 2013 CTP (CTP_Nov2013)'"
                             " comment this test and configure normally.")
     endif()
+    if(SIMBODY_COVERAGE)
+        message(FATAL_ERROR "Code coverage is not possible with MSVC")
```
chrisdembia: Please add a period at the end of the sentence.

CMakeLists.txt
```diff
@@ -41,6 +41,11 @@ endif()
 # Include GNUInstallDirs to get canonical paths
 include(GNUInstallDirs)
 
+# Declare the option for code coverage
+option(SIMBODY_COVERAGE
+       "Adding ability to assess test coverage. Code coverage is possible with gcc or clang."
```
chrisdembia: I think this line is over 80 columns! Consider: "Adding ability to assess test coverage (requires gcc or clang)".

.travis.yml
```diff
@@ -69,10 +70,11 @@ script:
...
   ## Run coverage and send result to codecov.io
-  - if [[ "$BTYPE" == "Coverage" ]];
+  - if [[ "$SIMBODY_COVERAGE" == "ON" ]];
     then
-    echo "Coverage";
+    echo "Running coverage target";
     make coverage;
+    echo "Sending coverage result to codecov.io";
     bash <(curl -s https://codecov.io/bash) -X gcov;
     fi
```
chrisdembia: I've been skeptical about if these multi-line constructs are safe in `travis-ci.yaml`. Do you know more about this? For example, there's some intense discussion here: https://github.com/travis-ci/travis-ci/issues/497. Unless you know that these multi-line constructs won't generate confusing build statuses/error messages, could we put this in one line? Or repeat the `if`, as done elsewhere in the yaml file.

Gjacquenot: Fixed in edeae47

CMakeLists.txt
Gjacquenot: Fixed in 94f256d

CMakeLists.txt
Gjacquenot: Fixed in faf6321

Gjacquenot:
@chrisdembia : I have taken into account your remarks. Let me know if you want me to rebase some of my commits.

chrisdembia:
Thanks for all those changes. I think this can be merged now.

Gjacquenot:
I agree with you @sherm1. It is not easy to maintain a code that we/you haven't written.

Files for covering you mentioned are taken from https://github.com/RWTH-HPC/CMake-codecov
They are not taken from CMake repository.

We could add this repository as a submodule, if we don't want to take the responsibility to maintain this part of the code. I have integrated and updated two files of this repository so as to have a standalone repository.

Do you want to create a submodule ? or keep that way? or another solution you may think of?

sherm1:
I don't think we should introduce a submodule for this since currently Simbody builds with no submodules at all. Any thoughts, @chrisdembia ?

chrisdembia:
I think it's fine to add the *.cmake files, since they are not part of CMake. They are mostly for use on travis. 

Alternately, we could download them on travis from https://github.com/RWTH-HPC/CMake-codecov  instead of adding files to the repository. This would make it harder for people to run the coverage test on their own computers, but that would be fine with me.

chrisdembia:
I agree that we should not introduce a submodule.

sherm1:
OK, I'll merge as-is then. Thanks @Gjacquenot!

PR #552, opened 2017-03-13
Remove need for users to `include_directories(${Simbody_INCLUDE_DIR})`

chrisdembia:
This PR makes CMake export the directories that should be included by client projects. Therefore, client projects need only `target_link_libraries(<target> SimTKsimbody)` and their library/executable will have Simbody's include directory.

This will be helpful for OpenSim users, who might not expect the need to handle both OpenSim's and Simbody's include directories.

### DIFF
--- a/cmake/SampleCMakeLists.txt
+++ b/cmake/SampleCMakeLists.txt
@@ -15,7 +15,7 @@ set(my_header_files myexe.h)
 # able to guess, you'll need to tell find_package where to look.
 find_package(Simbody REQUIRED)
 
-# This is only necesasry if Simbody was built using CMake older than 3.0.2.
+# This is only necessary if Simbody was built using CMake older than 3.0.2.
 include_directories(${Simbody_INCLUDE_DIR})
 
 add_executable(myexe ${my_source_files} ${my_header_files})
--- a/cmake/SampleCMakeLists.txt
+++ b/cmake/SampleCMakeLists.txt
@@ -15,6 +15,7 @@ set(my_header_files myexe.h)
 # able to guess, you'll need to tell find_package where to look.
 find_package(Simbody REQUIRED)
 
+# This is only necesasry if Simbody was built using CMake older than 3.0.2.
 include_directories(${Simbody_INCLUDE_DIR})
 
 add_executable(myexe ${my_source_files} ${my_header_files})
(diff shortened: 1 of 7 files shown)


chrisdembia:
I tested this locally with a sample client project.

cmake/SampleCMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(my_header_files myexe.h)
 # able to guess, you'll need to tell find_package where to look.
 find_package(Simbody REQUIRED)
 
+# This is only necesasry if Simbody was built using CMake older than 3.0.2.
```
sherm1: Typo.

chrisdembia:
> Do we have a CMake expert out there who could review this for you?

@elen4, are you interested?

cmake/SampleCMakeLists.txt
chrisdembia: Oops; I've been making this typo a lot.

elen4:
Hi Chris, the PR looks fine to me. I also tested it locally on a Ubuntu 14.04 machine with a CMake 3.2.2 (backport), using a client project based on your sample CMakeLists, and it worked (except that it complained about the -std=c++11 flag not being set, but that's another story). I also tested an OpenSim build (API only), removing the `include_directories("${Simbody_INCLUDE_DIR}")` from the main CMakeLists, and that also worked.

As a side note, I think that you could also simplify the CMakeLists of Simbody itself by using also the `$<BUILD_INTERFACE:>` case instead of passing around the `SimTKSIMBODY_INCLUDE_DIRECTORIES` variable, but I'm not sure it's worthy it if you are still supporting older CMake versions (I mean, introducing all the `if(CMAKE_VERSION...` would complicate things instead of simplifying them)

chrisdembia:
Wow! Very thorough! Thank you very much @elen4, especially for testing OpenSim!!

> As a side note, I think that you could also simplify the CMakeLists of Simbody itself...

Yes...this would be nice. I thought about this, but chose not to do it for the reason you mentioned and because I didn't want to mess anything up. I would be fine with raising the CMake minimum required version, but was happy to delay this decision.


I was also happy to see that despite all 3 libraries specifying the same include directory, it only appears once in the gcc/clang invocation in the client project. CMake must determine that there are duplicates.

sherm1:
Thanks, @chrisdembia and @elen4! Merging ...

chrisdembia:
Thank you @elen4 and @sherm1 .

PR #554, opened 2017-03-24
Upon shutting down the visualizer GUI, stop listening for GUI events

chrisdembia:
This PR solves an issue that occurs when using OpenSim bindings in MATLAB with the Simbody Visualizer: MATLAB uses 100% CPU even after the OpenSim model (and thus the `SimTK::Visualizer`) is destructed. This was because the visualizer spawns a thread to listen to GUI events and this thread is not cancelled after destructing the visualizer. This PR causes shutting down the visualizer to cancel the event listening thread.

Due to an issue with the POSIX Threads for Windows library, we had to surround `READ` with some additional pthreads code; see [here](https://www.sourceware.org/pthreads-win32/manual/pthread_cancel.html):

> POSIX specifies that a number of system calls (basically, all system calls that may block, such as read(2) , write(2) , wait(2) , etc.) and library functions that may call these system calls (e.g. fprintf(3) ) are cancellation points. Pthreads-win32 is not integrated enough with the C library to implement this, and thus none of the C library functions is a cancellation point.

> A workaround for these calls is to temporarily switch to asynchronous cancellation (assuming full asynchronous cancellation support is installed). So, checking for cancellation during a read system call, for instance, can be achieved as follows:

> 
```c++
pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, &oldCancelType);
read(fd, buffer, length);
pthread_setcanceltype(oldCancelType, NULL);
```

@carmichaelong and I worked on this together. However, neither of us are very familiar with pthreads; if there is a better solution, please let us know. We considered having the GUI send a signal telling the thread to `pthread_exit()`, but that seemed roundabout (the simulator telling the visualizer to tell the simulator's listener to exit).

We have tested this locally on Windows and macOS, and have verified that multiple examples still work as expected.

### DIFF
--- a/Simbody/Visualizer/src/VisualizerProtocol.cpp
+++ b/Simbody/Visualizer/src/VisualizerProtocol.cpp
@@ -311,8 +311,8 @@ VisualizerProtocol::VisualizerProtocol
     // Spawn the thread to listen for events.
 
     pthread_mutex_init(&sceneLock, NULL);
-    pthread_t thread;
-    pthread_create(&thread, NULL, listenForVisualizerEvents, &visualizer);
+    pthread_create(&eventListenerThread, NULL, listenForVisualizerEvents,
+                   &visualizer);
 }
 
 // This is executed on the main thread at GUI startup and thus does not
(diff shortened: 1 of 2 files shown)


chrisdembia:
> Is it possible to test this on Ubuntu also?

Just tested on Ubuntu 14.04; works fine there too. I ran a bunch of examples that use the visualizer and I did not run into any issues.

I also added a comment to the code about why changing the canceltype is necessary.

chrisdembia:
Thank you @sherm1

PR #555, opened 2017-03-27
[travis] RelWithDebInfo -> Release

chrisdembia:
I would like for the Simbody binaries built by travis to eventually be used with the OpenSim binaries that are distributed to users. Initially, I had thought it might make sense to distribute RelWithDebInfo binaries to users, but since @aseth1 recently reported that Release binaries were substantially faster than RelWithDebInfo binaries on Windows, I think it makes sense to release *Release* binaries.

Therefore, I would like to switch Travis' build types from RelWithDebInfo to Release. This also helps make us consistent between Travis and AppVeyor (which is already using Release).


sherm1:
I'm OK with this change but surprised to hear there is a performance difference. I had measured no difference with Simbody. If there is a difference I suspect it is because of how the compiler flags are set in the CMakeLists.txt -- for Simbody I set them explicitly to be the same for Release and RelWithDebInfo. By default I believe RelWithDebInfo uses a lower optimization level.

chrisdembia:
OpenSim seems to use the same optimization flags in Release and RelWithDebInfo:

```
    set(CMAKE_CXX_FLAGS_RELEASE
    "/MP${mxcpu} /D NDEBUG /MD  /O2 /Ob2 /Oi /bigobj /GS- ${CL_INST_SET}"
        CACHE STRING "VC++ Release build compile flags" FORCE)
    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
    "/MP${mxcpu} /D NDEBUG /MD  /O2 /Ob2 /Oi /Zi /bigobj /GS- ${CL_INST_SET}"
        CACHE STRING "VC++ RelWithDebInfo build compile flags" FORCE)
```

Perhaps there was some confusion.

Another reason to consider the switch is that Simbody's Release distribution is 20% of the size of its RelWithDebInfo installation.

chrisdembia:
I'll try to chat with people here about whether we would prefer Release or RelWithDebInfo on Linux/macOS.

sherm1:
Yeah, the theory is that if we ship the big RelWithDebInfo binaries then we'll get better information from end user crashes (because of all the symbol names carried around in the binaries). But maybe that doesn't actually happen in practice? If it isn't helping then shipping tiny little Release binaries does seem appealing! It would still be good to use RelWithDebInfo as the standard builds for developers though, it is definitely helpful for us.

chrisdembia:
I'd like to discuss this with @aymanhab @klshrinidhi and @aseth1 first.

PR #557, opened 2017-03-30
[travis] Upload simbody binaries to sourceforge.

chrisdembia:
To speed up travis builds for OpenSim, we previously uploaded Simbody binaries to
[Bintray](https://bintray.com/) after successful builds on the master branch.
However, we have not been able to upload new binaries to Bintray because
Bintray only lets us update a given release about 100 times.  This means that
OpenSim's Travis builds are relying on a very old Simbody build.

In order to maintain this type of cache, we must move to a different server. In
this PR, we now upload binaries to sourceforge, here:
http://myosin.sourceforge.net/simbody_latest_builds/. (Note: This location is
deliberately somewhat obscure so that users don't really stumble upon it; it's
really for development purposes.)

I wanted to avoid storing a plaintext password for ssh'ing into the sourceforge
server. This results in a fairly convoluted script/procedure. I have tried to
document what I've done so that I have a shot at reproducing the steps in the
future.

I verified that these changes work; you can see that some of the binaries at
https://sourceforge.net/projects/myosin/?source=navbar .


chrisdembia:
I made some additional changes; this is ready for review.

chrisdembia:
Thank you for reviewing.

> Looks good though I'm not able to review it with any depth.

I have attempted to follow all the documentation from Sourceforge and Travis CI, and I have also read a bunch of Stack Overflow posts about using Travis CI to upload files to a server with SSH. My biggest concern was that if something gets messed up (e.g., Sourceforge closes our myosin project, or I accidentally change the user password on Sourceforge), it would be hard for someone else to figure out what I did. To mitigate this, I tried to document the steps I took.

> To confirm: looks like you are deploying RelWithDebInfo builds on Linux & OSX using clang, and a Linux Debug build using gcc. Is that your intention?

That is what I am doing. I'd like to keep the decision about switching RelWithDebInfo to Release as a separate issue. As for Debug, I'd prefer to use clang, but I don't want to undo the code coverage, which requires gcc. In a subsequent PR, I could create a separate build job for clang Debug (or perhaps there are other solutions to retaining both the code coverage check and a clang Debug build). I have verified that the simbody gcc Debug binaries work with opensim-core's clang Debug build; do you think this is alright for now, or is it not a good idea to mix gcc and clang binaries?

sherm1:
I think it is fine -- just wanted to verify that the code was how you wanted it. Thanks!

chrisdembia:
Thank you @sherm1

PR #559, opened 2017-04-04
Visualizer relative search path

### LINKED ISSUE #558: Add relative location from libSimTKsimbody.(so|dylib) to simbody-visualizer to the visualizer search path
PR #64 cleverly inserted the build-time install location of the `simbody-visualizer` to the visualizer search path. However, for OpenSim users who get a binary distribution of OpenSim, this build-time absolute path will be incorrect. This is mostly a problem on macOS (on Windows, `simbody-visualizer` will be on the PATH for OpenSim users; on Linux, hopefully OpenSim will be installed into common system-wide directories), but is also a problem for Linux users installing into non-standard directories. 

The specific use cases I am thinking of are OpenSim MATLAB and Python bindings.

When using Simbody's shared libraries, the code that spawns the visualizer is always in libSimTKsimbody.(so|dylib). We also always know the relative location of the simbody-visualizer from libSimTKsimbody.(so|dylib) (so long as the installation is not reorganized). If we know the location of libSimTKsimbody.(so|dylib), then we can find the simbody-visualizer by combining this known location with the relative path of simbody-visualizer from SimTKsimbody. Fortunately, on most UNIX operating systems, it is possible to obtain the location of the shared library that has been loaded (http://stackoverflow.com/questions/1681060/library-path-when-dynamically-loaded):

```cpp
#include <dlfcn.h>
__attribute__((constructor))
void on_load(void) {
    Dl_info dl_info;
    dladdr((void *)on_load, &dl_info);
    fprintf(stderr, "module %s loaded\n", dl_info.dli_fname);
}
```

I have tested this mechanism on macOS, and I have gotten it to work with OpenSim MATLAB and Python bindings.

@sherm1 , would you be interested in a PR that adds `getParentDir(dl_info.dli_fname) + "/" + rel_path_of_simbody_visualizer_from_lib_dir` (pseudocode) to the visualizer search path on platforms where `dlfcn.h` is available?

On Windows, or when using static libraries (on any platform), I don't think it's possible to use this mechanism. Therefore, for Windows, it would be nice if the search path included the PATH environment variable. This would allow OpenSim to decrease its reliance on the OPENSIM_HOME environment variable, which we are hoping we will no longer need to set.

chrisdembia:
This PR fixes #558: the path from the SimTKsimbody library to the simbody-visualizer is added as a search path in VisualizerProtocol. To facilitate this, I added the function `Pathname::getFunctionLibraryDirectory()`. I also added a small test, and updated the documentation.

I tested this locally on both macOS and Ubuntu 14.04. I was able to rename the install directory and successfully get ChainExample to spawn the visualizer; this behavior was not possible before this PR. I also tested this PR with OpenSim on macOS and was able to verify that OpenSim's MATLAB bindings are able to find simbody-visualizer even if the original simbody install directory no longer exists.

Unfortunately, I realized that this PR may not be sufficient for OpenSim's Python bindings to find simbody-visualizer. This is more of an OpenSim issue though (it depends on how we expect the Python bindings to find the SimTK libraries; do we copy them into the Python package or edit DYLD_LIBRARY_PATH/RPATH?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,10 @@ to ParallelExecutor, mutex state lock.
 * Started using RPATH on OSX so that users need not set `DYLD_LIBRARY_PATH` to
   run `simbody-visualizer` or the example executables, regardless of where you
   install Simbody.
+* Improved the ability to find the simbody-visualizer executable when Simbody
+  is installed in non-standard locations or if the Simbody installation is
+  relocated (even to different computers). This enhancement is only for
+  non-Windows operating systems.
 * Fixed a bug when compiling on macOS (OSX) with SDK MacOSX10.12.sdk, related
   to the POSIX function `clock_gettime()`.
   [Issue #523](https://github.com/simbody/simbody/issues/523),
--- a/SimTKcommon/include/SimTKcommon/internal/Pathname.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Pathname.h
@@ -265,6 +265,23 @@ public:
     /// Get the absolute pathname of the directory which contains the 
     /// currently executing program.
     static std::string getThisExecutableDirectory();
+    /// Get the absolute pathname of the directory which contains the
+    /// library/binary from which func was loaded. It is expected that func
+    /// is defined in a dynamically-linked library; if func was statically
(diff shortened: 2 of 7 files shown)


chrisdembia:
Thank you!

Issue #522, opened 2016-08-30
Some simbody 3.5.3 tests fail on Ubuntu 14.04 LTS


wagglefoot:
Following 

`ctest -j8`

we get

```
98% tests passed, 5 tests failed out of 202

Total Test time (real) =  32.44 sec

The following tests FAILED:
	162 - TestReverseMobilizersStatic (Failed)
	174 - GazeboReactionForceWithAppliedForceRigidStatic (Failed)
	175 - TestCustomConstraints (Failed)
	176 - TestCustomConstraintsStatic (Failed)
	190 - GazeboReactionForceWithAppliedForceCompliantStatic (Failed)
Errors while running CTest
```

This is on a Ubuntu machine running 14.04 with `cmake version 3.2.2`.

chrisdembia:
Thank you for sharing. What compiler (and version) did you use? It is suspicious that the static version of some tests fail, but the shared version passes. Is this from a fresh build directory? Did you perhaps update any software on your computer between when you created the build directory and now? What do the tests output? `ctest -j8 --output-on-failure`

sherm1:
I notice these are the "static" versions of the tests (i.e. linked with static libraries). I don't think we have tried those for a long time. Do you get the same failures with the dynamic tests? I believe the static tests require some `#define` magic -- maybe those particular tests don't have it.

wagglefoot:
@chrisdembia this all happened within the space of a few hours and I have the most recent versions of cmake and swig (not that it is used). And they were updated today before I downloaded the source (today).

Alas;

```
clang --version
Ubuntu clang version 3.4-1ubuntu3 (tags/RELEASE_34/final) (based on LLVM 3.4)
Target: x86_64-pc-linux-gnu
Thread model: posix
```

```
gcc --version
gcc (Ubuntu 4.9.3-8ubuntu2~14.04) 4.9.3
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

clnsmith:
Hi all,

I'm building on Scientific Linux 6 using GCC 4.8.2 and Lapack 3.6.1 and I am getting the same failed tests:

The following tests FAILED:
	164 - GazeboReactionForceWithAppliedForceRigidStatic (Failed)
	172 - TestReverseMobilizersStatic (Failed)
[...]

PR #568, opened 2017-07-22
[static] Allow only static compilation of simbody

Gjacquenot:
This PR adds the possibility to compile simbody in a all static way

We use it with 

    cmake -DBUILD_DYNAMIC_LIBRARIES:BOOL=OFF -DBUILD_STATIC_LIBRARIES:BOOL=ON ....

A message error is thrown in case these two options are set to `FALSE`...

Reason of this PR:

When needing only static version, we can avoid dynamic compilation (save 50% compilation time). Also  dynamic compilation requires to have blas/lapack to be compiled with -fPIC in case of use of static version.
Disabling dynamic compilation allows to use static versions of blas/lapack coming from Linux distro, that do not use -fPIC for static libraries on 64 bit platform.

1) This PR does not change anything to the previous configuration and compilation steps for simbody

2) With static compilation, all tests do not succeed, because of tolerance problem

-  81 - GazeboReactionForceWithAppliedForceRigidStatic (Failed)
-  95 - GazeboReactionForceWithAppliedForceCompliantStatic (Failed)
- 104 - TestReverseMobilizersStatic (Failed)

One can see a log of these errors [here](https://travis-ci.org/Gjacquenot/simbody/jobs/256158634)

I have discussed this with @chrisdembia , who told me that this problem was known and already reported [here](https://github.com/simbody/simbody/issues/522).

3) Finally, I have fixed a broken mingw table in README.md.


chrisdembia:
I think it would be nice to be able to build only the static libraries. I did a quick pass over the changes and they mostly look good. I can do an actual review in about a week.

Gjacquenot:
Hi @chrisdembia , I have taken into account your remarks.

Gjacquenot:
For time being, static versions of blas and lapack should be found by CMake using flag `BLA_STATIC` before `find_package`. Unfortunately, this CMake flag does not work, no library is found. When specifying manually the blas and lapack libraries a naive way, it leads to linking problem, like `multiple definition of `xerbla_'`. One needs something like `-Wl,--start-group $(LAPACK)/blas.a $(LAPACK)/lapack.a -Wl,--end-group`, plus libgfortran.a. I will post a Dockerfile, where one links statically againt lapack, blas and libgfortran, I already did it a long time ago...

chrisdembia:
> @chrisdembia, did you get a chance to pull this and try the build?

No, good point. I'll try this out soon.

sherm1:
Thanks!

chrisdembia:
I just tested this locally on macOS and it worked great. The installation worked fine when setting `BUILD_DYNAMIC_LIBRARIES` to `OFF`. As expected, setting both `BUILD_DYNAMIC_LIBRARIES` and `BUILD_STATIC_LIBRARIES` to `OFF` caused CMake to generate an error message.

The only thing I noticed is that binary examples are only installed if dynamic libraries are built. I think this is fine, especially since the source code for the examples is always installed.

LGTM.

PR #570, opened 2017-08-15
Do not store Force* in State cache.

### LINKED ISSUE #567: States cannot be used with different Systems
When we added support for parallel calculation of forces, we coupled State objects to the System they came from, by caching `Force*`s from the System in the State (see [here](https://github.com/simbody/simbody/blob/master/Simbody/src/GeneralForceSubsystem.cpp#L643) and [here](https://github.com/simbody/simbody/blob/master/Simbody/src/GeneralForceSubsystem.cpp#L713)).  This prevents users from easily using a state from one instance of a System with a different but identical instance of the System.

Related: https://github.com/opensim-org/opensim-core/issues/1455, https://github.com/opensim-org/opensim-core/issues/1785

chrisdembia:
Fixes #567.

This PR replaces the use of `Force*`s in the state cache with `ForceIndex`, as suggested by @sherm1 . I tested that these changes work fine with OpenSim, and that performance is not affected.

This PR will allow me to make a fix to `OpenSim::StatesTrajectory`, and may help @mitkof6 with some issues he had in OpenSim.

### DIFF
--- a/Simbody/src/GeneralForceSubsystem.cpp
+++ b/Simbody/src/GeneralForceSubsystem.cpp
@@ -58,16 +58,16 @@ public:
     virtual CalcForcesTask* clone() const = 0;
     
     virtual void initializeAll(
-            const State& s,
-            const Array_<Force*>& enabledNonParallelForces,
-            const Array_<Force*>& enabledParallelForces,
+            const Array_<Force*>& forces, const State& s,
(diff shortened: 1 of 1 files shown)


chrisdembia:
Thanks @sherm1 for the quick review.

mitkof6:
Thanks @chrisdembia I will give it a try.

Issue #574, opened 2017-08-30
std::memcpy does not work with ADOL-C type adouble


antoinefalisse:
We are trying to bring automatic differentiation to Simbody using ADOL-C. We have observed that std::memcpy does not work with adoubles, since this will copy the location in the tape and results in a shallow copy rather than a deep copy. Instead of using std::memcpy, we have seen that we could use std::copy which works with adoubles. We read that std::copy would delegate to std::memcpy when it is appropriate. We therefore would like to change the calls to std::memcpy by std::copy.

sherm1:
:+1: Good idea!

PR #575, opened 2017-08-30
repalce std::memcpy by std::copy

antoinefalisse:
We are trying to bring automatic differentiation to Simbody using ADOL-C. We have observed that std::memcpy does not work with adoubles, since this will copy the location in the tape and results in a shallow copy rather than a deep copy. Instead of using std::memcpy, we have seen that we could use std::copy which works with adoubles. We read that std::copy would delegate to std::memcpy when it is appropriate. We therefore would like to change the calls to std::memcpy by std::copy. Note that building the libraries in Debug mode (not with the other modes) gives a warning related to the use of std::copy.

Issue #574 
@sherm1 @chrisdembia


chrisdembia:
@antoinefalisse thanks for mentioning the warnings. We try to keep the codebase clean of any warnings; could you paste the exact warning?

antoinefalisse:
@chrisdembia thanks for replying.

Here is the warning:

> warning C4996: 'std::copy::_Unchecked_iterators::_Deprecate': Call to 'std::copy' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'

I read that `stdext::checked_array_iterator` can be used to wrap the unsafe array pointers ([see this link](https://codeyarns.com/2010/08/30/visual-c-c4996-warning-on-copy-with-array-parameters/)).

In practice, we could, for instance (MatrixHelperRep_Vector.h line 284), replace

`std::copy(this->m_data, this->m_data + this->length()*this->m_eltSize, p->m_data);` 

by

`std::copy(this->m_data, this->m_data +   this->length()*this->m_eltSize, stdext::checked_array_iterator<S*>(p->m_data, this->length()*this->m_eltSize));`

I have applied that everywhere I used `std::copy` and there are no more warning in Debug mode.

Would that be an appropriate approach?

chrisdembia:
That looks like a good solution except that `stdext::checked_array_iterator` is only available with Visual C++. [This page](https://stackoverflow.com/questions/25716841/checked-array-iteratort-in-c11) is about handling `stdext::checked_array_iterator` in a cross-platform way, but it does not have any good solutions.

You could go back to defining a `SimTK::copyArray` function that has a platform-dependent implementation:

```cpp
template< class InputIt, class OutputIt >
OutputIt SimTK::copyArray(InputIt first, int length, OutputIt d_first ) {
    return std::copy(first, first + length, 
    #ifdef _WIN32
        stdext::make_checked_array_iterator(d_first, length)
    #else
        d_first
    #endif
    );
}
```

I'm not sure about the best type for the `length` parameter.

If only these 3 headers need this function, you could put it in some header that these 3 files share (rather than in common.h).

sherm1:
I think it would be better just to locally suppress the 4996 warning in this case. I don't think we would want to use checked iterators anyway since they can be slow. We have examples of that suppression already in Simbody, for example [here](https://github.com/simbody/simbody/blob/master/SimTKmath/src/MultibodyGraphMaker.cpp#L40).

chrisdembia:
> I think it would be better just to locally suppress the 4996 warning in this case.

Ah that sounds good. Since this warning appears in headers, the warning could be pushed at the top of the header and popped at the bottom, as described [here](https://msdn.microsoft.com/en-us/library/2c8f766e.aspx) and done in Simbody ([push](https://github.com/simbody/simbody/blob/3469006e99a94e9a3c7201a5308d298d20acf07f/SimTKmath/include/simmath/internal/common.h#L86) and [pop](https://github.com/simbody/simbody/blob/3469006e99a94e9a3c7201a5308d298d20acf07f/SimTKmath/include/simmath/internal/common.h#L182)). Note, however, that it seems people may have had [trouble disabling this warning for std::copy](https://stackoverflow.com/questions/25716841/checked-array-iteratort-in-c11#comment40434549_25716929).

chrisdembia:
See https://stackoverflow.com/questions/38060411/visual-studio-2015-wont-suppress-error-c4996 for how to disable this particular warning. It seems that the pragma for C4996 may not work.

chrisdembia:
@antoinefalisse I think the idea would be to make the following change [here](https://github.com/antoinefalisse/simbody/blob/5da70e669b8734c9da050b3ddbad9d94771c2b1e/SimTKcommon/include/SimTKcommon/internal/common.h#L260):

```cpp
#define _SECURE_SCL_DEPRECATE 0
#include <algorithm>
#undef _SECURE_SCL_DEPRECATE
```

or, preferably:

```cpp
#define _SCL_SECURE_NO_WARNINGS
#include <algorithm>
#undef _SCL_SECURE_NO_WARNINGS
```

chrisdembia:
I would first try and see if the 4996 pragma works.

sherm1:
There are some other things to try. `-DNO_WARN_MBCS_MFC_DEPRECATION` looks promising. 

And apparently the warning has been [removed in VS2017](https://blogs.msdn.microsoft.com/vcblog/2013/07/08/mfc-support-for-mbcs-deprecated-in-visual-studio-2013).

antoinefalisse:
The only way I managed to get rid of the warning is by adding `#define _SCL_SECURE_NO_WARNINGS` [here](https://github.com/simbody/simbody/blob/d81faaeea0a7d6cc5d46510dd2c0122231d42d43/SimTKcommon/include/SimTKcommon/internal/common.h#L123) (right before `#include <cstdio>`). The other options did not seem to work ( 4996 pragma, `NO_WARN_MBCS_MFC_DEPRECATION`, `#define _SCL_SECURE_NO_WARNINGS` before `#include <algorithm>`, `#define _SECURE_SCL_DEPRECATE 0`before `#include <algorithm>`).

sherm1:
If that works, it would be fine instead to add `-D_SCL_SECURE_NO_WARNINGS` unconditionally to the CMake build instead of modifying the code. We don't ever want those warnings and that define will be ignored except in Visual Studio. I'm not sure exactly where that would go but I think @chrisdembia would know.

chrisdembia:
> I'm not sure exactly where that would go but I think @chrisdembia would know.

I'll talk with @antoinefalisse about the best place for this.

SimTKcommon/sharedTarget/CMakeLists.txt
```diff
@@ -15,7 +15,11 @@ if(BUILD_UNVERSIONED_LIBRARIES)
     set_target_properties(${SHARED_TARGET} PROPERTIES
         PROJECT_LABEL "Code - ${SHARED_TARGET}"
         SOVERSION ${SIMBODY_SONAME_VERSION})
-
+		
+    if(WIN32 AND ${CMAKE_VERSION} GREATER 3.0)
```
chrisdembia: Could you add a comment here and in the staticTarget CMakeLists explaining why this was introduced? Perhaps "Warning 4996 is emitted when using std::copy with raw pointers, as is done in MatrixHelperRep*.h" You could use the longer explanation in the root CMakeLists.txt as well.

Three other changes you could consider: 

1. You could make the version check like [this one](https://github.com/simbody/simbody/blob/master/SimTKcommon/staticTarget/CMakeLists.txt#L25):

```
if(WIN32 AND NOT ${CMAKE_VERSION} VERSION_LESS 3.0.2)
```

2. Also, perhaps say that the purpose of this line is so that downstream targets do not see the warning.

3. Since this warning is emitted only in debug mode, you could use the following so that the flag is only specified when building in debug:

```
target_compile_options(${SHARED_TARGET} INTERFACE
    "$<$<CONFIG:Debug>:/wd4996>")
```

antoinefalisse:
We edited the main [CMake file](https://github.com/antoinefalisse/simbody/blob/995a3ad2c9e11675b0059e812721cccf6c26bb70/CMakeLists.txt#L433) and the warning is now gone.

SimTKcommon/sharedTarget/CMakeLists.txt
antoinefalisse: Ok, I'll do that.

chrisdembia:
The travis test is failing because you inserted tab characters into the CMakeLists. Consider changing your text editor settings to use spaces instead of tabs.

chrisdembia:
🎆 the first small step :)

antoinefalisse:
Great!

PR #583, opened 2017-09-30
Avoid Winstantiation-after-specialization from AppleClang9

### LINKED ISSUE #581: Clang warning about ignored explicit instantiation of Measure_<Vector>::Zero
Apple Clang 9 gives the following warning:

```cpp
/Users/chris/repos/simbody/4/simbody/SimTKcommon/Simulation/src/Measure.cpp:62:1: warning:
      explicit instantiation of 'Zero' that occurs after an explicit specialization
      has no effect [-Winstantiation-after-specialization]
template class Measure_<Vector>::Zero;
^
/Users/chris/repos/simbody/4/simbody/SimTKcommon/Simulation/include/SimTKcommon/internal/Measure.h:376:27: note:
      previous template specialization is here
class Measure_< Vector >::Zero : public Measure_< Vector >::Constant {

```

I could remove the explicit instantiations, but are they still useful with MSVC, considering that MSVC does not have two-phase lookup (although, newer versions of MSVC will have two-phase lookup)?

chrisdembia:
This PR fixes #581 by following @sherm1's suggestion to suppress the warning.

### DIFF
--- a/SimTKcommon/Simulation/src/Measure.cpp
+++ b/SimTKcommon/Simulation/src/Measure.cpp
@@ -35,6 +35,12 @@
 #include <cassert>
 #include <algorithm>
 
+#if defined(__clang__) && __has_warning("-Winstantiation-after-specialization")
+    // Avoid the harmless warning "explicit instantiation of 'Zero' that
+    // occurs after an explicit specialization has no effect"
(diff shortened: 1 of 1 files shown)


chrisdembia:
Yeah I just learned about it too. Really  neat.

PR #587, opened 2017-10-02
Replace ThreadLocal with c++11 thread_local

chrisdembia:
This PR replaces `SimTK::ThreadLocal` with C++11's `thread_local`. 

Unfortunately, the Travis CI test fails on OSX because only Apple Clang 8 and up support this keyword (https://stackoverflow.com/questions/28094794/why-does-apple-clang-disallow-c11-thread-local-when-official-clang-supports/29929949#29929949).

That means that users will either need OSX Sierra (10.12) or need to use normal clang (not Apple Clang), perhaps via `brew install llvm`.

Related: #423  #422

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/587)


chrisdembia:
I think that it would be nice to support older versions of macOS, so I will make the following change: I'll leave in the `SimTK::ThreadLocal` class template but it will be implemented using C++11 `thread_local` on all platforms except older macOS (where it'll use pthreads). I'll have to add the constraint that `SimTK::ThreadLocal` can only be a member variable if it's static.

chrisdembia:
I'm not sure if this limitation on `thread_local` means that we can't run binaries on older versions of macOS if the binaries were built using `thread_local`. I'm fine requiring developers to use 10.12, but not fine requiring users to have 10.12.

sherm1:
> I'm fine requiring developers to use 10.12, but not fine requiring users to have 10.12.

I trust your judgment on this: are you sure it's worthwhile going through this pain to support backlevel Macs? By the time this is released, won't most people have upgraded?

chrisdembia:
I saw Jeff Reinbolt at a conference this summer and he had an older version of macOS/OSX (maybe even 10.9), and I think he'd like to run OpenSim.

Also, I'll try to find an older mac to do some testing on (to see if binaries that uses `thread_local` work on pre-10.12).

scpeters:
> I saw Jeff Reinbolt at a conference this summer and he had an older version of macOS/OSX (maybe even 10.9), and I think he'd like to run OpenSim.

For reference, Homebrew's policy is to support the three most recent releases of macOS/OSX, so they have just dropped support for 10.10. I've been following that same approach with our software at Open Robotics.

chrisdembia:
Oh; thanks for that info. Sounds like a reasonable policy.

scpeters:
> Also, I'll try to find an older mac to do some testing on (to see if binaries that uses thread_local work on pre-10.12).

You can target older versions of macOS with travis:

* https://docs.travis-ci.com/user/reference/osx/#OS-X-Version

You could probably modify the travis.yml file in a test branch if you didn't want to change your default CI settings. Also, travis has a limited number of mac executors; there's a bit of a wait during the daytime hours for those builds:

* https://www.traviscistatus.com/

scpeters:
Oh, I guess you already knew about the travis osx support: db1842cd227a3be8dcaa040457592d08474b9731

chrisdembia:
Thanks @scpeters .

I've ensured that building with Xcode 8 produces binaries that can run on macOS/OSX 10.11 (both using Travis and locally).

I've edited the README to list Xcode 8 as the minimum required version.

This is ready for review.

chrisdembia:
@sherm1 let me know if there's anything you'd like me to do before you'd be interested in reviewing this.

sherm1:
@chrisdembia this is ready to merge - :lgtm:.

---

Reviewed 12 of 14 files at r1, 2 of 2 files at r3, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/587#-:-L1ZoMenfFvIb3KBR2zO:b8ns215)*

chrisdembia:
Thanks @sherm1

Issue #537, opened 2016-11-14
MultibodyGraphMaker incorrectly handles an intermediate massless body


aseth1:
This issue was first raised in OpenSim [here](https://github.com/opensim-org/opensim-core/issues/1388)

The translation of a body (pelvis) is defined by two sliders. The first slider mobilizes a massless body and the second slider mobilizes the full mass of the pelvis. Two massive bodies (right and left thighs) mobilized by pins are connected to the pelvis.  

The MultigraphMaker introduces two FreeJoints and two slave bodies (a pair for each thigh)  which seems incorrect :
```
MULTIBODY GRAPH
---------------

10 BODIES:
 0  0: ground mass=inf mob=-1 master=-1
  jointsAsParent=[ 0 7 8]         jointsAsChild=[]        slaves=[]
 1  1: Platform mass=1 mob=0 master=-1
  jointsAsParent=[ 1]     jointsAsChild=[ 0]      slaves=[]
 2  2: Pelvis_massless mass=0 mob=1 master=-1
  jointsAsParent=[ 2]     jointsAsChild=[ 1]      slaves=[]
 3  3: Pelvis mass=0.5 mob=2 master=-1
  jointsAsParent=[ 3 4]   jointsAsChild=[ 2]      slaves=[]
 4  1: LeftThigh mass=0.05 mob=3 master=-1
  jointsAsParent=[ 5]     jointsAsChild=[ 3 7]    slaves=[ 8]
 5  1: RightThigh mass=0.05 mob=5 master=-1
  jointsAsParent=[ 6]     jointsAsChild=[ 4 8]    slaves=[ 9]
 6  2: LeftShank mass=0.05 mob=4 master=-1
  jointsAsParent=[]       jointsAsChild=[ 5]      slaves=[]
 7  2: RightShank mass=0.05 mob=6 master=-1
  jointsAsParent=[]       jointsAsChild=[ 6]      slaves=[]
 8  4: #LeftThigh_slave_1 mass=nan mob=7 master=4
  jointsAsParent=[]       jointsAsChild=[ 3]      slaves=[]
 9  4: #RightThigh_slave_1 mass=nan mob=8 master=5
  jointsAsParent=[]       jointsAsChild=[ 4]      slaves=[]

9 JOINTS:
 0  0:     PlatformToGround               ground->Platform              WeldJoint loopc=-1
 1  1: PelvisMasslessToPlatform             Platform->Pelvis_massless      SliderJoint loopc=-1
[...]

PR #592, opened 2017-11-29
Fix MultibodyGraphMaker bug when handling massless bodies

sherm1:
MultibodyGraphMaker takes bodies and joints and finds a spanning tree. It works breadth-first by adding all the base bodies (level 1, bodies with joints to Ground), then all the bodies with joints to base bodies (level 2), etc. until it reaches a level where it can't find any body. If there are still bodies left one of them needs to be connected to Ground by a free joint, then we can resume.

A massless body modifies this algorithm because we can't end a branch with one. So if we encounter a massless body at level n, rather than wait for level n+1's turn, we continue on lengthening the branch top n+1, n+2, etc. until we hit a massful body. Then we go back to the rest of the level n bodies and move up to n+1.

This PR fixes a bug where the algorithm did not properly keep track of the fact that, while working on level n, it may also have added some bodies at n+1 or higher due to massless bodies. So then when moving on to n+1 it would decide there were no bodies remaining to be added. That meant it would go try to make a new base body by adding a free joint.

The proposed fix is to keep track of all the joints that were added so we can check when at level n+1 if an already-mobilized body was actually added during this same tree sweep and if so count that as a success and keep going.

Resolves issue #537 reported by @aseth1.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/592)

### DIFF
--- a/SimTKmath/include/simmath/MultibodyGraphMaker.h
+++ b/SimTKmath/include/simmath/MultibodyGraphMaker.h
@@ -568,6 +568,8 @@ public:
     defined in the input joint. In that case you should use a reverse joint
     when you build the system. **/
     bool isReversedFromJoint() const {return isReversed;}
+	/** Return the level of the outboard body (Ground is level 0) **/
+	int getLevel() const {return level;}
 
 private:
 friend class MultibodyGraphMaker;
--- a/SimTKmath/include/simmath/MultibodyGraphMaker.h
+++ b/SimTKmath/include/simmath/MultibodyGraphMaker.h
@@ -568,8 +568,8 @@ public:
     defined in the input joint. In that case you should use a reverse joint
     when you build the system. **/
     bool isReversedFromJoint() const {return isReversed;}
-	/** Return the level of the outboard body (Ground is level 0) **/
-	int getLevel() const {return level;}
+    /** Return the level of the outboard body (Ground is level 0) **/
+    int getLevel() const {return level;}
 
 private:
 friend class MultibodyGraphMaker;
--- a/SimTKmath/src/MultibodyGraphMaker.cpp
+++ b/SimTKmath/src/MultibodyGraphMaker.cpp
@@ -613,9 +613,9 @@ findHeaviestUnassignedReverseJoint(int inboardBody) const {
 // TODO: keep a list of unprocessed joints so we don't have to go through
 // all of them again at each level.
 void MultibodyGraphMaker::growTree() {
-    // Record the joints we added during this subtree sweep, and the level
-    // at which they were added. That way if we jumped levels ahead due to
-    // massless bodies we can take credit and keep going rather than quit.
+    // Record the joints for which we added mobilizers during this subtree
+    // sweep. That way if we jumped levels ahead due to massless bodies we
+    // can take credit and keep going rather than quit.
     std::set<int> jointsAdded;
     for (int level=1; ;++level) { // level of outboard (mobilized) body
         bool anyMobilizerAdded = false;
--- a/SimTKmath/src/MultibodyGraphMaker.cpp
+++ b/SimTKmath/src/MultibodyGraphMaker.cpp
@@ -613,27 +613,27 @@ findHeaviestUnassignedReverseJoint(int inboardBody) const {
 // TODO: keep a list of unprocessed joints so we don't have to go through
 // all of them again at each level.
 void MultibodyGraphMaker::growTree() {
-	// Record the joints we added during this subtree sweep, and the level
-	// at which they were added. That way if we jumped levels ahead due to
-	// massless bodies we can take credit and keep going rather than quit.
-	std::set<int> jointsAdded;
+    // Record the joints we added during this subtree sweep, and the level
+    // at which they were added. That way if we jumped levels ahead due to
+    // massless bodies we can take credit and keep going rather than quit.
+    std::set<int> jointsAdded;
     for (int level=1; ;++level) { // level of outboard (mobilized) body
         bool anyMobilizerAdded = false;
         for (int jNum=0; jNum<getNumJoints(); ++jNum) {
             // See if this joint is at the right level (meaning its inboard
             // body is at level-1) and is available to become a mobilizer.
             const Joint& joint = getJoint(jNum);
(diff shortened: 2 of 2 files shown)


sherm1:
@aseth1 for testing and review, please.
@chrisdembia or @tkuchida, can you review this and verify that it fixes the problem?

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592#-:-L-5B7uBn4rCGYUaVDGo:b-kasttp)*

aseth1:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592)*

aseth1:
:lgtm:
I don't have a good enough grasp on the graph making algorithm to discern if it will work for cases where there are multiple massless bases that have massive leaves, but it looks reasonable. Patching OpemSim with this fix resolves the [issue](https://github.com/opensim-org/opensim-core/issues/2024#issuecomment-347760497).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592#-:-L-8N62mKjC13KYa5Zzy:b-79opxw)*

sherm1:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SimTKmath/src/MultibodyGraphMaker.cpp, line 614 at r1](https://reviewable.io:443/reviews/simbody/simbody/592#-L-5AjtYm28ZUmgM6wDf-r1-614:-L-AHxFUzZPtrpPVQgL6:bylknwt) ([raw file](https://github.com/simbody/simbody/blob/7046f85d9a0d5227ca4964b96e333c8f651a1e55/SimTKmath/src/MultibodyGraphMaker.cpp#L614)):*

Is it possible that this TODO could be removed now?
</blockquote></details>

Good question. Despite appearances the code I added doesn't address this TODO. It is still processing all n joints for each level in the tree. For a chain of n bodies there are n level so that's n² complexity. OTOH it quickly rejects joints it can't process so although it is O(n²) it is still very fast for most systems of interest. I suspect it would be painfully slow for a system of say n=1000 and we could reduce it to O(n) with fancier data structures so I still think the TODO still makes sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592)*

chrisdembia:
@sherm1 I'm willing to write test case(s) for this PR if you have any suggestions for specific topologies and conditions to check for.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SimTKmath/src/MultibodyGraphMaker.cpp, line 614 at r1](https://reviewable.io:443/reviews/simbody/simbody/592#-L-5AjtYm28ZUmgM6wDf-r1-614:-L-ETe4-L9DYLXu_0FMH:beywc9o) ([raw file](https://github.com/simbody/simbody/blob/7046f85d9a0d5227ca4964b96e333c8f651a1e55/SimTKmath/src/MultibodyGraphMaker.cpp#L614)):*

Good question. Despite appearances the code I added doesn't address this TODO. It is still processing all n joints for each level in the tree. For a chain of n bodies there are n level so that's n² complexity. OTOH it quickly rejects joints it can't process so although it is O(n²) it is still very fast for most systems of interest. I suspect it would be painfully slow for a system of say n=1000 and we could reduce it to O(n) with fancier data structures so I still think the TODO still makes sense.
</blockquote></details>

Okay thanks for explaining.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592#-:-L-ETgTXMQRMn2CkGsYg:b752qrp)*

sherm1:
@chrisdembia that would be awesome - thanks! There really should be a separate regression test for MultibodyGraphMaker but there isn't one -- so far I hand-tested it using Gazebo2Simbody and some sdf files, and Ajay patched it into OpenSim and ran the test suite successfully.

It would be great if you are willing to add a TestMultibodyGraphMaker regression test (in simbody/SimTKmath/tests). Even if it just includes a couple of test cases it would then be very easy to add more once the infrastructure is there. To catch the presenting problem, we would just need to show that (for example) a 5-body chain of pin joints (ground-body1-body2-body3-body4-body5) produces the same topology whether body 2 is massless or not. We could then backfill with some basic tests (chain, tree, loop) either now or later, depending on how much energy you have for this. There are more-esoteric features that definitely should have had unit tests from the start, but I didn't know that at the time I wrote MBGraphMaker.

Would be easiest if I merge the current PR so you can add the test in a separate one? I think it is safe enough since OpenSim is the primary user and its tests pass.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592#-:-L-EXGPhlr63L5PE0UEQ:b11lap0)*

chrisdembia:
@sherm1 that all sounds great (including merging this PR). Thanks for the guidance. I put the test case under O for my AOI.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592#-:-L-Eiq4D45nzi4hUsrWW:bfh5fra)*

sherm1:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SimTKmath/src/MultibodyGraphMaker.cpp, line 614 at r1](https://reviewable.io:443/reviews/simbody/simbody/592#-L-5AjtYm28ZUmgM6wDf-r1-614:-L-EpDwkGbpsr-rLfQ1_:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/7046f85d9a0d5227ca4964b96e333c8f651a1e55/SimTKmath/src/MultibodyGraphMaker.cpp#L614)):*

Okay thanks for explaining.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/592)*

PR #600, opened 2017-12-19
ADOLC Common

antoinefalisse:
+@sherm1, +@chrisdembia. This PR contains the changes in `common.h` that will enable setting`Real `as `adouble` when building Simbody with ADOLC. However, for now, `Real `is still set to `double ` so that we can build Simbody with ADOLC in CI and have tests. There are a couple of additional minor changes in `common.h`. This PR also contains some changes in the CMake files that were needed to copy the ADOLC dll that is needed in tests relying on ADOLC drivers. `simbody_cmake_macros` has been directly copied from one of @chrisdembia's projects.  Finally, this PR contains a simple test case that does some trivial use of `adouble `, ie it computes the derivative of a simple function. Please let me know if you would like to see other tests in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/600)

PR #603, opened 2018-01-03
ADOLC Ntraits

antoinefalisse:
+@Sherm1, +@chrisdembia. This PR contains the changes in `NTraits.h` necessary to build Simbody with ADOLC. Since a new exception was introduced and used in `NTraits.h`, I also included `Exception.h` and `ExceptionMacros.h` in the PR. This PR also contains various tests in `TestADOLCCommon.cpp` verifying that `NTraits<adouble>` works properly. I might need some guidance to design additional tests to make sure all is fine. Note that this PR is based on another PR (ADOLC common #600 ) that is still in review. Therefore, changes in other files should disappear when that former PR is merged.

I have attempted to include all previous comments. In particular:
- I have followed the suggestions of @Sherm1 regarding `isNumericallyEqual()`.
- I have tried to adjust the comments to take ADOL-C cases into account. I do not know whether we want to specifically mention each additional case or just mention that more cases exist when using ADOL-C. I have chosen the second approach.
- I wanted to include free function templates to reduce text as suggested by @chrisdembia but it did not work out at first attempt. I had errors when `T=SimTK::Vector` (or `RowVector`, `Matrix`). This mostly occurred in tests but also for example in Assembler (line 568 with `abs`). I also had errors related to ambiguous calls (eg `TestBicubicSurface` line 585 with `min`) . I had defined the templates for example as follows: `template <typename T> T abs(const T& v) { return NTraits<T>::abs(v); }`. I decided not to include the free function templates for now but this is open to discussion.

Here are a few additional remarks:
- When testing `isNumericallyEqual(complex<float>,adouble)` and `isNumericallyEqual(conjugate<float>,adouble)`, I had to specify a tolerance to get it working. Tolerances smaller than 1e-7 resulted in failure. I am not sure whether this is expected.
- To avoid warnings related to ADOL-C, I included the same few lines in `ExceptionMacros.h` (where `adolc/adolc.h` is included) as in `common.h` (where `adolc/adouble.h` is included). This seems like code duplication but I do not know how else to do.
- There are still lines longer than 80 characters. I did it on purpose in cases like `Widest` . Please let me know if I should change that.

Questions:
- For unit tests, do you want them all in different cpps or I can just add them in `TestADOLCCommon.cpp` as it is for now?

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/603)

### DIFF
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1022,14 +1022,12 @@ SimTK_NTRAITS_CONJ_SPEC(double,float);SimTK_NTRAITS_CONJ_SPEC(double,double);
 // no support for conjugate<adouble> at the moment. However, these structs
 // are necessary for compilation.
 #ifdef SimTK_REAL_IS_ADOUBLE
-    template<> template<> struct NTraits< conjugate<float> >::Result<adouble> {
-        typedef conjugate<Widest<float,adouble>::Type> W;
-        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
-    };
+    template<> template<> struct NTraits< conjugate<float> >::Result<adouble>
+    {   typedef conjugate<Widest<float,adouble>::Type> W;
+        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub; };
     template<> template<> struct NTraits< conjugate<double> >::Result<adouble>
     {   typedef conjugate<Widest<double,adouble>::Type> W;
-        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
-    };
+        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub; };
 #endif
 
 
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -153,8 +153,8 @@ template <> struct Narrowest<float,double>             {typedef float  Type; typ
 template <> struct Narrowest<double,float>             {typedef float  Type; typedef float Precision;};
 template <> struct Narrowest<double,double>            {typedef double Type; typedef double Precision;};
 #ifdef SimTK_REAL_IS_ADOUBLE
-    /// Be careful that Narrowest(float,adouble) is adouble since it is the
-    /// narrowest type than can hold the value and the derivative.
+    /// Be careful that Narrowest(float,adouble)::Type is adouble since it is
+    /// the narrowest type than can hold the value and the derivative.
     template <> struct Narrowest<float,adouble>
       {typedef adouble Type; typedef float Precision;};
     template <> struct Narrowest<adouble,float>
@@ -1401,8 +1401,8 @@ template <> class CNT<double> : public NTraits<double> { };
         static bool isNaN   (const T& t) {return SimTK::isNaN(t);}
         static bool isInf   (const T& t) {return SimTK::isInf(t);}
         /* Methods to use for approximate comparisons. Perform comparison in
-        /* the wider of the two precisions, using the default tolerance from
-        /* the narrower of the two precisions. */
+        the wider of the two precisions, using the default tolerance from
+        the narrower of the two precisions. */
         static double getDefaultTolerance()
         {return RTraits<T>::getDefaultTolerance();}
         static bool isNumericallyEqual(const T& t, const float& f)
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1240,14 +1240,24 @@ template <> class CNT<double> : public NTraits<double> { };
 } // namespace SimTK
 
 #ifdef SimTK_REAL_IS_ADOUBLE
-    // Must declare certain friend functions of adouble in the global namespace
-    // so that they are visible within NTraits<adouble>.
+    // If we implement NTraits::abs(const T& t) as { abs(t); }, then we get an
+    // infinite recursion. Therefore, we use { ::abs(t); } so that the compiler
+    // can find the friend functions that ADOL-C declares in adouble.h and
+    // internal/adubfunc.h. However, this is not sufficient: the standard says,
+    // "A name prefixed by the unary scope operator :: ... shall be declared in
+    // global namespace scope or shall be a name whose declaration is visible
+    // in global scope because of a using-directive." ADOL-C does not declare
+    // these functions in global namespace scope (the functions are declared
+    // within a class, at the same time they are marked as friends), so we must
+    // make them "visible in global scope [with] a using-directive."
     // https://stackoverflow.com/questions/18449855/c-friend-function-hidden-by-class-function/18451369
     adub sqrt(const badouble&);
     adub sin(const badouble&);
     adub cos(const badouble&);
     adub floor(const badouble&);
     adouble pow(const badouble&, const badouble&);
+    adouble pow(double, const badouble&);
+    adub pow(const badouble&, double);
     adub exp(const badouble&);
     adub log(const badouble&);
     adub tan(const badouble&);
@@ -1361,7 +1371,9 @@ template <> class CNT<double> : public NTraits<double> { };
         static T    sin(const T& t)     {return ::sin(t);}
         static T    cos(const T& t)     {return ::cos(t);}
         static T    floor(const T& t)   {return ::floor(t);}
-        static T    pow(const T& t, const T& order) {return ::pow(t, order);}
+        template <typename TBase, typename TExp>
+        static T pow(const TBase& base, const TExp& exponent)
+            {return ::pow(base, exponent);}
         static T    exp(const T& t)     {return ::exp(t);}
         static T    log(const T& t)     {return ::log(t);}
         static T    tan(const T& t)     {return ::tan(t);}
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -31,8 +31,8 @@
  * adouble if using ADOL-C but there is no support for complex<adouble> and
  * conjugate<adouble> at the moment). Each of these may be modified by a
  * negator, which does not change the in-memory \e representation but negates
- * the \einterpretation. Thus there are 12 distinct scalar types (more if using
- * ADOL-C): 2 precisions each of real, complex, and conjugate and their
+ * the \e interpretation. Thus there are 12 distinct scalar types (more if
+ * using ADOL-C): 2 precisions each of real, complex, and conjugate and their
  * negators.
  * @verbatim
  *      The Scalar Types
@@ -114,11 +114,16 @@ template <> struct Widest<float,double>             {typedef double      Type;  
 template <> struct Widest<double,float>             {typedef double      Type;  typedef double      Precision;};
 template <> struct Widest<double,double>            {typedef double      Type;  typedef double      Precision;};
 #ifdef SimTK_REAL_IS_ADOUBLE
-    template <> struct Widest<float,adouble>        {typedef adouble     Type;  typedef adouble     Precision;};
-    template <> struct Widest<double,adouble>       {typedef adouble     Type;  typedef adouble     Precision;};
-    template <> struct Widest<adouble,float>        {typedef adouble     Type;  typedef adouble     Precision;};
-    template <> struct Widest<adouble,double>       {typedef adouble     Type;  typedef adouble     Precision;};
-    template <> struct Widest<adouble,adouble>      {typedef adouble     Type;  typedef adouble     Precision;};
+    template <> struct Widest<float,adouble>
+      {typedef adouble     Type;  typedef double     Precision;};
+    template <> struct Widest<double,adouble>
+      {typedef adouble     Type;  typedef double     Precision;};
+    template <> struct Widest<adouble,float>
+      {typedef adouble     Type;  typedef double     Precision;};
+    template <> struct Widest<adouble,double>
+      {typedef adouble     Type;  typedef double     Precision;};
+    template <> struct Widest<adouble,adouble>
+      {typedef adouble     Type;  typedef double     Precision;};
 #endif
 template <class R1, class R2> struct Widest< complex<R1>,complex<R2> > { 
     typedef complex< typename Widest<R1,R2>::Type > Type; 
@@ -150,11 +155,16 @@ template <> struct Narrowest<double,double>            {typedef double Type; typ
 #ifdef SimTK_REAL_IS_ADOUBLE
     /// Be careful that Narrowest(float,adouble) is adouble since it is the
     /// narrowest type than can hold the value and the derivative.
-    template <> struct Narrowest<float,adouble>        {typedef adouble Type; typedef adouble Precision;};
-    template <> struct Narrowest<adouble,float>        {typedef adouble Type; typedef adouble Precision;};
-    template <> struct Narrowest<double,adouble>       {typedef adouble Type; typedef adouble Precision;};
-    template <> struct Narrowest<adouble,double>       {typedef adouble Type; typedef adouble Precision;};
-    template <> struct Narrowest<adouble,adouble>      {typedef adouble Type; typedef adouble Precision;};
+    template <> struct Narrowest<float,adouble>
+      {typedef adouble Type; typedef float Precision;};
+    template <> struct Narrowest<adouble,float>
+      {typedef adouble Type; typedef float Precision;};
+    template <> struct Narrowest<double,adouble>
+      {typedef adouble Type; typedef double Precision;};
+    template <> struct Narrowest<adouble,double>
+      {typedef adouble Type; typedef double Precision;};
+    template <> struct Narrowest<adouble,adouble>
+      {typedef adouble Type; typedef double Precision;};
 #endif
 template <class R1, class R2> struct Narrowest< complex<R1>,complex<R2> > { 
     typedef complex< typename Narrowest<R1,R2>::Type >  Type; 
@@ -189,13 +199,16 @@ public:
 #ifdef SimTK_REAL_IS_ADOUBLE
     template <> class RTraits<adouble> {
     public:
-        static const adouble& getEps()         {static const adouble c=std::numeric_limits<double>::epsilon(); return c;}
-        static const adouble& getSignificant() {static const adouble c=pow(getEps(), 0.875); return c;}
+        static const adouble& getEps() {
+            static const adouble c=std::numeric_limits<double>::epsilon();
+            return c;}
+        static const adouble& getSignificant() {
+            static const adouble c=pow(getEps(), 0.875); return c;}
         /// The default numerical error tolerance is always a double even when
         /// using real as adouble. The rationale was to enable using the
         /// isNumericallyEqual() functions as they are specified. The user
         /// should be careful when using getDefaultTolerance() while taping.
-        static double getDefaultTolerance()    {return getSignificant().value();}
+        static double getDefaultTolerance() {return getSignificant().value();}
     };
 #endif
 
@@ -353,19 +366,22 @@ inline bool isNumericallyEqual(const double& a, const double& b,
     /// real part of the adouble, of type double, is considered. The user
     /// should be careful when using isNumericallyEqual() while taping.
     inline bool isNumericallyEqual(const adouble& a, const double& b,
-                                   double tol = RTraits<adouble>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<adouble>::getDefaultTolerance())
     {   return isNumericallyEqual(a.value(), b, tol); }
     /// Compare a double with an adouble for approximate equality. Only the
     /// real part of the adouble, of type double, is considered. The user
     /// should be careful when using isNumericallyEqual() while taping.
     inline bool isNumericallyEqual(const double& a, const adouble& b,
-                                   double tol = RTraits<adouble>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<adouble>::getDefaultTolerance())
     {   return isNumericallyEqual(a, b.value(), tol); }
     /// Compare two adoubles for approximate equality. Only the
     /// real part of the adoubles, of type double, are considered. The user
     /// should be careful when using isNumericallyEqual() while taping.
     inline bool isNumericallyEqual(const adouble& a, const adouble& b,
-                                   double tol = RTraits<adouble>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<adouble>::getDefaultTolerance())
     {   return isNumericallyEqual(a.value(), b.value(), tol); }
 #endif
 
@@ -378,17 +394,21 @@ inline bool isNumericallyEqual(const double& a, const float& b,
                                double tol = RTraits<float>::getDefaultTolerance())
 {   return isNumericallyEqual(a, (double)b, tol); }
 #ifdef SimTK_REAL_IS_ADOUBLE
-    /// Compare a float and an adouble for approximate equality at float precision.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+    /// Compare a float and an adouble for approximate equality at float
+    /// precision. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     inline bool isNumericallyEqual(const float& a, const adouble& b,
-                                   double tol = RTraits<float>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<float>::getDefaultTolerance())
     {   return isNumericallyEqual(a, b.value(), tol); }
-    /// Compare a float and an adouble for approximate equality at float precision.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+    /// Compare a float and an adouble for approximate equality at float
+    /// precision. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     inline bool isNumericallyEqual(const adouble& a, const float& b,
-                                   double tol = RTraits<float>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<float>::getDefaultTolerance())
     {   return isNumericallyEqual(a.value(), b, tol); }
 #endif
 
@@ -413,13 +433,15 @@ inline bool isNumericallyEqual(int a, const double& b,
     /// part of the adouble, of type double, is considered. The user should be
     /// careful when using isNumericallyEqual() while taping.
     inline bool isNumericallyEqual(const adouble& a, int b,
-                                   double tol = RTraits<adouble>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<adouble>::getDefaultTolerance())
     {   return isNumericallyEqual(a.value(), b, tol); }
     /// %Test an adouble for approximate equality to an integer. Only the real
     /// part of the adouble, of type double, is considered. The user should be
     /// careful when using isNumericallyEqual() while taping.
     inline bool isNumericallyEqual(int a, const adouble& b,
-                                   double tol = RTraits<adouble>::getDefaultTolerance())
+                                   double tol =
+                                   RTraits<adouble>::getDefaultTolerance())
     {   return isNumericallyEqual(a, b.value(), tol); }
 #endif
 
@@ -494,19 +516,24 @@ isNumericallyEqual(int a, const std::complex<P>& b,
                    double tol = RTraits<P>::getDefaultTolerance())
 {   return isNumericallyEqual(b,a,tol); }
 #ifdef SimTK_REAL_IS_ADOUBLE
-    /// %Test whether a complex number is approximately equal to a particular adouble.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+    /// %Test whether a complex number is approximately equal to a particular
+    /// adouble. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     template <class P> inline bool
     isNumericallyEqual(const std::complex<P>& a, const adouble& b,
-                       double tol = RTraits<typename Narrowest<P,adouble>::Precision>::getDefaultTolerance())
-    {   return isNumericallyEqual(a.real(),b,tol) && isNumericallyEqual(a.imag(),0.,tol); }
-    /// %Test whether a complex number is approximately equal to a particular adouble.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+                       double tol = RTraits<typename Narrowest<P,adouble>::
+                       Precision>::getDefaultTolerance())
+    {   return isNumericallyEqual(a.real(),b,tol) &&
+        isNumericallyEqual(a.imag(),0.,tol); }
+    /// %Test whether a complex number is approximately equal to a particular
+    /// adouble. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     template <class P> inline bool
     isNumericallyEqual(const adouble& a, const std::complex<P>& b,
-                       double tol = RTraits<typename Narrowest<P,adouble>::Precision>::getDefaultTolerance())
+                       double tol = RTraits<typename Narrowest<P,adouble>::
+                       Precision>::getDefaultTolerance())
     {   return isNumericallyEqual(b,a,tol);   }
 #endif
 
@@ -541,19 +568,24 @@ isNumericallyEqual(int a, const conjugate<P>& b,
                    double tol = RTraits<P>::getDefaultTolerance())
 {   return isNumericallyEqual(b,a,tol); }
 #ifdef SimTK_REAL_IS_ADOUBLE
-    /// %Test whether a conjugate number is approximately equal to a particular adouble.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+    /// %Test whether a conjugate number is approximately equal to a particular
+    /// adouble. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     template <class P> inline bool
     isNumericallyEqual(const conjugate<P>& a, const adouble& b,
-                       double tol = RTraits<typename Narrowest<P, adouble>::Precision>::getDefaultTolerance())
-    {   return isNumericallyEqual(a.real(),b,tol) && isNumericallyEqual(a.imag(),0.,tol); }
-    /// %Test whether a conjugate number is approximately equal to a particular adouble.
-    /// Only the real part of the adouble, of type double, is considered. The
-    /// user should be careful when using isNumericallyEqual() while taping.
+                       double tol = RTraits<typename Narrowest<P, adouble>::
+                       Precision>::getDefaultTolerance())
+    {   return isNumericallyEqual(a.real(),b,tol) &&
+        isNumericallyEqual(a.imag(),0.,tol); }
+    /// %Test whether a conjugate number is approximately equal to a particular
+    /// adouble. Only the real part of the adouble, of type double, is
+    /// considered. The user should be careful when using isNumericallyEqual()
+    /// while taping.
     template <class P> inline bool
     isNumericallyEqual(const adouble& a, const conjugate<P>& b,
-                       double tol = RTraits<typename Narrowest<P,adouble>::Precision>::getDefaultTolerance())
+                       double tol = RTraits<typename Narrowest<P,adouble>::
+                       Precision>::getDefaultTolerance())
     {   return isNumericallyEqual(b,a,tol); }
 #endif
 
@@ -702,8 +734,10 @@ public:
     static bool isNumericallyEqual(const T& a, int b) {return SimTK::isNumericallyEqual(a,b);}
     static bool isNumericallyEqual(const T& a, int b, double tol) {return SimTK::isNumericallyEqual(a,b,tol);}
     #ifdef SimTK_REAL_IS_ADOUBLE
-        static bool isNumericallyEqual(const T& a, const adouble& b) {return SimTK::isNumericallyEqual(a,b);}
-        static bool isNumericallyEqual(const T& a, const adouble& b, double tol) {return SimTK::isNumericallyEqual(a,b,tol);}
+        static bool isNumericallyEqual(const T& a, const adouble& b)
+        {    return SimTK::isNumericallyEqual(a,b);  }
+        static bool isNumericallyEqual(const T& a, const adouble& b,
+            double tol) {    return SimTK::isNumericallyEqual(a,b,tol);  }
     #endif
 
     // The rest are the same as the real equivalents, with zero imaginary part.              
@@ -916,8 +950,10 @@ public:
     static bool isNumericallyEqual(const T& a, int b) {return SimTK::isNumericallyEqual(a,b);}
     static bool isNumericallyEqual(const T& a, int b, double tol) {return SimTK::isNumericallyEqual(a,b,tol);}
     #ifdef SimTK_REAL_IS_ADOUBLE
-        static bool isNumericallyEqual(const T& a, const adouble& b) {return SimTK::isNumericallyEqual(a,b);}
-        static bool isNumericallyEqual(const T& a, const adouble& b, double tol) {return SimTK::isNumericallyEqual(a,b,tol);}
+        static bool isNumericallyEqual(const T& a, const adouble& b)
+        {    return SimTK::isNumericallyEqual(a,b);  }
+        static bool isNumericallyEqual(const T& a, const adouble& b,
+            double tol) {    return SimTK::isNumericallyEqual(a,b,tol);  }
     #endif
 
     // The rest are the same as the real equivalents, with zero imaginary part.              
@@ -990,8 +1026,8 @@ SimTK_NTRAITS_CONJ_SPEC(double,float);SimTK_NTRAITS_CONJ_SPEC(double,double);
         typedef conjugate<Widest<float,adouble>::Type> W;
         typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
     };
-    template<> template<> struct NTraits< conjugate<double> >::Result<adouble> {
-        typedef conjugate<Widest<double,adouble>::Type> W;
+    template<> template<> struct NTraits< conjugate<double> >::Result<adouble>
+    {   typedef conjugate<Widest<double,adouble>::Type> W;
         typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
     };
 #endif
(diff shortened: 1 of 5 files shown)


antoinefalisse:
@chrisdembia, I added a couple of friend functions of adouble (for fabs, fmax, fmin) in `NTraits` as Travis was complaining. Not sure if it was on purpose that you initially left them out.

sherm1:
+@sherm1 review checkpoint.

---

Reviewed 7 of 8 files at r1.
Review status: 7 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h, line 90 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HVsvnLTS8SwPqqqzw:-L2HVsvnLTS8SwPqqqzx:b-5bp2j4) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h#L90)):*
> ```objc
>         #pragma warning(pop)
>     #endif
> #endif
> ```

Better to do this code just once in `common.h` if possible. There is almost identical code there except it includes only `adouble.h`. What do you think about including `adolc.h` instead in common.h so you don't have to do it here?

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 58 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HWw0qFdjB251Hlpbb:-L2HWw0qFdjB251Hlpbc:b-jhcntv) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L58)):*
> ```cpp
> 
> // Various unit tests verifying that NTraits<adouble> works properly
> void testNTraitsADOLC() {
> ```

BTW great to see this test case -- thanks!

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 129 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HX-dnPfKKn6DGYuTF:-L2HX-dnPfKKn6DGYuTG:bhdbima) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L129)):*
> ```cpp
>     SimTK_TEST(isNumericallyEqual(xad,cjd));
>     SimTK_TEST(isNumericallyEqual(cjf,xad,1e-7));
>     SimTK_TEST(isNumericallyEqual(xad,cjf,1e-7));
> ```

These tolerances should not be necessary. The tolerance is supposed to be figured out from the narrowest of the compared objects. That probably isn't being done correctly in the new isNumericallyEqual() implementations.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 139 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HXbX1j8YwLfg8CcMx:-L2HXbX1j8YwLfg8CcMy:bjb6zhk) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L139)):*
> ```cpp
> 
>     trace_on(0);
>     double c = NTraits<adouble>::value(a);
> ```

Should use
```c++
    SimTK_TEST_MUST_THROW_EXC(NTraits<adouble>::value(a),
            SimTK::Exception::ADOLCTapingNotAllowed
            );
```
That will make sure the exception gets thrown and then absorb it so the overall test doesn't abort.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 150 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HY3-0bVbqmAPpBm45:-L2HY3-0bVbqmAPpBm46:b-oqaazl) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L150)):*
> ```cpp
>             SimTK_SUBTEST(testExceptionTaping);
>         }
>         catch (const std::exception& e) {
> ```

Remove the try/catch here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2HYEzg9Qz0Z1eACJrK:b-f66542)*

sherm1:
Looking very good, @antoinefalisse. Review pass complete, a few comments - PTAL. 
- Note that `Precision` in `Narrowest` and `Widest` should always be `float` or `double` -- when fixed you won't need those explicit 1e-7 tolerances any more.
- Please format new code to 80 characters, even when the surrounding code is wrong. Much easier to review that way.

---

Reviewed 1 of 8 files at r1.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 34 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HnNJqTV21RMctanRG:-L2HnNJqTV21RMctanRH:bg6ne3z) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L34)):*
> ```objc
>  * conjugate<adouble> at the moment). Each of these may be modified by a
>  * negator, which does not change the in-memory \e representation but negates
>  * the \einterpretation. Thus there are 12 distinct scalar types (more if using
> ```

BTW missing space after `\e`

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 121 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HrmhTQR-6rmGswjdF:-L2HrmhTQR-6rmGswjdG:b-pq61jm) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L121)):*
> ```objc
>     template <> struct Widest<adouble,float>        {typedef adouble     Type;  typedef adouble     Precision;};
>     template <> struct Widest<adouble,double>       {typedef adouble     Type;  typedef adouble     Precision;};
>     template <> struct Widest<adouble,adouble>      {typedef adouble     Type;  typedef adouble     Precision;};
> ```

`Precision` should be `float` or `double` -- it is used to calculate tolerances.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 154 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HoFbafyH2a0urCbkF:-L2HoFbafyH2a0urCbkG:b-pfahs1) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L154)):*
> ```objc
>     /// narrowest type than can hold the value and the derivative.
>     template <> struct Narrowest<float,adouble>        {typedef adouble Type; typedef adouble Precision;};
>     template <> struct Narrowest<adouble,float>        {typedef adouble Type; typedef adouble Precision;};
> ```

Precision should be `float` here even though the calculation has to be done in an adouble. The float value has much less precision than the adouble so the result of combining them can't have more precision than that.

This is why you had to put those `1e-7`s in your test cases with `complex<float>` to get them to pass -- that uses Precision to figure out the correct tolerance.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 157 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HrxIAy8YZvNt-PprN:-L2HrxIAy8YZvNt-PprO:b-5ttfw0) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L157)):*
> ```objc
>     template <> struct Narrowest<double,adouble>       {typedef adouble Type; typedef adouble Precision;};
>     template <> struct Narrowest<adouble,double>       {typedef adouble Type; typedef adouble Precision;};
>     template <> struct Narrowest<adouble,adouble>      {typedef adouble Type; typedef adouble Precision;};
> ```

Precision should always be `float` or `double`.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 200 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HqYT6uZSz0gtdsB_v:-L2HqYT6uZSz0gtdsB_w:b2pm61) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L200)):*
> ```objc
>         static double getDefaultTolerance()    {return getSignificant().value();}
>     };
> #endif
> ```

BTW it is harder to review lines longer than 80 characters because they wrap in reviewable. I realize you are being consistent with the other unfortunately-formatted code in this file. It would still be better if new code sticks to the 80 character limit.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 549 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Ht4Hjtu1gK7iu_NVy:-L2Ht4Hjtu1gK7iu_NVz:biphp20) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L549)):*
> ```objc
>     template <class P> inline bool
>     isNumericallyEqual(const conjugate<P>& a, const adouble& b,
>                        double tol = RTraits<typename Narrowest<P, adouble>::Precision>::getDefaultTolerance())
> ```

BTW this line is where it matters what's in `Narrowest::Precision`.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2Htz0G5hWglyJU1w5D:behd447) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*
> ```objc
>     adub fabs(const badouble&);
>     adub fmax(const badouble&, const badouble&);
>     adub fmin(const badouble&, const badouble&);
> ```

BTW why are some of the above `adub` and some `adouble`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2HvAFO-iKiI60XJd-H:b-4yydu5)*

chrisdembia:
> For unit tests, do you want them all in different cpps or I can just add them in TestADOLCCommon.cpp as it is for now?

I think this is fine as-is.

---

Reviewed 7 of 8 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h, line 90 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HVsvnLTS8SwPqqqzw:-L2Hn86e4xrT8dvLz2mb:bmirf02) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h#L90)):*

Better to do this code just once in `common.h` if possible. There is almost identical code there except it includes only `adouble.h`. What do you think about including `adolc.h` instead in common.h so you don't have to do it here?
</blockquote></details>

I thought we could try to avoid including all of ADOL-C (via `adolc.h`) unless necessary, to reduce compile times. Most code only needs the `adouble` type and operators, which are available with `adouble.h`.

However, I think it's fine we just include `adolc.h` in `common.h`.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2ICJibI1BlCEPga0wC:bn1p9qr) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*
https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/adouble.h#L210

>    The class Adub
   ---- Basically used as a temporary result.  The address for an
        adub is usually generated within an operation.  That address
        is "freed" when the adub goes out of scope (at destruction time).
   ---- operates just like a badouble, but it has a destructor defined for it.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 31 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Ho4uPyoWN0W8A43Xy:-L2Ho4uPyoWN0W8A43Xz:benhtpd) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L31)):*
> ```cpp
> using std::cout;
> using std::endl;
> using std::cin;
> ```

`using std::cin;` should not be necessary.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 114 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2I92Psa7LPl-Pbmxge:-L2I92Psa7LPl-Pbmxgf:bbmqmjf) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L114)):*
> ```cpp
>     cjf = SimTK::conjugate<float>(xf,0);
>     SimTK::conjugate<double> cjd;
>     cjd = SimTK::conjugate<double>(xd,0.);
> ```

Is there a reason you didn't construct these in a single line (`std::complex<float> cf(xf, 0.);`)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2ID7iglzMEt-eV6q9x:bg8t8fw)*

chrisdembia:
> I wanted to include free function templates to reduce text as suggested by @chrisdembia but it did not work out at first attempt.

I think I got these functions working here: https://github.com/chrisdembia/simbody/tree/ntraits_free_function_templates

The key was specifying the return type as `NTraits<T>::T` to avoid these template functions from being selected when `T` was not a scalar type.

One downside of providing these free function templates is that client code may have to be updated to avoid ambiguous lookup (between `std::abs` and `SimTK::abs`).

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2IUncmtTK45je0Yj7u:b-285oyr)*

sherm1:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h, line 90 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HVsvnLTS8SwPqqqzw:-L2IeUVoAlQlq2pmwUif:bcrlsgh) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h#L90)):*

I thought we could try to avoid including all of ADOL-C (via `adolc.h`) unless necessary, to reduce compile times. Most code only needs the `adouble` type and operators, which are available with `adouble.h`.

However, I think it's fine we just include `adolc.h` in `common.h`.
</blockquote></details>

I _think_ ExceptionMacros.h gets included almost everywhere that common.h does in Simbody so including adolc.h here is likely no different than including it in common.h. Would be worth checking to make sure I'm seeing that right, but if so then I don't think it is saving any time to attempt just to include adouble.h.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

sherm1:
> One downside of providing these free function templates  ...

Yes, I think that is too dangerous. It is bad enough that ADOLC injects methods into the global namespace :(.

---

Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2IfBXDDw-bPYurp0qv:b-3v121r)*

antoinefalisse:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h, line 90 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HVsvnLTS8SwPqqqzw:-L2Lil_kY1o-pwUvbfGF:b80p80c) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/include/SimTKcommon/internal/ExceptionMacros.h#L90)):*

I _think_ ExceptionMacros.h gets included almost everywhere that common.h does in Simbody so including adolc.h here is likely no different than including it in common.h. Would be worth checking to make sure I'm seeing that right, but if so then I don't think it is saving any time to attempt just to include adouble.h.
</blockquote></details>

I replaced `adouble.h` by `adolc.h` in `common.h`.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 34 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HnNJqTV21RMctanRG:-L2LjVl1h9qDHsjtzS2n:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L34)):*

BTW missing space after `\e`
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 121 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HrmhTQR-6rmGswjdF:-L2Lk4mIv_tsfsYvZk8T:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L121)):*

`Precision` should be `float` or `double` -- it is used to calculate tolerances.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 154 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HoFbafyH2a0urCbkF:-L2Lk72y10kCXpITYzUZ:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L154)):*

Precision should be `float` here even though the calculation has to be done in an adouble. The float value has much less precision than the adouble so the result of combining them can't have more precision than that.

This is why you had to put those `1e-7`s in your test cases with `complex<float>` to get them to pass -- that uses Precision to figure out the correct tolerance.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 157 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HrxIAy8YZvNt-PprN:-L2Lk9jfr_i-FFm6uqwU:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L157)):*

Precision should always be `float` or `double`.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 200 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HqYT6uZSz0gtdsB_v:-L2LsGqv4jl-Mcg9icxY:bny1l10) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L200)):*

BTW it is harder to review lines longer than 80 characters because they wrap in reviewable. I realize you are being consistent with the other unfortunately-formatted code in this file. It would still be better if new code sticks to the 80 character limit.
</blockquote></details>

Okay, I changed all lines longer than 80 characters.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 31 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Ho4uPyoWN0W8A43Xy:-L2LsLMNJevFsqyPdM4W:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L31)):*

`using std::cin;` should not be necessary.
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 114 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2I92Psa7LPl-Pbmxge:-L2Ltw9rofxnAfuHQpj9:b-83q4gn) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L114)):*

Is there a reason you didn't construct these in a single line (`std::complex<float> cf(xf, 0.);`)?
</blockquote></details>

No :)

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 129 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HX-dnPfKKn6DGYuTF:-L2Lu-hE4upSAVoAzz5g:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L129)):*

These tolerances should not be necessary. The tolerance is supposed to be figured out from the narrowest of the compared objects. That probably isn't being done correctly in the new isNumericallyEqual() implementations.
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 139 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HXbX1j8YwLfg8CcMx:-L2LtSlIJonJXDPqVr5G:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L139)):*

Should use
```c++
    SimTK_TEST_MUST_THROW_EXC(NTraits<adouble>::value(a),
            SimTK::Exception::ADOLCTapingNotAllowed
            );
```
That will make sure the exception gets thrown and then absorb it so the overall test doesn't abort.
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 150 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HY3-0bVbqmAPpBm45:-L2LtS-B1_PAieAdzpRd:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L150)):*

Remove the try/catch here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

antoinefalisse:
@sherm1, @chrisdembia, this PR is ready for review again. I have corrected the mistakes in `widest `and `narrowest `and it works fine now.

> Yes, I think that is too dangerous. 

Based on this comment, I have left out the free function templates for now.

chrisdembia:
I only have one requested change; see the line comments.

@sherm1 may have a preference for the formatting of line breaks and white space.

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 154 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HoFbafyH2a0urCbkF:-L2klMRFpEr0U6z4dNkQ:b-x7joh5) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L154)):*

Done.
</blockquote></details>

The comment should now say `Narrowest<float,adouble>::Type` instead of `Narrowest(float,adouble)`.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1405 at r2](https://reviewable.io:443/reviews/simbody/simbody/603#-L2km2M23tZf2pPPObrA:-L2km2M23tZf2pPPObrB:bjl5575) ([raw file](https://github.com/simbody/simbody/blob/28f174b561b1e23ca7f63255e2ac4cf9519ea24a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1405)):*
> ```objc
>         /* Methods to use for approximate comparisons. Perform comparison in
>         /* the wider of the two precisions, using the default tolerance from
>         /* the narrower of the two precisions. */
> ```

This comment's formatting is confusing. The one-line `/* ... */` format was used because that's all that works for a multi-line comment inside a macro (I think), but this is not a macro. So you could remove the two `/*` within the comment.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 114 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2I92Psa7LPl-Pbmxge:-L2kmYXiXoBjUMokOQt-:b-8fx2u7) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L114)):*

No :)
</blockquote></details>

:+1: 

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 129 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HX-dnPfKKn6DGYuTF:-L2kmeASmfKpnEL6R-VD:ba580ca) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/tests/TestADOLCCommon.cpp#L129)):*

Done.
</blockquote></details>

Hooray!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2kmQHRTSGFaXs-uN3R:b-ir36za)*

antoinefalisse:
Thanks @chrisdembia. I made the changes.

---

Review status: 7 of 8 files reviewed at latest revision, 11 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 154 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HoFbafyH2a0urCbkF:-L2l1dkqQjvWroPm7Y-u:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L154)):*

The comment should now say `Narrowest<float,adouble>::Type` instead of `Narrowest(float,adouble)`.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1405 at r2](https://reviewable.io:443/reviews/simbody/simbody/603#-L2km2M23tZf2pPPObrA:-L2l1n35VuDrSGexNPec:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/28f174b561b1e23ca7f63255e2ac4cf9519ea24a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1405)):*

This comment's formatting is confusing. The one-line `/* ... */` format was used because that's all that works for a multi-line comment inside a macro (I think), but this is not a macro. So you could remove the two `/*` within the comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2l2PQCycSTvSMgx3Sa:b-lj8vtn)*

sherm1:
:lgtm: -- just a few new comments.

---

Reviewed 3 of 4 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 200 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2HqYT6uZSz0gtdsB_v:-L2rNzM4QAwzqSZU6IFL:b-ucfknh) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L200)):*

Okay, I changed all lines longer than 80 characters.
</blockquote></details>

Thanks!

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2rQOHInkgfcaYHbTKi:bj7w4cm) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*

https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/adouble.h#L210

>    The class Adub
   ---- Basically used as a temporary result.  The address for an
        adub is usually generated within an operation.  That address
        is "freed" when the adub goes out of scope (at destruction time).
   ---- operates just like a badouble, but it has a destructor defined for it.

</blockquote></details>

I looked through the adolc code and found some of these but it still wasn't obvious why this particular list is here or where it came from. For example, the code has several overloads of `pow()` but only one of those is listed here. Why? I don't have any objection but something this obscure definitely needs to be explained in comments.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 12 at r3](https://reviewable.io:443/reviews/simbody/simbody/603#-L2rRE1oH_EOXioLjbCZ:-L2rRE1oH_EOXioLjbC_:b-8dlagm) ([raw file](https://github.com/simbody/simbody/blob/46ca93af35519dcd629ff63a2681717230394c2c/SimTKcommon/tests/TestADOLCCommon.cpp#L12)):*
> ```cpp
>  * Authors: Antoine Falisse                                                   *
>  * Contributors: Michael Sherman, Chris Dembia                                                              *
>  *                                                                            *
> ```

BTW formatting problem.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2rRXMl7aucAxGuv2If:b-ube3tc)*

antoinefalisse:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2rZ-a7uFkJ_fty-tDM:ba3xec1) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*

I looked through the adolc code and found some of these but it still wasn't obvious why this particular list is here or where it came from. For example, the code has several overloads of `pow()` but only one of those is listed here. Why? I don't have any objection but something this obscure definitely needs to be explained in comments.
</blockquote></details>

I think @chrisdembia might explain that better than me.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

antoinefalisse:
Review status: 3 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 12 at r3](https://reviewable.io:443/reviews/simbody/simbody/603#-L2rRE1oH_EOXioLjbCZ:-L2r_xdpUokjyv9y--kE:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/46ca93af35519dcd629ff63a2681717230394c2c/SimTKcommon/tests/TestADOLCCommon.cpp#L12)):*

BTW formatting problem.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

sherm1:
Reviewed 1 of 2 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

chrisdembia:
Reviewed 1 of 1 files at r3, 1 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2rdczcAamcF42mnZc0:bpp2ib4) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*
I suggest we provide the following comment (replacing the briefer one that's there now around line 1243):

> If we implement NTraits<adouble>::abs(const T& t) as { abs(t); }, then we get an infinite recursion. Therefore, we use { ::abs(t); } so that the compiler can find the friend functions that ADOL-C declares in adouble.h and internal/adubfunc.h. However, this is not sufficient: the standard says, "A name prefixed by the unary scope operator :: ... shall be declared in global namespace scope or shall be a name whose declaration is visible in global scope because of a using-directive." ADOL-C does not declare these functions in global namespace scope (the functions are declared within a class, at the same time they are marked as friends), so we must make them "visible in global scope [with] a using-directive."
https://stackoverflow.com/questions/18449855/c-friend-function-hidden-by-class-function/18451369


@sherm1, nice catch about the multiple overloads of `pow()`. Since the exponent is often not active, I suggest we also add add declarations for the following (declared by ADOL-C [here](https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/adouble.h#L181) and [here](https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/internal/adubfunc.h#L62):

```cpp
friend adouble pow(double, const badouble&);
friend adub pow (const badouble&, double);
```

Then, we could change `NTraits<adouble>::pow()` to:

```cpp
template <typename TBase, typename TExp>
static T pow(const TBase& base, const TExp& exponent) {return ::pow(base, exponent);}
```

Some of the `::pow()` functions return `adouble` and some return `adub`, we might need to use the following instead:

```cpp
template <typename TBase, typename TExp, typename TResult>
static TResult pow(const TBase& base, const TExp& exponent) {return ::pow(base, exponent);}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

chrisdembia:
:lgtm:

---

Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2rwVHNx3BRJV-oBQsl:bnfp4nl)*

antoinefalisse:
@sherm1, @chrisdembia, thanks for the comments. This PR is now ready for review again.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2rvs1Ia5oHdYBkYpNm:b-n6qh0h) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*

I suggest we provide the following comment (replacing the briefer one that's there now around line 1243):

> If we implement NTraits<adouble>::abs(const T& t) as { abs(t); }, then we get an infinite recursion. Therefore, we use { ::abs(t); } so that the compiler can find the friend functions that ADOL-C declares in adouble.h and internal/adubfunc.h. However, this is not sufficient: the standard says, "A name prefixed by the unary scope operator :: ... shall be declared in global namespace scope or shall be a name whose declaration is visible in global scope because of a using-directive." ADOL-C does not declare these functions in global namespace scope (the functions are declared within a class, at the same time they are marked as friends), so we must make them "visible in global scope [with] a using-directive."
https://stackoverflow.com/questions/18449855/c-friend-function-hidden-by-class-function/18451369


@sherm1, nice catch about the multiple overloads of `pow()`. Since the exponent is often not active, I suggest we also add add declarations for the following (declared by ADOL-C [here](https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/adouble.h#L181) and [here](https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/internal/adubfunc.h#L62):

```cpp
friend adouble pow(double, const badouble&);
friend adub pow (const badouble&, double);
```

Then, we could change `NTraits<adouble>::pow()` to:

```cpp
template <typename TBase, typename TExp>
static T pow(const TBase& base, const TExp& exponent) {return ::pow(base, exponent);}
```

Some of the `::pow()` functions return `adouble` and some return `adub`, we might need to use the following instead:

```cpp
template <typename TBase, typename TExp, typename TResult>
static TResult pow(const TBase& base, const TExp& exponent) {return ::pow(base, exponent);}
```
</blockquote></details>

Thanks @chrisdembia for the explanation. FYI the first option you suggested worked but the second option gave me errors (no instance of `pow()` matching the needed definition for `c=pow(getEps(), (double)1.25L);`, ie (const adouble, double). This is a bit confusing since this seems to be the definition provided by the second friend function in your comment...) . In the latest commit, I have thus included the first option.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2rxqekxdIVgcHoyZma:bv3z0oj)*

antoinefalisse:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2v4M9tpDhbqwCGO7GS:b16qoa8) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*

Thanks @chrisdembia for the explanation. FYI the first option you suggested worked but the second option gave me errors (no instance of `pow()` matching the needed definition for `c=pow(getEps(), (double)1.25L);`, ie (const adouble, double). This is a bit confusing since this seems to be the definition provided by the second friend function in your comment...) . In the latest commit, I have thus included the first option.
</blockquote></details>

@chrisdembia, I am wondering if it is complete and correct.

- There are multiple overloads of fmax and fmin ([https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/internal/adubfunc.h#L80](url)). So I guess we should add:


    ```
    adub fmin(double, const badouble&);
    adub fmin(const badouble&, double);

    adub fmax(double, const badouble&);
    adub fmax(const badouble&, double);
    
and change `Ntraits<adouble>::fmax` and `Ntraits<adouble>::fmin` to

```
template <typename Targ1, typename Targ2>
static T max(const Targ1& t, const Targ2& t2)
    {return ::fmax(t,t2);}

template <typename Targ1, typename Targ2>
static T min(const Targ1& t, const Targ2& t2)
    {return ::fmin(t,t2);}

```

- I am wondering whether it works fine. 1) in the new PR ADOLC negator, I created a test that is fine in Visual Sudio but that does not pass one of the Travis jobs (gcc with ADOLC-ON, see [https://travis-ci.org/simbody/simbody/builds/329162176?utm_source=github_status&utm_medium=notification](url))). This seems related to this line ([https://github.com/antoinefalisse/simbody/blob/adolc_negator/SimTKcommon/tests/TestADOLCCommon.cpp#L149](url)). When I adjust the test and replace `pow(x,3)` by `x`, then it is fine (also gcc with ADOLC-ON). 2) IntelliSense in Visual Studio tells me, for the six friend functions I added (2 for pow, 2 for fmax, and 2 for fmin) that "the function definition is not found".  This makes me wonder if this is working as expected...
```


---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

**Comment by chrisdembia** [comment]
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2vDprLMoXD7w7F4gVE:buy8q2h) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*

@chrisdembia, I am wondering if it is complete and correct.

- There are multiple overloads of fmax and fmin ([https://gitlab.com/adol-c/adol-c/blob/master/ADOL-C/include/adolc/internal/adubfunc.h#L80](url)). So I guess we should add:


    ```
    adub fmin(double, const badouble&);
    adub fmin(const badouble&, double);

    adub fmax(double, const badouble&);
    adub fmax(const badouble&, double);
    
and change `Ntraits<adouble>::fmax` and `Ntraits<adouble>::fmin` to

```
template <typename Targ1, typename Targ2>
static T max(const Targ1& t, const Targ2& t2)
    {return ::fmax(t,t2);}

template <typename Targ1, typename Targ2>
static T min(const Targ1& t, const Targ2& t2)
    {return ::fmin(t,t2);}

```

- I am wondering whether it works fine. 1) in the new PR ADOLC negator, I created a test that is fine in Visual Sudio but that does not pass one of the Travis jobs (gcc with ADOLC-ON, see [https://travis-ci.org/simbody/simbody/builds/329162176?utm_source=github_status&utm_medium=notification](url))). This seems related to this line ([https://github.com/antoinefalisse/simbody/blob/adolc_negator/SimTKcommon/tests/TestADOLCCommon.cpp#L149](url)). When I adjust the test and replace `pow(x,3)` by `x`, then it is fine (also gcc with ADOLC-ON). 2) IntelliSense in Visual Studio tells me, for the six friend functions I added (2 for pow, 2 for fmax, and 2 for fmin) that "the function definition is not found".  This makes me wonder if this is working as expected...
</blockquote></details>

I thought your change to use `NTraits<adouble>::pow()` would work, if you use the first of the two options in my previous comment. I think the GCC issue was that there is no conversion from `adub` to `negator<adouble>`; only from `adouble` to `negator<adouble>`.

If you are concerned that the wrong function is being called, maybe you can put a breakpoint in `NTraits<adouble>::pow()` and step into `::pow()` to see which variant is called. I'm not sure this would work on Windows because the ADOL-C we use was probably not built with debug symbols.

I can try this out on my computer later today.

Consider using `T1` and `T2` instead of `Targ1`, etc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*
<!-- Sent from Reviewable.io -->

antoinefalisse:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1230 at r1](https://reviewable.io:443/reviews/simbody/simbody/603#-L2Htz0G5hWglyJU1w5C:-L2vKj77jSBQIU1l7MMj:bil2pum) ([raw file](https://github.com/simbody/simbody/blob/77977abbea1dab92981540b911c89e0742fff1d7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1230)):*
Thanks. I am not sure I fully understand your comment though. If I am right, what has changed when if switched from `y = (negator<adouble>&)pow(x,3);`  to `y = (negator<adouble>&)NTraits<adouble>::pow(x,3);` is that the GCC issue changed from "invalid cast of an rvalue expression of type `adub` to type `simTK::negator<adouble>&`" to invalid cast of an rvalue expression of type `SimTK::Ntraits<adouble>::T` {aka `adouble`} to type `simTK::negator<adouble>&`". This last one is surprising since when using `x` instead of `pow(x,3`) it works and `x` is an `adouble`. There is probably something I do not get.

I have tried debugging but as you guessed it does not step into `::pow()`

BTW do you understand why the second option you suggested did not work (see older comment with reference to `static const T c=pow(getEps(), (double)1.25L);`).

> I can try this out on my computer later today.

Thanks.

> Consider using T1 and T2 instead of Targ1, etc.

Sure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603)*

chrisdembia:
> This last one is surprising since when using x instead of pow(x,3) it works and x is an adouble. 

`x` is an l-value (loosely speaking, because it has a name), while `pow(x,3)` is an r-value (it has no name, and its return value is a temporary object). I think there are limits on how you can cast r-values.

> I can try this out on my computer later today.

I tried this out earlier; you already saw my comments on the negator PR.

antoinefalisse:
Thanks. Both PRs are ready then.

---

Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2wvglJFUuKfqucNqwN:b-dpuyg6)*

chrisdembia:
:lgtm:

---

Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/603#-:-L2wxWpRoeW3Fp3mIFkP:bnfp4nl)*

sherm1:
Is this ready to merge now?

chrisdembia:
Woohoo! 🎆

Issue #545, opened 2016-12-23
Debian installation of Simbody 3.5.3 doesn't include visualizer or examples


sherm1:
On Ubuntu 16.04 I installed Simbody with
```
sudo apt-get install libsimbody-dev simbody-doc
```
The installation worked fine when I used the libraries (see issue #543). However, I couldn't find the `simbody-visualizer` executable or the example programs so I couldn't do anything with it other than link.

Also the documents installed in /usr/share/doc/simbody were still gzipped after installation.

@j-rivero and @chrisdembia, is this expected behavior?

chrisdembia:
What about `sudo apt-get install libsimbody`? Or maybe `sudo apt-get install libsimbody3.5v5`? See [here](https://packages.debian.org/search?keywords=simbody&searchon=names&suite=all&section=all). I suspect this would not fix the gzipped documents though.

sherm1:
Thanks, Chris. I tried those. Apparently `libsimbody` doesn't exist, and when I asked for `libsimbody3.5v5` it said "already installed" so I think it must have come along with `libsimbody-dev`.

sherm1:
> Try installing simbody-doc

Thanks, Steve. I did install that along with libsimbody-dev. It provided the gzipped documenation I mentioned but apparently not the examples.

scpeters:
Sorry, I didn't read carefully enough!

scpeters:
I think this is the debian metadata for what gets installed by the `simbody-doc` package:

* https://anonscm.debian.org/cgit/debian-science/packages/simbody.git/tree/debian/simbody-doc.install

YuriOsokin:
Hello, will this problem be fixed in the near future?

sherm1:
> Hello, will this problem be fixed in the near future?

Likely not unless @j-rivero is working on it. Building from source is the best option at the moment.

moorepants:
It is also possible to install using conda on a Debian system. The conda packages include the visualizer.

chrisdembia:
Thanks @moorepants. @YuriOsokin, see https://github.com/simbody/simbody#windows-mac-and-linux-using-conda

chrisdembia:
I just installed a fresh copy of 18.04 and had no issue building the visualizer with the following commands:

```
   10  sudo apt install git
   12  git clone https://github.com/simbody/simbody
   14  cd simbody/
   15  mkdir build
   17  sudo apt install cmake
   19  cd build/
[...]

PR #610, opened 2018-01-08
Update debian metadata to version 3.6

j-rivero:
Some updates:

 * libsimbody3.5 package changes to libsimbody3.6 (ABI changes requires new library names).
 * -dbg packages are now autogenerated
 * Usual changelog update to new version

Nothing big.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/610)


chrisdembia:
:lgtm:

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[doc/debian/control, line 39 at r1](https://reviewable.io:443/reviews/simbody/simbody/610#-L2N5gjgizYyhkVReA3d:-L2N5gjgizYyhkVReA3e:b5u9x5w) ([raw file](https://github.com/simbody/simbody/blob/972a3c2aa3591de1f474126552989df741c927c8/doc/debian/control#L39)):*
> ```
>          liblapack-dev,
> 	 libsimbody3.6 (= ${binary:Version}),
> 	 ${misc:Depends}
> ```

BTW Should these be spaces or tabs?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/610#-:-L2N6L1plY8AJsZ05-xG:bnfp4nl)*

chrisdembia:
@j-rivero I edited the release date to be 16 January. I hope that's okay.

sherm1:
Per [#545](https://github.com/simbody/simbody/issues/545) the previous Debian release was missing some critical pieces (like the visualizer!). I believe you've already discussed this, but I want to make sure it didn't get lost. Otherwise, :lgtm:

---

Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[doc/debian/control, line 2 at r2](https://reviewable.io:443/reviews/simbody/simbody/610#-L2mCoO_lSGJudwhLAMs:-L2mCoO_lSGJudwhLAMt:b-dz15sw) ([raw file](https://github.com/simbody/simbody/blob/e38bb4e7fdbbe6d8a660675252cfa33a883eea0a/doc/debian/control#L2)):*
> ```
> Source: simbody
> Maintainer: Michael Sherman <msherman@stanford.edu>
> ```

BTW email address should be `sherm1@gmail.com` since I left Stanford.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/610#-:-L2mDScNagjsK3TtFEl4:b-v776mi)*

chrisdembia:
In #606 @j-rivero said

> We probably will need another package that ships the binaries (simbody could be use) since all that we have now are the libraries.

Thanks @j-rivero. The packages already contain the `simbody-visualizer` executable. I'm fine with the binary examples being in a separate `simbody` package, but I think `simbody-visualizer` should come with `libsimbody*`. There was also an issue with the documentation not being unzipped.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/610#-:-L2mI-XbT6pJ5_sE9BWN:bodteql)*

chrisdembia:
*[doc/debian/control, line 2 at r2](https://reviewable.io:443/reviews/simbody/simbody/610#-L2mCoO_lSGJudwhLAMs:-L2mJmdrl5srITEtdAyp:blqpm05) ([raw file](https://github.com/simbody/simbody/blob/e38bb4e7fdbbe6d8a660675252cfa33a883eea0a/doc/debian/control#L2)):*

BTW email address should be `sherm1@gmail.com` since I left Stanford.
</blockquote></details>

I've just updated this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/610)*

j-rivero:
> BTW Should these be spaces or tabs?

It does not matter, the format accept both.

> Per #545 the previous Debian release was missing some critical pieces (like the visualizer!). I believe you've already discussed this, but I want to make sure it didn't get lost. Other

I added the inclusion of simbody package that host the libexec directory in 7ef166688d38ce3b9e882f4a171a128acf88e4a2

> BTW email address should be sherm1@gmail.com since I left Stanford.

Also fixed in 7ef166688d38ce3b9e882f4a171a128acf88e4a2

doc/debian/simbody.install
```diff
@@ -0,0 +1 @@
+usr/libexec/*
```
chrisdembia: @j-rivero I think the `simbody-visualizer` executable should be included in the `libsimbody3.6` package.

j-rivero: umm that is an interesting point. According to the [FHS libexec](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html) is used by binaries designed not to call directly by the user. Is this case? If it is, then I think that you are right. If we want to invoke the binaries directly, we probably need to move them to bin/ and use the simbody package.

chrisdembia:
@j-rivero I'm testing the creation of the debian package locally, and I noticed a typo in the `rules` file: in the flag `-DMAKE_BUILD_TYPE:STRING=RelWithDebInfo`, `MAKE` should be `CMAKE`. Note that it seems Debian inserts its own `-DCMAKE_BUILD_TYPE=None` on the same command.

chrisdembia:
@j-rivero an issue that @sherm1 raised with the debian packages is that some of the documentation files are compressed:

screen shot 2018-02-01 at 9 56 12 pm

in /usr/share/doc/simbody/examples/src:
screen shot 2018-02-01 at 9 59 41 pm



Is it possible for the md, pdf, and cpp files to not be compressed?


The Doxygen documentation works, which is great :)

Does it make sense to address the items above within this PR? We could also merge this PR as-is, create the 3.6 release, and handle these changes later. The debian files in this repo aren't the ones used to create the debian package, anyway.

chrisdembia:
I think this is how we should proceed:

- In this PR, fix the issue with `CMAKE_BUILD_TYPE` being set incorrectly (@j-rivero, I can make this change if you provide me some guidance about how CMAKE_BUILD_TYPE should be handled for debian).
- Merge this PR.
- Make the release.
- Address the issue with compressed documentation/examples later.

@sherm1 let me know what you think of this plan.

sherm1:
> @sherm1 let me know what you think of this plan.

Sounds OK to me, @chrisdembia. I do feel that the last Debian release left out so many pieces that Simbody wasn't actually useful (e.g., no visualizer, no examples, no docs). I would very much like to get the whole thing on Debian this time. Is there any danger of these things falling through the cracks?

chrisdembia:
Thanks @j-rivero . Do you think you'll have time in the next few days to look into why the documentation/examples are compressed? If so, I can wait for making the release. If not, no problem. I'll just go ahead and make the release today or tomorrow.

j-rivero:
sorry for the late response but I'm in the middle of busy weeks of work

> In this PR, fix the issue with CMAKE_BUILD_TYPE being set incorrectly (@j-rivero, I can make this change if you provide me some guidance about how CMAKE_BUILD_TYPE should be handled for debian).

I've fixed it. Inside Debian flags are handle [in a different way](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=730636) but for building the packages manually it is probably better to set the CMAKE_BUILD_TYPE to `RelWithDebInfo`.

> Sounds OK to me, @chrisdembia. I do feel that the last Debian release left out so many pieces that Simbody wasn't actually useful (e.g., no visualizer, no examples, no docs).

I've implemented the build and installation of examples in the libsimbody-dev package (so they are not compressed), pdfs without compression and the visualizer.  You can check the testing packages in the [OSRF buildfarm](https://build.osrfoundation.org/job/simbody-pkg_builder-master-debian_sid-amd64/26/). Please be sure that they contain what you want.

>  I would very much like to get the whole thing on Debian this time. Is there any danger of these things falling through the cracks?

In my opinion the best way would be to check Debian or the status of the next Ubuntu LTS and provide patches that directly fix your problems. The repository used for releasing in debian is here https://salsa.debian.org/science-team/simbody (Ubuntu will sync from it).

Note that the [Debian QA checker is complaining about the path of the visualizer](https://build.osrfoundation.org/job/simbody-pkg_builder-master-debian_sid-amd64/26/consoleFull#-2145829291c767c74a-22ce-41e3-b8e9-c82dba4fba96), it probably needs to include the arch triplet (like  `/usr/lib/x86_64-linux-gnu/libexec/`). We need to fix that (patching current install system or adding a patch in Debian to move the visualizer to there) before we can upload the fixes.

As a side note, keeping the debian/ metadata in this two places is error-prone and ugly. I would be happy if someone can implement a mechanism to keep it in sync.

sherm1:
Thanks, @j-rivero -- I really appreciate your work on this!

chrisdembia:
Thanks @j-rivero I'll try testing this again. How do I push a change to the OSRF buildfarm so I can test a fix to the libexec issue?

chrisdembia:
Based on this discussion: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=798739

It seems likely that CMake only adds the arch triplet if `CMAKE_INSTALL_PREFIX=/usr`. When I will test this later, however, the installation is done one directory up from where I run the command the build the package.

chrisdembia:
I ran the command `dpkg-buildpackage -b -rfakeroot -us -uc` and the documentation still seems to be compressed.

The libraries are appropriately located within an arch triplet folder, but the `simbody-visualizer` is not.

Maybe I didn't update my commit appropriately. I'll try again.

j-rivero:
> Maybe I didn't update my commit appropriately. I'll try again.

I tried it again in my local system and here seems to work. I used file-roller to open the `libsimbody-doc` and `libsimbody` deb files (go for the data.tar.gz) directory and check the files in there. If I have time today I will try to generate a jenkins run so we can be sure about it.

chrisdembia:
Thank you @j-rivero

j-rivero:
> The libraries are appropriately located within an arch triplet folder, but the simbody-visualizer is not.

I synced debian info between this PR and [debian](https://salsa.debian.org/science-team/simbody/tree/master/debian). Using debian data I triggered a [Jenkins build for testing](https://build.osrfoundation.org/view/main/view/debian/job/simbody-pkg_builder-master-debian_sid-amd64/30/). From the packages in Jenkins:

 * In libsimbody3.5 I see the visualizer at `./usr/libexec/simbody/`
 * In symbody-doc, the pdfs are not compressed in  `./usr/share/doc/simbody/*.pdf`

@chrisdembia you were right about that some of the files in examples were not compressed. I think that it is now fixed in 972a3c2aa3591de1f474126552989df741c927c8.

There is only one week left before Ubuntu closes the sync from Debian, I can directly fixes these problems for simbody3.5 but we would need the 3.6 version ASAP since it requires some interaction from Debian developers that takes some time.

chrisdembia:
Thanks @j-rivero . Given the timeline how about we merge this , I'll create the release tomorrow, and then any other fixes to the debian package can be patched downstream in the debian repository?

j-rivero:
> Thanks @j-rivero . Given the timeline how about we merge this , I'll create the release tomorrow, and then any other fixes to the debian package can be patched downstream in the debian repository?

sounds good to me.

chrisdembia:
@sherm1 I'm merging this now; I hope that's okay.

sherm1:
👍

PR #613, opened 2018-01-15
ADOLC negator

antoinefalisse:
+@sherm1, +@chrisdembia. This PR contains the changes in `negator.h` necessary to build Simbody with ADOLC. This PR also contains various tests in `TestADOLCCommon.cpp` verifying that `negator<adouble>` works properly. Note that this PR is based other PRs (ADOLC Ntraits #603 and ADOLC common #600) that are still in review. Changes in other files should thus disappear when those PRs are merged. Please let me know whether I should wait for the other PRs to be merged before going forward.

One remark: as already discussed with @chrisdembia, 
```
negator(const adouble& t) {
            v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));
        }
```
might be problematic but I could not get rid of the `value()` without getting errors. I have tried different options:

- Using the following statement gives me an error (cannot convert from 'const adouble' to 'double')
`negator(const adouble& t) {v = -N((typename NTraits<N>::Precision)t);}`

- Using the following statement gives me an error (cannot convert from 'const adouble' to 'SimTK::conjugate)
`negator(const adouble& t) {v = -N(t);}`

- Using the following statement gives me a warning (conversion from 'double' to 'const std::complex::_Ty)
`negator(const adouble& t) {v = -N(NTraits<adouble>::value(t));}`

- Using the following statement is error and warning free
`negator(const adouble& t) {v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));}`

Here was the conclusion of @chrisdembia:

> Perhaps we can go with the last option until we run into issues with it. I like it because it is safe, as value() will give an exception if taping.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/613)

### DIFF
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1341,8 +1341,16 @@ template <> class CNT<double> : public NTraits<double> { };
         static const T& real(const T& t) { return t; }
         static T&       real(T& t)       { return t; }
         static const T& imag(const T&)   { return getZero(); }
+#if defined(__clang__)
+#pragma clang diagnostic push
+// The function `T& imag(T&)` generates a null-dereference warning.
+#pragma clang diagnostic ignored "-Wnull-dereference"
+#endif
         static T&       imag(T&)
             { assert(false); return *reinterpret_cast<T*>(0); }
+#if defined(__clang__)
+#pragma clang diagnostic pop
+#endif
         static const TNeg& negate(const T& t)
             {return reinterpret_cast<const TNeg&>(t);}
         static       TNeg& negate(T& t) {return reinterpret_cast<TNeg&>(t);}
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1022,14 +1022,12 @@ SimTK_NTRAITS_CONJ_SPEC(double,float);SimTK_NTRAITS_CONJ_SPEC(double,double);
 // no support for conjugate<adouble> at the moment. However, these structs
 // are necessary for compilation.
 #ifdef SimTK_REAL_IS_ADOUBLE
-    template<> template<> struct NTraits< conjugate<float> >::Result<adouble> {
-        typedef conjugate<Widest<float,adouble>::Type> W;
-        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
-    };
+    template<> template<> struct NTraits< conjugate<float> >::Result<adouble>
+    {   typedef conjugate<Widest<float,adouble>::Type> W;
+        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub; };
     template<> template<> struct NTraits< conjugate<double> >::Result<adouble>
     {   typedef conjugate<Widest<double,adouble>::Type> W;
-        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub;
-    };
+        typedef W Mul; typedef W Dvd; typedef W Add; typedef W Sub; };
 #endif
 
 
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -153,8 +153,8 @@ template <> struct Narrowest<float,double>             {typedef float  Type; typ
 template <> struct Narrowest<double,float>             {typedef float  Type; typedef float Precision;};
 template <> struct Narrowest<double,double>            {typedef double Type; typedef double Precision;};
 #ifdef SimTK_REAL_IS_ADOUBLE
-    /// Be careful that Narrowest(float,adouble) is adouble since it is the
-    /// narrowest type than can hold the value and the derivative.
+    /// Be careful that Narrowest(float,adouble)::Type is adouble since it is
+    /// the narrowest type than can hold the value and the derivative.
     template <> struct Narrowest<float,adouble>
       {typedef adouble Type; typedef float Precision;};
     template <> struct Narrowest<adouble,float>
@@ -1401,8 +1401,8 @@ template <> class CNT<double> : public NTraits<double> { };
         static bool isNaN   (const T& t) {return SimTK::isNaN(t);}
         static bool isInf   (const T& t) {return SimTK::isInf(t);}
         /* Methods to use for approximate comparisons. Perform comparison in
-        /* the wider of the two precisions, using the default tolerance from
-        /* the narrower of the two precisions. */
+        the wider of the two precisions, using the default tolerance from
+        the narrower of the two precisions. */
         static double getDefaultTolerance()
         {return RTraits<T>::getDefaultTolerance();}
         static bool isNumericallyEqual(const T& t, const float& f)
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1240,14 +1240,24 @@ template <> class CNT<double> : public NTraits<double> { };
 } // namespace SimTK
 
 #ifdef SimTK_REAL_IS_ADOUBLE
-    // Must declare certain friend functions of adouble in the global namespace
-    // so that they are visible within NTraits<adouble>.
+    // If we implement NTraits::abs(const T& t) as { abs(t); }, then we get an
+    // infinite recursion. Therefore, we use { ::abs(t); } so that the compiler
+    // can find the friend functions that ADOL-C declares in adouble.h and
+    // internal/adubfunc.h. However, this is not sufficient: the standard says,
+    // "A name prefixed by the unary scope operator :: ... shall be declared in
+    // global namespace scope or shall be a name whose declaration is visible
+    // in global scope because of a using-directive." ADOL-C does not declare
+    // these functions in global namespace scope (the functions are declared
+    // within a class, at the same time they are marked as friends), so we must
+    // make them "visible in global scope [with] a using-directive."
     // https://stackoverflow.com/questions/18449855/c-friend-function-hidden-by-class-function/18451369
     adub sqrt(const badouble&);
     adub sin(const badouble&);
     adub cos(const badouble&);
     adub floor(const badouble&);
     adouble pow(const badouble&, const badouble&);
+    adouble pow(double, const badouble&);
+    adub pow(const badouble&, double);
     adub exp(const badouble&);
     adub log(const badouble&);
     adub tan(const badouble&);
@@ -1361,7 +1371,9 @@ template <> class CNT<double> : public NTraits<double> { };
         static T    sin(const T& t)     {return ::sin(t);}
         static T    cos(const T& t)     {return ::cos(t);}
         static T    floor(const T& t)   {return ::floor(t);}
-        static T    pow(const T& t, const T& order) {return ::pow(t, order);}
+        template <typename TBase, typename TExp>
+        static T pow(const TBase& base, const TExp& exponent)
+            {return ::pow(base, exponent);}
         static T    exp(const T& t)     {return ::exp(t);}
         static T    log(const T& t)     {return ::log(t);}
         static T    tan(const T& t)     {return ::tan(t);}
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1271,7 +1271,11 @@ template <> class CNT<double> : public NTraits<double> { };
     adub log10(const badouble&);
     adub fabs(const badouble&);
     adub fmax(const badouble&, const badouble&);
+    adub fmax(double, const badouble&);
+    adub fmax(const badouble&, double);
     adub fmin(const badouble&, const badouble&);
+    adub fmin(double, const badouble&);
+    adub fmin(const badouble&, double);
 
     namespace SimTK {
 
@@ -1384,8 +1388,12 @@ template <> class CNT<double> : public NTraits<double> { };
         static T    sinh(const T& t)    {return ::sinh(t);}
         static T    cosh(const T& t)    {return ::cosh(t);}
         static T    tanh(const T& t)    {return ::tanh(t);}
-        static T    max(const T& t, const T& t2) {return ::fmax(t,t2);}
-        static T    min(const T& t, const T& t2) {return ::fmin(t,t2);}
+        template <typename T1, typename T2>
+        static T max(const T1& t, const T2& t2)
+            {return ::fmax(t,t2);}
+        template <typename T1, typename T2>
+        static T min(const T1& t, const T2& t2)
+            {return ::fmin(t,t2);}
         static T    log10(const T& t)   {return ::log10(t);}
         /* properties of this floating point representation, with memory
         /* addresses */
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1305,7 +1305,7 @@ template <> class CNT<double> : public NTraits<double> { };
         typedef T                ULessScalar;
         typedef T                Number;
         typedef T                StdNumber;
-        typedef T                Precision;
+        typedef double           Precision;
         typedef T                ScalarNormSq;
         template <class P> struct Result {
             typedef typename CNT<P>::template Result<adouble>::Mul Mul;
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h
@@ -1366,6 +1366,9 @@ template <> class CNT<double> : public NTraits<double> { };
             {return reinterpret_cast<const TWithoutNegator&>(t);}
         static       TWithoutNegator& updCastAwayNegatorIfAny(T& t)
             {return reinterpret_cast<TWithoutNegator&>(t);}
+        /** Method to use when we want to cast an adouble. This method calls
+        value() when casting an adouble to another scalar type, which prevents
+        taping */
         template <typename TRet>
         static TRet cast(const T& t,
             typename std::enable_if<std::is_same<TRet, adouble>::value>::type*
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -223,7 +223,7 @@ public:
     negator(int                t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const float&       t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}
-    negator(const adouble&     t) {v = NTraits<adouble>::cast<N>(t);}
+    negator(const adouble&     t) {v = -(NTraits<adouble>::cast<N>(t));}
 
     // Some of these may not compile if instantiated -- you can't cast a complex
     // to a float, for example.
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -69,7 +69,7 @@ template <class N> class negator;   // negator is only defined for numbers.
 /**
  * negator<N>, where N is a number type (real, complex, conjugate), is represented in 
  * memory identically to N, but behaves as though multiplied by -1, though at zero
- * cost. Only negators instantiated with the six (more if using ADOl-C) number
+ * cost. Only negators instantiated with the six (more if using ADOL-C) number
  * types (real, complex, conjugate) are allowed.
  */ 
 template <class NUMBER> 
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -223,7 +223,24 @@ public:
     negator(int                t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const float&       t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}
-    negator(const adouble&     t) {v = -(NTraits<adouble>::cast<N>(t));}
+    #ifdef SimTK_REAL_IS_ADOUBLE
+        // Allow converting an adouble to negator<N>.
+        // If N is adouble, then simply negate the adouble.
+        // (T is actually unused; just needed for the SFINAE).
+        template <typename T,
+            typename std::enable_if<
+                std::is_same<T, adouble>::value && 
+                std::is_same<N, adouble>::value, int>::type = 0>
+        negator(const adouble& t)
+        {   v = -N(t); }
+        // If N is not adouble, we must call value() (this prevents taping).
+        template <typename T,
+            typename std::enable_if<
+                std::is_same<T, adouble>::value &&
+                !std::is_same<N, adouble>::value, int>::type = 0>
+        negator(const adouble& t)
+        {v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));}
+    #endif
 
     // Some of these may not compile if instantiated -- you can't cast a complex
     // to a float, for example.
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -229,7 +229,7 @@ public:
         // (T is actually unused; just needed for the SFINAE).
         template <typename T,
             typename std::enable_if<
-                std::is_same<T, adouble>::value && 
+                std::is_same<T, adouble>::value &&
                 std::is_same<N, adouble>::value, int>::type = 0>
         negator(const adouble& t)
         {   v = -N(t); }
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -226,19 +226,18 @@ public:
     #ifdef SimTK_REAL_IS_ADOUBLE
         // Allow converting an adouble to negator<N>.
         // If N is adouble, then simply negate the adouble.
-        // (T is actually unused; just needed for the SFINAE).
-        template <typename T,
+        template <typename NN,
             typename std::enable_if<
-                std::is_same<T, adouble>::value &&
+                std::is_same<NN, adouble>::value &&
                 std::is_same<N, adouble>::value, int>::type = 0>
-        negator(const adouble& t)
+        negator(const NN& t)
         {   v = -N(t); }
         // If N is not adouble, we must call value() (this prevents taping).
-        template <typename T,
+        template <typename NN,
             typename std::enable_if<
-                std::is_same<T, adouble>::value &&
+                std::is_same<NN, adouble>::value &&
                 !std::is_same<N, adouble>::value, int>::type = 0>
-        negator(const adouble& t)
+        negator(const NN& t)
         {v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));}
     #endif
 
--- a/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
+++ b/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h
@@ -28,9 +28,9 @@
  * This file defines the negator<N> template which is an adaptor for
  * the numeric types N (Real, Complex, conjugate). negator must NOT
  * be instantiated for anything other than these three types (each
- * of which comes in three precisions). negator<N> is guaranteed to
- * have the same memory layout as N, except that the stored values
- * represent the *negative* value of that number.
+ * of which comes in two precisions; more if using ADOL-C). negator<N> is
+ * guaranteed to have the same memory layout as N, except that the stored
+ * values represent the *negative* value of that number.
  *
  * This is part of the SimTK Scalar package, which forms the basis
  * for composite numerical types like vectors and matrices. The negator
@@ -41,7 +41,8 @@
  *
  * The Scalar Types
  * ----------------
- * Here is a complete taxonomy of the scalar types we support.
+ * Here is a complete (not including ADOL-C cases) taxonomy of the scalar
+ * types we support.
  *
  * <scalar>    ::= <number> | negator< <number> >
  * <number>    ::= <standard> | <conjugate>
@@ -68,8 +69,8 @@ template <class N> class negator;   // negator is only defined for numbers.
 /**
  * negator<N>, where N is a number type (real, complex, conjugate), is represented in 
  * memory identically to N, but behaves as though multiplied by -1, though at zero
- * cost. Only negators instantiated with the nine number types (real, complex, 
- * conjugate) are allowed.
+ * cost. Only negators instantiated with the six (more if using ADOl-C) number
+ * types (real, complex, conjugate) are allowed.
  */ 
 template <class NUMBER> 
 class SimTK_SimTKCOMMON_EXPORT negator {
@@ -222,6 +223,11 @@ public:
     negator(int                t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const float&       t) {v = -N((typename NTraits<N>::Precision)t);}
     negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}
+    #ifdef SimTK_REAL_IS_ADOUBLE
+        negator(const adouble& t) {
+            v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));
+        }
+    #endif
 
     // Some of these may not compile if instantiated -- you can't cast a complex
     // to a float, for example.
@@ -271,6 +277,9 @@ template <class N2> friend class negator;
 //@{
 inline bool isNaN(const negator<float>&  x) {return isNaN(-x);}
 inline bool isNaN(const negator<double>& x) {return isNaN(-x);}
+#ifdef SimTK_REAL_IS_ADOUBLE
+    inline bool isNaN(const negator<adouble>& x) {return isNaN(-x);}
+#endif
 template <class P> inline bool
 isNaN(const negator< std::complex<P> >& x) {return isNaN(-x);}
 template <class P> inline bool
@@ -284,6 +293,9 @@ isNaN(const negator< conjugate<P> >&    x) {return isNaN(-x);}
 //@{
 inline bool isFinite(const negator<float>&  x) {return isFinite(-x);}
 inline bool isFinite(const negator<double>& x) {return isFinite(-x);}
+#ifdef SimTK_REAL_IS_ADOUBLE
+    inline bool isFinite(const negator<adouble>& x) {return isFinite(-x);}
+#endif
 template <class P> inline bool
 isFinite(const negator< std::complex<P> >& x) {return isFinite(-x);}
 template <class P> inline bool
@@ -297,6 +309,9 @@ isFinite(const negator< conjugate<P> >&    x) {return isFinite(-x);}
 //@{
 inline bool isInf(const negator<float>&  x) {return isInf(-x);}
 inline bool isInf(const negator<double>& x) {return isInf(-x);}
+#ifdef SimTK_REAL_IS_ADOUBLE
+    inline bool isInf(const negator<adouble>& x) {return isInf(-x);}
+#endif
 template <class P> inline bool
 isInf(const negator< std::complex<P> >& x) {return isInf(-x);}
 template <class P> inline bool
--- a/SimTKcommon/include/SimTKcommon/internal/common.h
+++ b/SimTKcommon/include/SimTKcommon/internal/common.h
@@ -120,7 +120,7 @@ or any other Index type to an argument expecting a certain Index type. **/
         #ifdef _MSC_VER
             // Ignore warnings from ADOL-C headers.
             #pragma warning(push)
-            // 'argument': conversion from 'size_t' to 'locint', possible loss 
+            // 'argument': conversion from 'size_t' to 'locint', possible loss
             // of data.
             #pragma warning(disable: 4267)
         #endif
--- a/SimTKcommon/tests/TestADOLCCommon.cpp
+++ b/SimTKcommon/tests/TestADOLCCommon.cpp
@@ -175,6 +175,11 @@ void testNegator() {
     SimTK_TEST(!isFinite((negator<adouble>&)xInf));
     SimTK_TEST(isInf((negator<adouble>&)xInf));
     SimTK_TEST(!isInf((negator<adouble>&)xad));
+    // ensure consistent behavior between double and adouble
+    double a = 5;
+    adouble ad = 5;
+    SimTK_TEST((negator<double>)a == (negator<adouble>)ad);
+    SimTK_TEST((negator<double>&)a == (negator<adouble>&)ad);
 }
 
 // Various unit tests verifying that cast() works properly
--- a/SimTKcommon/tests/TestADOLCCommon.cpp
+++ b/SimTKcommon/tests/TestADOLCCommon.cpp
@@ -146,7 +146,8 @@ void testNegator() {
     adouble x;
     adouble y;
     x <<= xp[0];
-    y = (negator<adouble>&)NTraits<adouble>::pow(x,3);
+    auto result = NTraits<adouble>::pow(x,3);
+    y = (negator<adouble>&)result;
     double y0;
     y >>= y0;
     trace_off();
--- a/SimTKcommon/tests/TestADOLCCommon.cpp
+++ b/SimTKcommon/tests/TestADOLCCommon.cpp
@@ -142,59 +142,75 @@ void testNegator() {
     // Test evaluation of simple function and its derivative
     double xp[1];
     xp[0] = 2;
-    trace_on(2);
-    adouble x;
-    adouble y;
-    x <<= xp[0];
+    const short int TraceTag = 2;
+    trace_on(TraceTag);
+    adouble x, y; // Declare active variables
+    x <<= xp[0]; // Select independent variable
     auto result = NTraits<adouble>::pow(x,3);
-    y = (negator<adouble>&)result;
+    // Negate the result by reinterpretation rather than computation
+    y = reinterpret_cast<const negator<adouble>&>(result);
     double y0;
-    y >>= y0;
+    y >>= y0; // Select dependent variable
     trace_off();
-    // function evaluation
+    // Function evaluation. function(arg1,arg2,arg3,arg4,arg5) evaluates the
+    // desired function from the tape instead of executing the corresponding
+    // source code. arg1 is the tape identification, arg2 the number of
+    // dependent variables m, arg3 the number of independent variables n, arg4
+    // the independent vector x, and arg4 the dependent vector y = f(x) with
+    // f: R(n) -> R(m). Please see ADOL-C manual for more details.
     double f[1];
-    function(2, 1, 1, xp, f);
+    function(TraceTag, 1, 1, xp, f);
     SimTK_TEST(f[0] == -8.);
-    // derivative evaluation
-    double** J;
-    J = myalloc(1, 1);
-    jacobian(2, 1, 1, xp, J);
-    SimTK_TEST(J[0][0] == -3*NTraits<adouble>::pow(x,2));
-    myfree(J);
+    // Derivative evaluation. gradient(arg1,arg2,arg3,arg4) evaluates the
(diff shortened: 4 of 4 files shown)


sherm1:
@antoinefalisse -- #600 is merged and #603 is very close now.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L2rRoQ_lfqGNigZwLGq:b-r493ie)*

chrisdembia:
@antoinefalisse you can compile code on Windows using Clang, which also gives a similar error; https://blogs.msdn.microsoft.com/vcblog/2017/03/07/use-any-c-compiler-with-visual-studio/.

One way to get rid of the error is to remove the `&`:

```
y = (negator<adouble>)NTraits<adouble>::pow(x,3);
```

I'm not sure if this change is okay though; @sherm1 ?

Another way would be to first store the result of pow in a variable, so that it's an l-value rather than an r-value:

```
auto result = NTraits<adouble>::pow(x,3);
y = (negator<adouble>)result;
```

antoinefalisse:
Thanks. Removing the `&` might be fine to avoid the error but the test then throws the taping error due to the `value()` (I guess it comes from [https://github.com/antoinefalisse/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228](url))

The two-step approach seems fine though. I just kept the `&`in `y = (negator<adouble>&)result;`.

> you can compile code on Windows using Clang

Thanks, I will do that.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L2wbZHrIPEmoFD1YK1Y:bsp3s2g)*

chrisdembia:
> The two-step approach seems fine though. 

Okay good. We'll see if @sherm1 has further advice.

sherm1:
Reviewed 1 of 5 files at r1.
Review status: 1 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3QlNy3BTzV_UrDgpo2:b-at8c6e) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*
> ```objc
>     #ifdef SimTK_REAL_IS_ADOUBLE
>         negator(const adouble& t) {
>             v = -N((typename NTraits<N>::Precision)NTraits<adouble>::value(t));
> ```

This seems wrong since it loses track of the derivatives. When N is also adouble (i.e. NN==N in the comment above), the adouble must be preserved exactly but just negated.

I think you could get away with defining this constructor *only* when N=adouble. Then you could just write `v=-t` (that would take a little enable_if template magic). Alternatively, use an NTraits method to perform a cast from one scalar type to another -- that way you can call value() when going from adouble to anything else, but *not* call value() when going from adouble to adouble. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3RU4tSGo6bVIXKF_4r:b-gx6elh) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*

This seems wrong since it loses track of the derivatives. When N is also adouble (i.e. NN==N in the comment above), the adouble must be preserved exactly but just negated.

I think you could get away with defining this constructor *only* when N=adouble. Then you could just write `v=-t` (that would take a little enable_if template magic). Alternatively, use an NTraits method to perform a cast from one scalar type to another -- that way you can call value() when going from adouble to anything else, but *not* call value() when going from adouble to adouble. 


</blockquote></details>

Good point; I agree. The reason for the latter approach would be if a similar situation is likely to arise elsewhere. 

For the former approach, this might look like the following:

```cpp
#ifdef SimTK_REAL_IS_ADOUBLE
    template <typename std::enable_if<std::is_same<N, adouble>>::type* = nullptr>
    negator(const adouble& t) {
        v = -t;
    }
#endif
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3UKUZlSpV85Me-qEx2:b-xexhdx) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*

@chrisdembia, can you educate me about this line you wrote?

`template <typename std::enable_if<std::is_same<N, adouble>>::type* = nullptr>`

It does not seem to work so I am trying to understand what is wrong but I am not sure I understand what we are trying to do (this `<typename x = nullptr>` is quite unclear to me).

Errors reported:

`'std::is_same<NUMBER,adouble>': illegal type for non-type template parameter`
`syntax error: missing ',' before '*'`
`missing type specifier - int assumed`

</blockquote></details>

The following works:

```
    template <class P>
    negator(const P& t,
            typename std::enable_if<std::is_same<P, N>::value>::type* = nullptr)
    {   v = -t; }
```

But I'm not sure that's what we want.

Perhaps we do want those cast functions in `NTraits`, like `NTraits<T>::cast<T2>()`, etc. In this case, we would use `v = -NTraits<adouble>::cast<N>(t);` and the implementation of `NTraits<adouble>::cast<double>()` would call value()`, while the implementation of `NTraits<adouble>::cast<adouble>()` would not. This is really just repeating what @sherm1 said, I think.

@antoinefalisse I can look at this more later today.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3UQGbS8ey_PbRxs7Ab:b1rnh3h) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*

The following works:

```
    template <class P>
    negator(const P& t,
            typename std::enable_if<std::is_same<P, N>::value>::type* = nullptr)
    {   v = -t; }
```

But I'm not sure that's what we want.

Perhaps we do want those cast functions in `NTraits`, like `NTraits<T>::cast<T2>()`, etc. In this case, we would use `v = -NTraits<adouble>::cast<N>(t);` and the implementation of `NTraits<adouble>::cast<double>()` would call value()`, while the implementation of `NTraits<adouble>::cast<adouble>()` would not. This is really just repeating what @sherm1 said, I think.

@antoinefalisse I can look at this more later today.
</blockquote></details>

@antoinefalisse I pushed two commits; one to just fix a warning in NTraits, and another to implement `NTraits::cast()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3UQNBAHx6ZoF9U21CI:bp60i2o) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*

@antoinefalisse I pushed two commits; one to just fix a warning in NTraits, and another to implement `NTraits::cast()`.
</blockquote></details>

You might need to add tests and documentation for `cast()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 228 at r3](https://reviewable.io:443/reviews/simbody/simbody/613#-L3QlNy3BTzV_UrDgpo1:-L3V1pg0C1oLRlccE58a:b-56a03o) ([raw file](https://github.com/simbody/simbody/blob/416385792d7c97bbd0b1154373e9ba1324c54b51/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L228)):*

You might need to add tests and documentation for `cast()`.
</blockquote></details>

Cool thanks. I will do that.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3V37aTJ4WUVN9QuzhO:bz2rw1z) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*
> ```objc
>             typename std::enable_if<!std::is_same<TRet, adouble>::value>::type*
>                          = nullptr)
>         {return (TRet)NTraits<T>::value(t);}
> ```

@chrisdembia, this gives a warning (at least in debug mode):

`warning C4244: 'argument': conversion from 'double' to 'const std::complex<float>::_Ty', possible loss of data`

negator.h(226): note: see reference to function template instantiation `TRet SimTK::NTraits<adouble>::cast<std::complex<float>>(const SimTK::NTraits<adouble>::T &,void *)`

negator.h(226) is `negator(const adouble&     t) {v = NTraits<adouble>::cast<N>(t);}`

Do we want to push/pop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3VNgcPzaKbMbcylvMj:b-gpu4az) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

@chrisdembia, this gives a warning (at least in debug mode):

`warning C4244: 'argument': conversion from 'double' to 'const std::complex<float>::_Ty', possible loss of data`

negator.h(226): note: see reference to function template instantiation `TRet SimTK::NTraits<adouble>::cast<std::complex<float>>(const SimTK::NTraits<adouble>::T &,void *)`

negator.h(226) is `negator(const adouble&     t) {v = NTraits<adouble>::cast<N>(t);}`

Do we want to push/pop?
</blockquote></details>

You can usually avoid that warning by using the appropriate explicit cast, but we have an explicit cast in this case. Where exactly is the code invoking `negator<std::complex<float>>::negator(const adouble&)`?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3VPIsUSEXLS8UWsbkm:bn57azs) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

You can usually avoid that warning by using the appropriate explicit cast, but we have an explicit cast in this case. Where exactly is the code invoking `negator<std::complex<float>>::negator(const adouble&)`?
</blockquote></details>

I think it comes from [scalar](https://github.com/antoinefalisse/simbody/blob/master/SimTKcommon/Scalar/src/Scalar.cpp#L91)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3_5JBgqS6uU08Iy1GG:bto7v48) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

I think it comes from [scalar](https://github.com/antoinefalisse/simbody/blob/master/SimTKcommon/Scalar/src/Scalar.cpp#L91)
</blockquote></details>

Ah. Well `negator<std::complex<float>>::negator(const double&)` (argument as double, not adouble) doesn't generate a warning, so maybe there is a way around it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h, line 226 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3_jV0YKaGNz6_t62YX:-L3_jV0YKaGNz6_t62YY:b7b2hxp) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/negator.h#L226)):*
> ```objc
>     negator(const float&       t) {v = -N((typename NTraits<N>::Precision)t);}
>     negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}
>     negator(const adouble&     t) {v = NTraits<adouble>::cast<N>(t);}
> ```

@chrisdembia, btw isn't there a "-" missing here? `negator(const adouble&     t) {v = -NTraits<adouble>::cast<N>(t);}`

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3_jsyZHgVzVhjLd7An:brhmlbs) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

Ah. Well `negator<std::complex<float>>::negator(const double&)` (argument as double, not adouble) doesn't generate a warning, so maybe there is a way around it.
</blockquote></details>

I will try to find a way around.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
@antoinefalisse I pushed a small change to avoid warnings in NTraits.h.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3_n3km1f9Ae7rFDgZ4:bitvwcw) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

I will try to find a way around.
</blockquote></details>

I can look into this also.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L3_rv77Hs7e8FbLbISD:b-fw5vsl)*

antoinefalisse:
@chrisdembia, thanks but I don't see any new commits.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L3a4x6JT8OfwJJD4yYP:b-9xpqpa)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3a9wBE-Z8sA_-dNOqZ:b-k73z76) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

I can look into this also.
</blockquote></details>

thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3aOQoYlfCfmB2L-ygm:b-4fqxu7) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

thanks
</blockquote></details>

I pushed a fix to the warning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3cQLHzn-n305EDCEDh:b-w4batn) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

I pushed a fix to the warning.
</blockquote></details>

That's great, thanks. I see that `cast()` is no longer used. However, I think it might be useful for other cases later on. I suggest we keep it in that PR and I add comments and test cases to make sure it works properly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3iLc0aEn4BMG8lTYSu:b-op3h3v) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

That's great, thanks. I see that `cast()` is no longer used. However, I think it might be useful for other cases later on. I suggest we keep it in that PR and I add comments and test cases to make sure it works properly.
</blockquote></details>

@chrisdembia, I still wonder if it works as expected. It is not possible to do:

```
auto result = NTraits<adouble>::pow(x,3);
y = (negator<adouble>)result;
```

whereas 
```
auto result = NTraits<double>::pow(x,3);
y = (negator<double>)result;
```

is fine. The error message reports that there is no suitable user-defined conversion from `SimTK::adouble` to`SimTK::negator<adouble>`.

When debugging, the second implementation, ie with `double`, calls:

`negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}`

I therefore assume the` adouble` "counterpart" is not working as expected.

BTW when using `&`, both approaches call `operator N() const { return N(-v); }`, so that seems ok.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3iVmpE38wfKR75jGQJ:b-mzhfgc) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

@chrisdembia, I still wonder if it works as expected. It is not possible to do:

```
auto result = NTraits<adouble>::pow(x,3);
y = (negator<adouble>)result;
```

whereas 
```
auto result = NTraits<double>::pow(x,3);
y = (negator<double>)result;
```

is fine. The error message reports that there is no suitable user-defined conversion from `SimTK::adouble` to`SimTK::negator<adouble>`.

When debugging, the second implementation, ie with `double`, calls:

`negator(const double&      t) {v = -N((typename NTraits<N>::Precision)t);}`

I therefore assume the` adouble` "counterpart" is not working as expected.

BTW when using `&`, both approaches call `operator N() const { return N(-v); }`, so that seems ok.
</blockquote></details>

Okay I'll spend some time looking into this. That information is very helpful.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3iefHFASHRiHkLcY8E:b-183q98) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

Okay I'll spend some time looking into this. That information is very helpful.
</blockquote></details>

Okay I pushed a fix for that. The issue was that the template parameter T wasn't being used in the signature, so the compiler had no way of inferring what T should be if not provided explicitly in the place where we wanted to use this constructor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3iq_fQ1v6p18iTrgNN:b-g7uqe1) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*
Thanks, it seems fine now. For my understanding, is that correct that `y = (negator<adouble>&)result;` and `y = (negator<adouble>)result` will result in opposite `y`. The former actually negating result (so `y = -result`) whereas the latter only changing the type. I guess that is what is described in the following comment (took me a while to integrate that...):

>  // These are implicit conversions from numeric type NN to negator<N>. The value must be unchanged, so we must negate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3jkE9K5YYGvmaFZa44:b-lecht5) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*

Thanks, it seems fine now. For my understanding, is that correct that `y = (negator<adouble>&)result;` and `y = (negator<adouble>)result` will result in opposite `y`. The former actually negating result (so `y = -result`) whereas the latter only changing the type. I guess that is what is described in the following comment (took me a while to integrate that...):

>  // These are implicit conversions from numeric type NN to negator<N>. The value must be unchanged, so we must negate.
</blockquote></details>

Here is my understanding: `(negator<adouble>&)x` treats the memory at x as if it is of type `negator<adouble>`, whereas `(negator<adouble>)x` constructs a `negator<adouble>`, which if constructed with `5` will then represent `5`.  I think the idea is that `negator<adouble>` is supposed to be used with the former type of cast, because it costs nothing to reinterpret the memory and the whole point of the class is zero-cost negation.

However, it is still not clear to me why `(negator<adouble>)5` represents `5` and not `-5`. I would just ensure that behavior is consistent between adouble and double.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

antoinefalisse:
@chrisdembia, @sherm1, this PR is ready for review again.

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1370 at r6](https://reviewable.io:443/reviews/simbody/simbody/613#-L3V37aTJ4WUVN9QuzhN:-L3jqs5nhvYHX3U_uwC1:bj3ch7b) ([raw file](https://github.com/simbody/simbody/blob/8cc80bf6427571d47ca9011005611624ad61eac7/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1370)):*
Thanks. 
>  I would just ensure that behavior is consistent between adouble and double.

It is consistent, at least based on what I included in the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L3jr7-A6LeEg5TZ2YZ1:bg1p8zo)*

antoinefalisse:
@chrisdembia, one of the jobs of Travis CI failed. I don't think it is an error in the code since it returned: "The job exceeded the maximum time limit for jobs, and has been terminated.". Could you start it again? BTW do I have the rights to do that myself? Thanks

chrisdembia:
I restarted the job; I don't think you are able to do it.

antoinefalisse:
Thanks. All is good now.

sherm1:
Review pass done -- some questions, comments, and one bug. +@sherm1 

---

Reviewed 1 of 3 files at r10, 1 of 1 files at r11, 1 of 1 files at r12.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1308 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zu8AWrLr5DQhNVzmF:-L3zu8AWrLr5DQhNVzmG:b-mipnbt) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1308)):*
> ```objc
>         typedef T                Number;
>         typedef T                StdNumber;
>         typedef T                Precision;
> ```

Precision should be `double`, not `adouble`

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 145 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4--GEzvkeubfseQhsq:-L4--GEzvkeubfseQhsr:b-40ok34) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L145)):*
> ```cpp
>     double xp[1];
>     xp[0] = 2;
>     trace_on(2);
> ```

BTW rather than repeated use of "2" (easy to confuse with other 2's like the one right above it!), use a constant like
```c++
const short int TraceTag=2;
```
(or an enum).

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 150 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zwCRzNkXi6vtRYlWz:-L3zwCRzNkXi6vtRYlX-:b-ueip1f) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L150)):*
> ```cpp
>     x <<= xp[0];
>     auto result = NTraits<adouble>::pow(x,3);
>     y = (negator<adouble>&)result;
> ```

BTW should not use C-style cast here -- prefer C++ `reinterpret_cast`, `static_cast`, `const_cast`, or `dynamic_cast`. Also anything tricky like this needs a comment. I _think_ this should be: 
```c++
// Negate the result by reinterpretation rather than computation.
y = reinterpret_cast<const negator<adouble>&>(result);
```
(and below).

I know it's uglier, but it makes it clear when you are doing a dangerous type reinterpretation from a safe type cast (you do both below).

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 152 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4--l_0jIrL1uxPqyjU:-L4--l_0jIrL1uxPqyjV:b-s9okd5) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L152)):*
> ```cpp
>     y = (negator<adouble>&)result;
>     double y0;
>     y >>= y0;
> ```

BTW y0 appears to be unused? If this is intentional please explain.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 156 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zzD9V7Ek7Av6aXN3q:-L3zzD9V7Ek7Av6aXN3r:b-co4uun) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L156)):*
> ```cpp
>     // function evaluation
>     double f[1];
>     function(2, 1, 1, xp, f);
> ```

BTW would be helpful for naive readers to say what these arguments are. And the first one should have a name, not just a raw tag number.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 157 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-00-kqRL6lXK1D0iG:-L4-00-kqRL6lXK1D0iH:b-tn0ubk) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L157)):*
> ```cpp
>     double f[1];
>     function(2, 1, 1, xp, f);
>     SimTK_TEST(f[0] == -8.);
> ```

BTW cool -- this is a great test!

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 159 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-05JL7sSJDB5m05Vy:-L4-05JL7sSJDB5m05Vz:b3yo7x) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L159)):*
> ```cpp
>     SimTK_TEST(f[0] == -8.);
>     // derivative evaluation
>     double** J;
> ```

BTW this was confusing to me. I guess J has to be a column of row pointers and that myalloc() takes care of that? Took me a while to figure that out -- a comment would have helped.

Also, good RAII practice says you should allocate & initialize simultaneously:
```c++
double** J = myalloc(1,1);
```

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 177 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-43fUmlfRC7rLNkEI:-L4-43fUmlfRC7rLNkEJ:bf68w80) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L177)):*
> ```cpp
>     SimTK_TEST(!isFinite((negator<adouble>&)xInf));
>     SimTK_TEST(isInf((negator<adouble>&)xInf));
>     SimTK_TEST(!isInf((negator<adouble>&)xad));
> ```

BTW rather than repeat all these ugly casts consider making negated variables:
```c++
auto& nxNaN = reinterpret_cast<const negator<adouble>&>(xNaN);
// or
auto& nxNaN = NTraits<adouble>::negate(xNaN);
```

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 181 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zy6Q8b5CYPviNRIG3:-L3zy6Q8b5CYPviNRIG4:b91jo2b) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L181)):*
> ```cpp
>     double a = 5;
>     adouble ad = 5;
>     SimTK_TEST((negator<double>)a == (negator<adouble>)ad);
> ```

BTW should comment what you expect to happen here (the values should be unchanged by conversion to the negated type, negated by reinterpreting as the negated type). Consider:
```c++
SimTK_TEST(static_cast<negator<double>>(a) == a);
SimTK_TEST(static_cast<negator<adouble>>(ad) == ad);
SimTK_TEST(reinterpret_cast<negator<double>&>(a) == -a);
SimTK_TEST(reinterpret_cast<negator<adouble>&>(ad) == -ad);
```

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 187 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zvi3kqVcxutxUDxjG:-L3zvi3kqVcxutxUDxjH:b-70i8so) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L187)):*
> ```cpp
> // Various unit tests verifying that cast() works properly
> void testCast() {
>     // cast an adouble in a double
> ```

BTW  "in" -> "to" (and below)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L4-6Op43k5B_UOubkSI:b-nlq65v)*

antoinefalisse:
Thanks for the review! I made the changes so it should be ready for a new review pass. Thanks in advance!

---

Review status: 1 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1308 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zu8AWrLr5DQhNVzmF:-L49HKBHhplK2m_JAu1P:bkuayo7) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1308)):*

Precision should be `double`, not `adouble`
</blockquote></details>

Can I just replace T by `double`?

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 145 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4--GEzvkeubfseQhsq:-L443c5musHwfFFOD1zc:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L145)):*

BTW rather than repeated use of "2" (easy to confuse with other 2's like the one right above it!), use a constant like
```c++
const short int TraceTag=2;
```
(or an enum).
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 150 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zwCRzNkXi6vtRYlWz:-L49ILCOv0PjB-kRoS6u:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L150)):*

BTW should not use C-style cast here -- prefer C++ `reinterpret_cast`, `static_cast`, `const_cast`, or `dynamic_cast`. Also anything tricky like this needs a comment. I _think_ this should be: 
```c++
// Negate the result by reinterpretation rather than computation.
y = reinterpret_cast<const negator<adouble>&>(result);
```
(and below).

I know it's uglier, but it makes it clear when you are doing a dangerous type reinterpretation from a safe type cast (you do both below).
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 152 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4--l_0jIrL1uxPqyjU:-L49IRrSlqzAtWOqXBQq:b-dfsuf5) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L152)):*

One must use the `>>=` to tell ADOL-C that `y` is a dependent variable. Otherwise, ADOL-C would think there are no dependent variables. I agree that adding a comment would be a good idea.
</blockquote></details>

Done. I used the same terminology as in the ADOL-C manual.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 156 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zzD9V7Ek7Av6aXN3q:-L49J6niF3x_doyv5WYm:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L156)):*

BTW would be helpful for naive readers to say what these arguments are. And the first one should have a name, not just a raw tag number.
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 159 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-05JL7sSJDB5m05Vy:-L49JFVw1UoYM601oOyT:b6c6mwn) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L159)):*

It would be good to clean up this memory, perhaps using `myfree` provided by ADOL-C.
</blockquote></details>

Sorry for the confusion. I slightly adjusted the test. The function has only one independent variable so there is actually no need to use `jacobian()` since`gradient()` can be used. Also, no need to use this column of row pointers anymore. BTW, I think I was already using `myFree` but no need for it anymore.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 177 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-43fUmlfRC7rLNkEI:-L49Kkn_d9y07RI7yn-W:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L177)):*

BTW rather than repeat all these ugly casts consider making negated variables:
```c++
auto& nxNaN = reinterpret_cast<const negator<adouble>&>(xNaN);
// or
auto& nxNaN = NTraits<adouble>::negate(xNaN);
```
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 181 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zy6Q8b5CYPviNRIG3:-L49KnCJ8z-8wQxXjJHt:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L181)):*

BTW should comment what you expect to happen here (the values should be unchanged by conversion to the negated type, negated by reinterpreting as the negated type). Consider:
```c++
SimTK_TEST(static_cast<negator<double>>(a) == a);
SimTK_TEST(static_cast<negator<adouble>>(ad) == ad);
SimTK_TEST(reinterpret_cast<negator<double>&>(a) == -a);
SimTK_TEST(reinterpret_cast<negator<adouble>&>(ad) == -ad);
```
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 187 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zvi3kqVcxutxUDxjG:-L49KoDB-JruBdarMb_8:boyexbg) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L187)):*

BTW  "in" -> "to" (and below)
</blockquote></details>

Done. Thanks.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 201 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3y3sgbdTd-Yhiz6MhI:-L49LS_6KigWAkoLynQj:b-zcjvte) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L201)):*

Great test.
</blockquote></details>

Thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L49LUfCScwb5nRpPqVB:bt9yozt)*

sherm1:
Great -- thanks, @antoinefalisse. :lgtm:

---

Reviewed 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h, line 1308 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L3zu8AWrLr5DQhNVzmF:-L49SngRMnIh6hrycyy9:bb77ufg) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L1308)):*

Can I just replace T by `double`?
</blockquote></details>

Yes.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 168 at r13](https://reviewable.io:443/reviews/simbody/simbody/613#-L49TZyJ2_skb59n4A56:-L49TZyJ2_skb59n4A57:bncy3ts) ([raw file](https://github.com/simbody/simbody/blob/b751ef59fc05a2dbb72def24d497a232e80faf19/SimTKcommon/tests/TestADOLCCommon.cpp#L168)):*
> ```cpp
>     // the number of independent variables n (the number of dependent variables
>     // m = 1), arg3 the independent vector x, and arg4 the resulting gradient
>     // of f(x). Please see ADOL-C manual for more details.
> ```

BTW the above comments are very helpful -- thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L49Tpz4_7BimzpaKlEr:b1k5uvi)*

chrisdembia:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 159 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-05JL7sSJDB5m05Vy:-L49Z1Xpmha1Jg6-PuTi:bgdfw6u) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L159)):*
> BTW, I think I was already using myFree but no need for it anymore.

Oops sorry about that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L49ZHkHCHXZt3TZyHll:bnfp4nl)*

chrisdembia:
Reviewed 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613)*

sherm1:
Ready to merge once CI is good.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 159 at r12](https://reviewable.io:443/reviews/simbody/simbody/613#-L4-05JL7sSJDB5m05Vy:-L49fK1c7Xp9uZSzjzFx:bm5tesg) ([raw file](https://github.com/simbody/simbody/blob/5f4040912a98aac32343e6fc4ab18fb03d098dc5/SimTKcommon/tests/TestADOLCCommon.cpp#L159)):*

> BTW, I think I was already using myFree but no need for it anymore.

Oops sorry about that.
</blockquote></details>

@chrisdembia your comment here is blocking, I think unintentionally?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/613#-:-L49fR1xOYFO3fcTlz3k:bv848ip)*

chrisdembia:
> @chrisdembia your comment here is blocking, I think unintentionally?

Oops; I think I successfully unblocked.

The Debug build timed out; I restarted it.

chrisdembia:
CI passed; merging.

antoinefalisse:
Great!

PR #614, opened 2018-01-21
ADOLC Mat Vec

antoinefalisse:
@sherm1, @chrisdembia, this PR contains the changes in `Mat.h `and `Vec.h` necessary to build Simbody with ADOL-C. This PR also contains various tests in `TestADOLCCommon.cpp` verifying that it works properly. EDIT: parent branches have been merged. This PR is thus the next one in the ADOL-C queue.

One question about the tests, should the following operations be supported:

- `a/v` where a is an `adouble `and v a `Vec`
- `a-v` where a is an `adouble `and v a `Vec`
- `a-m` where a is an `adouble `and m a `Mat`

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/614)

PR #618, opened 2018-02-04
ADOLC scalar

antoinefalisse:
@sherm1,@chrisdembia, this PR contains the changes in `scalar.h` necessary to build Simbody with ADOL-C as well as unit tests in `TestADOLCCommon.cpp`. This PR is based on PR #614, currently in review.

A couple of remarks:

- `signBit()` and` sign()` use `value()` and therefore cannot be used when taping. We started discussing that [here](https://reviewable.io/reviews/simbody/simbody/590#-Ky2T6z1rmro3puMiCb9) but it was not clear to me which approach was best in the end so I opted for that one.

- the two `clampInPlace()` functions involving `negator<adouble>` also use `value()`. For some unclear reasons, the tests did not pass on Windows when not using `value()` although they passed on Mac (with clang).

- I used templates for `stepUp()`, `stepDown`, etc. to avoid code duplication. In the original code, there were only implementations for `double `and `float `so perhaps we want to delete the template specializations for the other cases? EDIT: due to the function templates, I had to bring minor changes in `function.h` and `testScalar.cpp`.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/618)


chrisdembia:
The tests seem very thorough! It will be good to get @sherm1's feedback on those points you raise.  +@sherm1

> In the original code, there were only implementations for doubleand floatso perhaps we want to delete the template specializations for the other cases?

An alternative is to use SFINAE (enable_if) so that the function templates only exist for float, double, and adouble. I got this to work locally:

```cpp
  /** This alias allows us to enable certain functions for only floating
  point types (float, double, and maybe ADOL-C's adouble). */
  // https://stackoverflow.com/questions/31596565/template-function-that-matches-only-certain-types
  template <typename T>
  using enable_if_floating_point = typename std::enable_if<
             std::is_same<T, float>::value
          || std::is_same<T, double>::value
  #ifdef SimTK_REAL_IS_ADOUBLE
          || std::is_same<T, adouble>::value
  #endif
          >::type;
 
  /** Interpolate smoothly from 0 up to 1 as the input argument goes from 0 to 1,
  with first and second derivatives zero at either end of the interval.

  @param[in]  x       The control parameter, in range [0,1].
  @return The smoothed output value, in range [0,1] but with first and second
  derivatives smoothly approaching zero as x approaches either end of its
  interval.

  See the documentation for stepAny() for a discussion about how to shift and
  scale this function to produce arbitrary steps.

  This function is overloaded for all the floating point precisions.
  Cost is 7 flops.
  @see stepDown(), stepAny() **/
  template <typename T, typename = enable_if_floating_point<T>>
  T stepUp(T x)
  {   assert(0 <= x && x <= 1);
      return x*x*x*(10+x*(6*x-15)); }  //10x^3-15x^4+6x^5
```

---

Reviewed 3 of 4 files at r1.
Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 339 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YEQbWEQbNfTudUUs0:-L4YEQbWEQbNfTudUUs1:b-e3k53r) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L339)):*
> ```objc
> #ifdef SimTK_REAL_IS_ADOUBLE
>     inline int sign(const adouble&          x)
>         {return -sign(NTraits<adouble>::value(-x)); }
> ```

I think the `-sign(-x)` type of implementation is for the `negator<T>` parameter. You could use `sign(NTraits<adouble>::value())` here.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 842 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGUf6LGP_BDyPLUnV:-L4YGUf6LGP_BDyPLUnW:b-t4s9ss) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L842)):*
> ```objc
>     ordinary adouble. **/
>     inline adouble clamp(double low, negator<adouble> v, double high)
>     {   return clamp(low,(adouble)v,high); }
> ```

Might want to use `NTraits<adouble>::max/min` to allow taping with this variant.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4YGchsewMtCcGY8ekj:b-9ki6h3) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*
> ```objc
> 
> This function is overloaded for all the floating point precisions.
> Cost is 7 flops.
> ```

If indeed we move to these function templates, this comment should be updated.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 405 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4Z66jSNnEGjIh0UXip:-L4Z66jSNnEGjIh0UXiq:buhtsg4) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/tests/TestADOLCCommon.cpp#L405)):*
> ```cpp
>     {   double h = 4;
>         adouble hd = 4;
>         SimTK_TEST(clampInPlace(b,h,2)==clampInPlace(b,hd,2) && hd==2); }
> ```

Just curious: is there a reason that you use an explicit `-2` as the lower bound in the scope above and the variable `b` in this scope? It's not clear to me what is different about these two tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L4YGpPQiA48HvLwIYTu:b-oppq5g)*

chrisdembia:
Reviewed 1 of 4 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Great, it seems to work fine.

---

Review status: 2 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 339 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YEQbWEQbNfTudUUs0:-L4cvvx3L_t6opQsVo74:boyexbg) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L339)):*

I think the `-sign(-x)` type of implementation is for the `negator<T>` parameter. You could use `sign(NTraits<adouble>::value())` here.
</blockquote></details>

Done. Thanks.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 842 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGUf6LGP_BDyPLUnV:-L4cymzBh2EEcziAf07X:b-qu5fxu) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L842)):*

Might want to use `NTraits<adouble>::max/min` to allow taping with this variant.
</blockquote></details>

There is a cast to `adouble` here so I think it will call the `clampInPlace()` implementation with the `max/min`, ie not the `negator<adouble>` implementation using `value()`. I went step by step in Debug and it seems to confirm that.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4cw9BCZfwylrgZlvh2:b-pswcon) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

If indeed we move to these function templates, this comment should be updated.
</blockquote></details>

Ok, I do not really know how that works the "flops". Further guidance on that might be useful (if this still makes sense to inform about the number of flops with the function templates).

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 405 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4Z66jSNnEGjIh0UXip:-L4cwuus_3cXCHD4-AmL:b-edrafd) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/tests/TestADOLCCommon.cpp#L405)):*

Just curious: is there a reason that you use an explicit `-2` as the lower bound in the scope above and the variable `b` in this scope? It's not clear to me what is different about these two tests.
</blockquote></details>

There are cases with `int`, I just wanted to make sure they were fine too. Perhaps better to declare the variables beforehand btw (done in latest commit).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L4cvJDJCyd572z9MfiS:b7mjnpc)*

chrisdembia:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4dOiVFFWnZc668qYsf:b70jwrk) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

Ok, I do not really know how that works the "flops". Further guidance on that might be useful (if this still makes sense to inform about the number of flops with the function templates).
</blockquote></details>

I think the flop count is fine, but "The function is overloaded for all the floating point precisions" could be "This function template works for float, double, and, if using ADOL-C."

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4hQufCvCI-9tw2Yamr:bi0fuh7) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

Oh ok, sorry I thought you meant something with the flops.

BTW now that we have function templates, not sure this is still ok:

```
template <typename T, typename = enable_if_floating_point<T>>
T stepDown(T x) {return 1.0 -stepUp(x);}
```

Shouldn't it be something like this (not sure about the second template argument when calling `stepUp<>`).

```
template <typename T, typename = enable_if_floating_point<T>>
T stepDown(T x) {return 1.0 -stepUp<T,enable_if_floating_point<T>>(x);}
```

Travis crashed cause it could not find `stepAny()` so this might be related.

</blockquote></details>

Ps: actually it is not really related to the issues with `stepAny()`.
I think I forgot to specify the types in the function templates when using `clampInPlace()`, eg I think it should be `clampInPlace((T)0.0,xadj,(T)1.0);`.
We might also need to adjust the original `TestScalar.cpp`:

From

```
SimTK_TEST(stepAny(-1,2,0,1,0.) == -1);
SimTK_TEST(stepAny(-1,2,0,1,.5) == 0);
SimTK_TEST(stepAny(-1,2,0,1,1.) == 1);
SimTK_TEST(stepAny(-1,2,0,1,0.f) == -1);
SimTK_TEST(stepAny(-1,2,0,1,.5f) == 0);
SimTK_TEST(stepAny(-1,2,0,1,1.f) == 1);
```


```
SimTK_TEST(stepAny(-1.,2.,0.,1.,0.) == -1);
SimTK_TEST(stepAny(-1.,2.,0.,1.,.5) == 0);
SimTK_TEST(stepAny(-1.,2.,0.,1.,1.) == 1);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,0.f) == -1);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,.5f) == 0);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,1.f) == 1);
```

I am still interested to know if we need to specify the function template arguments though (see previous comment) :)

Thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

chrisdembia:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4hUu5xvMD3sj2s-oAk:beidcr5) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

Ps: actually it is not really related to the issues with `stepAny()`.
I think I forgot to specify the types in the function templates when using `clampInPlace()`, eg I think it should be `clampInPlace((T)0.0,xadj,(T)1.0);`.
We might also need to adjust the original `TestScalar.cpp`:

From

```
SimTK_TEST(stepAny(-1,2,0,1,0.) == -1);
SimTK_TEST(stepAny(-1,2,0,1,.5) == 0);
SimTK_TEST(stepAny(-1,2,0,1,1.) == 1);
SimTK_TEST(stepAny(-1,2,0,1,0.f) == -1);
SimTK_TEST(stepAny(-1,2,0,1,.5f) == 0);
SimTK_TEST(stepAny(-1,2,0,1,1.f) == 1);
```


```
SimTK_TEST(stepAny(-1.,2.,0.,1.,0.) == -1);
SimTK_TEST(stepAny(-1.,2.,0.,1.,.5) == 0);
SimTK_TEST(stepAny(-1.,2.,0.,1.,1.) == 1);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,0.f) == -1);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,.5f) == 0);
SimTK_TEST(stepAny(-1.f,2.f,0.f,1.f,1.f) == 1);
```

I am still interested to know if we need to specify the function template arguments though (see previous comment) :)

Thanks


</blockquote></details>

You should not need to specify the template arguments for `stepUp()`; they should be deduced by the compiler. Also, the second template parameter has a default value, so you don't need to specify it.

And yes, you're right that you would need to alter `TestScalar`. This is technically a backwards-incompatible change, so we'll have to see if @sherm1 is okay with it. We could potentially use the templated functions to share the implementation of the previously-overloaded functions. For example:

```cpp
template <typename T, typename = enable_if_floating_point<T>>
T stepUpImpl(T x) { ... }

double stepUp(double x) { return stepUpImpl<double>(x); }
```

I wouldn't make this change until @sherm1 chimes in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L4hgUCwVSTmMtOgPW-b:b-wsvasg) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

You should not need to specify the template arguments for `stepUp()`; they should be deduced by the compiler. Also, the second template parameter has a default value, so you don't need to specify it.

And yes, you're right that you would need to alter `TestScalar`. This is technically a backwards-incompatible change, so we'll have to see if @sherm1 is okay with it. We could potentially use the templated functions to share the implementation of the previously-overloaded functions. For example:

```cpp
template <typename T, typename = enable_if_floating_point<T>>
T stepUpImpl(T x) { ... }

double stepUp(double x) { return stepUpImpl<double>(x); }
```

I wouldn't make this change until @sherm1 chimes in.
</blockquote></details>

Thanks for the explanation!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
I made minor changes in a couple of files that were necessary due to the implementation of template functions in `Scalar.h`. All is fine now, Travis CI is ok and all tests I wrote pass. This PR is therefore ready for another review pass.

sherm1:
Looks great! The only concern I have is that some of the step() functions lost the ability to accept integers as bounds. Those should be easy to add since they just need to be cast to T and then forwarded to the appropriate method.

---

Reviewed 2 of 4 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/include/SimTKcommon/internal/Function.h, line 395 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56m7EFKRF9897SXwup:-L56m7EFKRF9897SXwuq:b-eg7881) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/include/SimTKcommon/internal/Function.h#L395)):*
> ```objc
>         if ((x-m_x1)*m_sign >= 0) return m_y1;
>         // f goes from 0 to 1 as x goes from x0 to x1.
>         const Real f = stepAny((Real)0,(Real)1,m_x0,m_ooxr, x);
> ```

BTW should allow integers to work.

---

*[SimTKcommon/include/SimTKcommon/internal/Function.h, line 416 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56mBLGEi_tdjkzg4jB:-L56mBLGEi_tdjkzg4jC:b-mu7zc8) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/include/SimTKcommon/internal/Function.h#L416)):*
> ```objc
>           case 1: return dstepAny ((Real)1,m_x0,m_ooxr, x) * m_yr;
>           case 2: return d2stepAny((Real)1,m_x0,m_ooxr, x) * m_yr;
>           case 3: return d3stepAny((Real)1,m_x0,m_ooxr, x) * m_yr;
> ```

BTW should allow integer here.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L56jvJufTe5NPEkUtnB:b-ppbn52) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

Thanks for the explanation!
</blockquote></details>

Sorry I'm late to the party -- why do you want to give up the int-accepting forms of these methods. I added those because not having them caused trouble in practice with hard-to-decipher compiler errors. Usually they are very simple to add since they just need to call one of the float methods with a cast. I'm not happy with those getting lost and don't understand why that is necessary? Maybe you could use partial specialization to provide a generic int-accepting form that would work with all three floating types?

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 565 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56hqtNlxs1aRRVo6tk:-L56hqtNlxs1aRRVo6tl:b-qhaead) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L565)):*
> ```objc
> 
>     ADOL-C: the clamping is not done in-place for adoubles to avoid using if
>     statements and allow taping. */
> ```

BTW not sure why you say "not done in-place" here? These are definitely working in place (that is, the value v gets changed). You are using min/max rather than if statements but still working in place. Why does that internal detail need to appear in the doxygen comments to end users? Maybe include that as a developer comment instead? (Meaning use `//` rather than `/**`.)

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L56ikTYmH5ai1P257GY:b141c1m) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*
> ```objc
>     /** @copydoc SimTK::clampInPlace(double,double&,double)
> 
>     ADOL-C: the user will not be able to tape when using this function. */
> ```

BTW this shouldn't behave differently from the non-negated form (that is, it should work for taping). If necessary you can convert to adouble, clamp there, and then write it back (be sure to test).

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 254 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56lTgCugknUdnUVsKy:-L56lTgDQtr16sa_QLyu:bgs24p0) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/tests/TestADOLCCommon.cpp#L254)):*
> ```cpp
>     SimTK_TEST(vresal[1] == a+c);
>     SimTK_TEST(vresal[2] == a+d);
>     // substraction
> ```

BTW typo

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 308 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56l_HxgGirkFJS9J6_:-L56l_HxgGirkFJS9J6a:bgs24p0) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/tests/TestADOLCCommon.cpp#L308)):*
> ```cpp
>     SimTK_TEST(mresal[0][1] == m[0][1]);
>     SimTK_TEST(mresal[1][1] == a+e);
>     // substraction
> ```

BTW typo

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L56mE0T5hSerZk3_Cge:b-8o56r)*

chrisdembia:
Reviewed 2 of 4 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L5AWAXZsLBFwI6Zx8pp:bwi84zy) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*
We don't want to give up the int-accepting form. It was just an unintended consequence of trying to reuse the implementations by using function templates.

The overloads `int stepUp(int)` and `int stepDown(int)` still exist.

> Maybe you could use partial specialization to provide a generic int-accepting form that would work with all three floating types?

I don't follow.

My thought was the following:

```cpp
template <typename T, typename = enable_if_floating_point<T>>
T stepAnyImpl(T y0, T yRange,
                     T x0, T oneOverXRange,
                     T x) 
{   T xadj = (x-x0)*oneOverXRange;    
    assert(-NTraits<T>::getSignificant() <= xadj
           && xadj <= 1 + NTraits<T>::getSignificant());
    clampInPlace((T)0,xadj,(T)1);
    return y0 + yRange*stepUp(xadj); }

inline float stepAny(float y0, float yRange,
                     float x0, float oneOverXRange,
                     float x) {
    return stepAnyImpl<float>(y0, yRange, x0, oneOverXRange, x);
}
```


We could also just duplicate the implementations for each type, as is done in the existing code base.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 565 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56hqtNlxs1aRRVo6tk:-L5AQ5BXwSNLnRA7YgKw:bhkxrdq) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L565)):*

BTW not sure why you say "not done in-place" here? These are definitely working in place (that is, the value v gets changed). You are using min/max rather than if statements but still working in place. Why does that internal detail need to appear in the doxygen comments to end users? Maybe include that as a developer comment instead? (Meaning use `//` rather than `/**`.)
</blockquote></details>

I had suggested this comment to make it clear that memory is allocated here (whereas no new memory is necessary in the non-adouble `clampInPlace()` functions. I agree with moving the comment to a developer comment.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5AW3qgXPabNBDnb8mQ:bb6mt9b) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

BTW this shouldn't behave differently from the non-negated form (that is, it should work for taping). If necessary you can convert to adouble, clamp there, and then write it back (be sure to test).
</blockquote></details>

:+1:

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5Furcs-Y7dp3JNJYoD:b99607z) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

:+1:
</blockquote></details>

Are you meaning something like this:

```
inline negator<adouble>& clampInPlace(adouble low, negator<adouble>& v,
                                          adouble high)
    {
        const adouble& temp = v;
        adouble temp2 = temp;
        clampInPlace(low, temp2, high);
        v = temp2;
        return v;
    }
```

I do not think I got it right with the `const` and `&` btw.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

chrisdembia:
*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5GN7XlV2f0S9qJmh7v:b8dnp3f) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

Are you meaning something like this:

```
inline negator<adouble>& clampInPlace(adouble low, negator<adouble>& v,
                                          adouble high)
    {
        const adouble& temp = v;
        adouble temp2 = temp;
        clampInPlace(low, temp2, high);
        v = temp2;
        return v;
    }
```

I do not think I got it right with the `const` and `&` btw.
</blockquote></details>

Would the following work? 

```cpp
inline negator<adouble>& clampInPlace(adouble low, negator<adouble>& v,
                                          adouble high)
    {
        adouble temp = v;
        v = clampInPlace(low, temp, high);
        return v;
    }
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
I made all required changes. It seems fine now. I think this PR is ready for a new review pass.

---

Review status: 3 of 4 files reviewed at latest revision, 7 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 903 at r1](https://reviewable.io:443/reviews/simbody/simbody/618#-L4YGchsewMtCcGY8eki:-L5HN2dzAv3zN1a6M5FC:b-b3rp25) ([raw file](https://github.com/simbody/simbody/blob/f69916b21ec2130bee71eb8c32466d9923633042/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L903)):*

We don't want to give up the int-accepting form. It was just an unintended consequence of trying to reuse the implementations by using function templates.

The overloads `int stepUp(int)` and `int stepDown(int)` still exist.

> Maybe you could use partial specialization to provide a generic int-accepting form that would work with all three floating types?

I don't follow.

My thought was the following:

```cpp
template <typename T, typename = enable_if_floating_point<T>>
T stepAnyImpl(T y0, T yRange,
                     T x0, T oneOverXRange,
                     T x) 
{   T xadj = (x-x0)*oneOverXRange;    
    assert(-NTraits<T>::getSignificant() <= xadj
           && xadj <= 1 + NTraits<T>::getSignificant());
    clampInPlace((T)0,xadj,(T)1);
    return y0 + yRange*stepUp(xadj); }

inline float stepAny(float y0, float yRange,
                     float x0, float oneOverXRange,
                     float x) {
    return stepAnyImpl<float>(y0, yRange, x0, oneOverXRange, x);
}
```


We could also just duplicate the implementations for each type, as is done in the existing code base.

</blockquote></details>

I followed the suggestion of @chrisdembia. It seems to work fine. I also added comments.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 565 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56hqtNlxs1aRRVo6tk:-L5HMowkd_vPqDdW_lf3:bhcso0g) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L565)):*

I had suggested this comment to make it clear that memory is allocated here (whereas no new memory is necessary in the non-adouble `clampInPlace()` functions. I agree with moving the comment to a developer comment.
</blockquote></details>

I adjusted the comments (trying to make them clearer) and moved them as developer comments too

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5HLwyfQ8zUbfHigofe:bp39dvr) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

Would the following work? 

```cpp
inline negator<adouble>& clampInPlace(adouble low, negator<adouble>& v,
                                          adouble high)
    {
        adouble temp = v;
        v = clampInPlace(low, temp, high);
        return v;
    }
```
</blockquote></details>

Yes, sorry...

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 254 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56lTgCugknUdnUVsKy:-L5HMaqCzw0fl4nZnuFQ:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/tests/TestADOLCCommon.cpp#L254)):*

BTW typo
</blockquote></details>

Done.

---

*[SimTKcommon/tests/TestADOLCCommon.cpp, line 308 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56l_HxgGirkFJS9J6_:-L5HMbQXp5Uh1GCxxxb-:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/tests/TestADOLCCommon.cpp#L308)):*

BTW typo
</blockquote></details>

Done.

---

*[SimTKcommon/include/SimTKcommon/internal/Function.h, line 395 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56m7EFKRF9897SXwup:-L5HMee0szo1pIwbtWEb:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/include/SimTKcommon/internal/Function.h#L395)):*

BTW should allow integers to work.
</blockquote></details>

Done.

---

*[SimTKcommon/include/SimTKcommon/internal/Function.h, line 416 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56mBLGEi_tdjkzg4jB:-L5HMfydG1SBeAj-m78_:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/include/SimTKcommon/internal/Function.h#L416)):*

BTW should allow integer here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L5HNPnAxnxCMEzTq7NB:b-vfnhy7)*

chrisdembia:
Reviewed 1 of 4 files at r2, 2 of 3 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5Lur6b42MbI0878zgo:b1kws71) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

Yes, sorry...
</blockquote></details>

I lost track; do you have tests for this function to ensure it works correctly (same behavior as the equivalent `double` overload)?

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 931 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvM4ob-uDZFgR74qy:-L5LvM4ob-uDZFgR74qz:bsft0s7) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L931)):*
> ```objc
> // We use the templated function stepAnyImpl() to share the implementations of
> // stepAny() with floating point types (double, float, and adouble (if using
> // ADOL-C). This allows maintaining the ability to use integers as bounds.
> ```

I think you're missing a closing parenthesis.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1054 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvUvWbpedFc5xLy8F:-L5LvUvWbpedFc5xLy8G:b-v69tmx) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1054)):*
> ```objc
> // We use the templated function dstepAnyImpl() to share the implementations of
> // dstepAny() with floating point types (double, float, and adouble (if using
> // ADOL-C). This allows maintaining the ability to use integers as bounds.
> ```

Missing parenthesis.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1153 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvahNeWBIdDwgrMEq:-L5LvahNeWBIdDwgrMEr:b-5l6dux) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1153)):*
> ```objc
> // We use the templated function d3stepAnyImpl() to share the implementations
> // of d3stepAny() with floating point types (double, float, and adouble (if
> // using ADOL-C). This allows maintaining the ability to use integers as
> ```

Missing parenthesis

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5Lzt9_tpgSuYmygZf6:bilbfwh) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

I lost track; do you have tests for this function to ensure it works correctly (same behavior as the equivalent `double` overload)?
</blockquote></details>

There is [this test](https://github.com/antoinefalisse/simbody/blob/adolc_scalar/SimTKcommon/tests/TestADOLCCommon.cpp#L427) ensuring that it works as its `double` counterpart.

But I can also those two if this is what you suggest:

```
    {   double h = 4;
        adouble hd = 4;
        auto nha = reinterpret_cast<const negator<double>&>(h);
        double hb = nha;
        nha = clampInPlace(b,hb,a);
        auto nhb = reinterpret_cast<const negator<double>&>(h);
        auto nhd = reinterpret_cast<const negator<adouble>&>(hd);
        SimTK_TEST(clampInPlace(b,nhb,a)==nha &&
        clampInPlace(b,nhb,a)==clampInPlace(b,nhd,a) && nhd==b); }
    {   double h = 4;
        adouble hd = 4;
        auto nha = reinterpret_cast<const negator<double>&>(h);
        double hb = nha;
        nha = clampInPlace(b,hb,a);
        auto nhb = reinterpret_cast<const negator<double>&>(h);
        auto nhd = reinterpret_cast<const negator<adouble>&>(hd);
        SimTK_TEST(clampInPlace(b,nhb,a)==nha &&
        clampInPlace(b,nhb,a)==clampInPlace(bd,nhd,ad) && nhd==b); }
```

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 931 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvM4ob-uDZFgR74qy:-L5LyzvF8op_lSOvWBTa:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L931)):*

I think you're missing a closing parenthesis.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1054 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvUvWbpedFc5xLy8F:-L5Lz08IHPlRV7fMrrTD:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1054)):*

Missing parenthesis.
</blockquote></details>

Done.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1153 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvahNeWBIdDwgrMEq:-L5Lz1JjNH2gZKbXLAcv:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1153)):*

Missing parenthesis
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

chrisdembia:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5M9rbrrj8HXmZrTVb1:baskbhr) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

There is [this test](https://github.com/antoinefalisse/simbody/blob/adolc_scalar/SimTKcommon/tests/TestADOLCCommon.cpp#L427) ensuring that it works as its `double` counterpart.

But I can also those two if this is what you suggest:

```
    {   double h = 4;
        adouble hd = 4;
        auto nha = reinterpret_cast<const negator<double>&>(h);
        double hb = nha;
        nha = clampInPlace(b,hb,a);
        auto nhb = reinterpret_cast<const negator<double>&>(h);
        auto nhd = reinterpret_cast<const negator<adouble>&>(hd);
        SimTK_TEST(clampInPlace(b,nhb,a)==nha &&
        clampInPlace(b,nhb,a)==clampInPlace(b,nhd,a) && nhd==b); }
    {   double h = 4;
        adouble hd = 4;
        auto nha = reinterpret_cast<const negator<double>&>(h);
        double hb = nha;
        nha = clampInPlace(b,hb,a);
        auto nhb = reinterpret_cast<const negator<double>&>(h);
        auto nhd = reinterpret_cast<const negator<adouble>&>(hd);
        SimTK_TEST(clampInPlace(b,nhb,a)==nha &&
        clampInPlace(b,nhb,a)==clampInPlace(bd,nhd,ad) && nhd==b); }
```
</blockquote></details>

Thanks for pointing out that previous test. I'm having trouble wrapping my head around what your newly-proposed test is testing. I think the test you link to is sufficient. It's up to you if you want to include this additional test, though if you do it would be good to include a comment.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1153 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvahNeWBIdDwgrMEq:-L5MAVj6zlvu6ywQsJJM:b-l42ghv) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1153)):*

Done.
</blockquote></details>

I'm not seeing the fix. The issue may be related to merge conflicts that I caused by merging your Mat Vec PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

antoinefalisse:
Review status: 0 of 2 files reviewed at latest revision, 6 unresolved discussions.

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 685 at r2](https://reviewable.io:443/reviews/simbody/simbody/618#-L56ikTYmH5ai1P257GX:-L5MC6eUx1pq2UoQv_uX:b440k4h) ([raw file](https://github.com/simbody/simbody/blob/dab8d4b917bb8e1e5737da89b76bdd6796981cdf/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L685)):*

Thanks for pointing out that previous test. I'm having trouble wrapping my head around what your newly-proposed test is testing. I think the test you link to is sufficient. It's up to you if you want to include this additional test, though if you do it would be good to include a comment.
</blockquote></details>

Ok, I removed these two tests. I thought you wanted to make sure that casting to a `adouble` was fine (so what we do in the define of the function) so I had reproduced the same scheme with `double` (not sure it is clear). Anyways, I think the original tests are enough too

---

*[SimTKcommon/Scalar/include/SimTKcommon/Scalar.h, line 1153 at r3](https://reviewable.io:443/reviews/simbody/simbody/618#-L5LvahNeWBIdDwgrMEq:-L5MClwe0OveQJPKTshU:b-c5uiae) ([raw file](https://github.com/simbody/simbody/blob/1c26954884bf931bb75e5798d7df60ef11a8dcac/SimTKcommon/Scalar/include/SimTKcommon/Scalar.h#L1153)):*

I'm not seeing the fix. The issue may be related to merge conflicts that I caused by merging your Mat Vec PR.
</blockquote></details>

Sorry I hadn't pushed yet (but I had done locally already). Should be fine now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618)*

chrisdembia:
:lgtm:

---

Reviewed 2 of 4 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L5kGA7IEVvzJlzHRNqi:bnfp4nl)*

chrisdembia:
@sherm1: the `step*` functions can accept integer bounds again. Can you take another pass at this when you get a chance?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L6saHRfK4pn9m6NvWeG:b-b6uvfy)*

sherm1:
Looks great -- thanks! :lgtm:

---

Reviewed 1 of 3 files at r3, 2 of 4 files at r4.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L7MhoWuZ_QoQccCQEuS:bbzy36i)*

sherm1:
This is ready to merge, @antoinefalisse.

chrisdembia:
@sherm1 what merge policy do you use now? Is it that anyone can merge once `code-review/reviewable` has a green check?

antoinefalisse:
@chrisdembia, I don't think I have the rights to merge. Can you do it? Thanks.

sherm1:
Sorry, I forgot that Antoine's branch is in the Simbody repo rather than in a different fork. Either you or I should do the merge, @chrisdembia. This one could benefit from a squash-merge since it is spread over 13 commits. I'll leave it to you to do that Chris, because that requires creating a good commit message to describe the overall intent of the PR.

antoinefalisse:
BTW @chrisdembia, before merging we may want to adjust the unclear comments [here ](https://reviewable.io/reviews/simbody/simbody/626#-) as suggested by @sherm1 since `scalar.h` is part of this PR (sorry for the lack of clarity). Do you have any ideas for clarifying those comments?

sherm1:
I think I understand better now after reading the discussion. Is the intent to note that there will be heap activity due to the need to allocate intermediate `adouble` expressions? And therefore the operation will be much slower in adouble than double? Maybe it could just say that directly in the `//` comment?

chrisdembia:
@sherm1 yes. 

@antoinefalisse after you update the Scalar comments, I'll squash-merge.

antoinefalisse:
@chrisdembia, I adjusted the comments. If you are fine with them, I think this PR can be merged.

chrisdembia:
:lgtm:

---

Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/simbody/simbody/618#-:-L7WKePyt0N6E43NS0I1:bnfp4nl)*

chrisdembia:
Great job, @antoinefalisse

PR #624, opened 2018-02-22
[doxygen] Fix brief description for HuntCrossleyContact

tkuchida:
From https://simbody.github.io/simbody-3.6-doxygen/api/classSimTK_1_1HuntCrossleyContact.html:

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/624)

PR #625, opened 2018-02-22
[doxygen] Revert brief description of HuntCrossleyContact

tkuchida:
Return full citation to brief description per original author's intention.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/625)


sherm1:
Miscommunication, sorry: I liked the change in #624 -- no need to change it back unless you actually prefer it this way.

PR #654, opened 2018-11-15
FreeBSD compatibility patch

yurivict:


chrisdembia:
What testing have you tried?

yurivict:
> What testing have you tried?

The 'test' target.

yurivict:
ping

SimTKcommon/src/ParallelExecutor.cpp
```diff
@@ -29,6 +29,11 @@
 #include <algorithm>
 #include <mutex>
 
+#if defined(__FreeBSD__)
```
chrisdembia: Consider using a consistent style. `#ifdef ...`

yurivict: ```#if defined()``` is more generic, another clauses can be easily added to it, like ```#if defined(A) || defined(B)```.

adamkewley:

PR #655, opened 2018-11-15
Add INSTALL_DOCS variable

yurivict:
If the package has DOCS=off option, it is easiest when CMakeLists.txt supports this option too.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/655)


chrisdembia:
Won't this cause docs to not be installed by default?

yurivict:
> Won't this cause docs to not be installed by default?

You are right. I made it ```ON``` by default.

yurivict:
Could you please consider merging this PR?

Thanks,
Yuri

Issue #658, opened 2018-11-23
Group-writable files after installation


morxa:
After installation, a number of files is group-writable, which is uncommon. The affected files:
```
/usr/lib64/libSimTKcommon.so.3.6
/usr/lib64/libSimTKmath.so.3.6
/usr/lib64/libSimTKsimbody.so.3.6
/usr/libexec/simbody/simbody-visualizer
```

Is there a particular reason for this?

Background: I'm currently packaging simbody for Fedora, and this was noticed by `rpmlint`. I can fix this downstream, but  I'd prefer to fix this upstream. I can submit a PR if you're interested.

chrisdembia:
We explicitly set these permissions (see [here](https://github.com/simbody/simbody/search?q=PERMISSIONS&unscoped_q=PERMISSIONS)) but I think it is not necessary. I think a PR would be great but we should probably wait to hear what @sherm1 thinks.

sherm1:
A PR sounds good to me! I don't think they should be writable -- thanks.

richmattes:
I have a patch from when I tried to package this for Fedora several months ago - I'll try to update it and submit a PR

PR #660, opened 2018-11-24
Don't set permissions when installing

richmattes:
Remove manually set permissions when installing targets and files.
Addresses #658.

Signed-off-by: Rich Mattes <richmattes@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/660)


sherm1 (commented):
If you remove _all_ the permissions, do you get the right ones for all installed files? Even on Windows?

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/660#-:-LS6zKrD5j_e5n4c7dII:b-347b6e)* status: 0 of 8 files reviewed, all discussions resolved

richmattes:
I guess it depends on what you mean by "right."  By [default](https://cmake.org/cmake/help/v3.12/command/install.html), `install(FILES...` uses 0644 permissions (user rwx, group r, world r) on files and 0755 (user rwx, group rx, and world rx) on directories.   `install(PROGRAMS...` sets the file permissions to 0755, which is intended to be used for things like scripts that need to be executable but aren't targets.  `install(TARGETS...` applies 0755 permissions to executables and libraries that are installed, though I can't find any specific statements confirming that in the documentation.

In practice, I've never seen anyone override CMake permissions - they tend to always use 644 for files and 755 for directories, scripts and executables.

chrisdembia (approved):
OpenSim does not set permissions explicitly and we have no issues.

sherm1 (commented):
I don't think CI is able to check that installation works properly. I'm worried that since Simbody uses `install(FILES...` to install DLLs, the default 0644 permissions won't be correct. Someone went to a lot of trouble to set those permissions -- I suspect there was a reason for that. Might have been due to an old version of CMake, but I'm not sure. Let's make sure we've tested this on all three platforms before merging it. Any platform volunteers?

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/660#-:-LSBaxytBAU7RbVPG5BF:bvjpl6d)* status: 0 of 8 files reviewed, all discussions resolved

richmattes:
Switching to `install(PROGRAMS...` for the DLLs in the `Platform` subdirectory might correct that.  The LIB files appear to be treated consistently with the other static targets.

Other than that, the only change between the default CMake permissions and the manually set permissions is that Simbody adds GROUP_WRITE to everything.  Is there something in simbody's development or deployment workflow that requires GROUP_WRITE?

adamkewley:

Note: in this PR's case, the installation steps in simbody have been changed by other PRs, so this issue may now be fixed either way.

Issue #662, opened 2018-12-28
Is Explicitly Setting CMP0003 and CMP0005 Necessary?


artin-f:
Two CMake policies are explicitly set to `NEW` here:
https://github.com/simbody/simbody/blob/43c2645c634bc93488deab52da231708b4654a9f/CMakeLists.txt#L21-L24

Both `CMP0003` and `CMP0005` were added in CMake version `2.6` [[ref]](https://cmake.org/cmake/help/v3.13/manual/cmake-policies.7.html#manual:cmake-policies(7)).  Since Simbody requires CMake version `2.8.10` and above, both policies are set to `NEW` by default.  Therefore, explicitly setting them seems unnecessary.

The above code doesn't cause any issues in Simbody operation, but removing it will improve readability and maintainability a little bit.

artin-f:
Great, thank you for your feedback.  You can assign this issue to me.  I'll open a PR shortly.

sherm1:
Thanks, Artin!

PR #663, opened 2018-12-28
Remove Unnecessary CMake Policies

artin-f:
This PR addresses issue #662.  It removes the explicit setting of CMake policies CMP0003 and CMP0005.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/663)


artin-f:
Thanks for the tip @sherm1.

Issue #666, opened 2019-04-19
Smooth Contact Model


antoinefalisse:
Hi @sherm1 and @chrisdembia,

@gilserrancoli and I have been using a smooth approximation of the Hunt-Crossley contact model available in Simbody for our applications (OpenSim-based predictive simulations of movements). The rationale for using such smooth model was the use of gradient-based optimization that requires twice-continuously differentiable models. 

We believe this smooth model could be useful for the community, in particular with software like Moco, and therefore propose to merge it into Simbody.

sherm1:
Resolved by #667.

PR #667, opened 2019-04-19
Smooth contact model

antoinefalisse:
@sherm1 and @chrisdembia,

This PR contains the files implementing the smooth contact model mentioned in #666.

The model is based on the original implementation of the Hunt-Crossley contact model available in Simbody. Yet the proposed implementation differs from the original implementation as it does not rely on a `GeneralContactSubsystem`. Instead, it assumes contact between a sphere and a plane that can have an offset and a slope with respect to the ground. Originally, our motivation for not relying on a `GeneralContactSubsystem` was to be able to use algorithmic differentiation in our optimal control problems (which is working fine although we are no longer using ADOL-C as we discussed earlier but are using CasADi instead). Since our applications rely on simple contacts between spheres and plane (e.g., interaction feet-ground during walking), we did not need to use a `GeneralContactSubsystem` and could use a simplified contact model. This also has the advantage that we only need to realize to velocity (instead of dynamics) in our applications, which is computationally beneficial.

The proposed smooth contact model assumes the same properties for both contacting materials and relies on hyperbolic tangent functions for smoothing the _if statements_ of the original model. For example, the following if statement:
```
y = 0, if x < d
y = a, if x >= d
```
can be approximated by:
```
f = 0.5 + 0.5*tanh(b(x-d))
y = a*f
```
where `b `is a parameter determining the smoothness of the transition. We use `tanh `functions in the expression of the Hertz force (transition between zero force and existing force when there is indentation) and in the expression of the Hunt-Crossley force (transition between zeros force and existing force depending on the indentation velocity). We have assigned constant values to the parameters determining the `tanh `transition smoothness in both cases. Another feature of our proposed contact model is that is assumes positive force at all time. In practice, there is thus a force even if there is no contact. The equations we use in the model are described in `HuntCrossleyForce_smooth.h`.

Our proposed implementation lives in three files:

- HuntCrossleyForce_smooth.h

- HuntCrossleyForceImpl_smooth.h

- HuntCrossleyForce_smooth.cpp

We also worked out a test (very similar to the original test):

- TestHuntCrossleyForce_smooth.cpp

We welcome any feedback about whether you would be keen in integrating such model in Simbody and about how to improve the model. We can also provide you with more information about how the model behaves. As is now, there is no comparison with the original model. We can work on that but we first wanted to know whether you would consider having such model in Simbody.

 @chrisdembia, I was thinking to use this model in Moco this summer; please shoot if you think it should have other features or is not designed in the right way.

Ps: perhaps we could consider another name for the model as well.

Thanks in advance,
@GilSerrancoli and Antoine

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/667)


sherm1:
Hi, @antoinefalisse. This sounds useful. Before embarking on a review though, can you explain why you want this to be a Simbody built-in force rather than a [Custom Force element](https://simbody.github.io/simbody-3.6-doxygen/api/classSimTK_1_1Force_1_1Custom.html#details)? At first glance it looks like it could be implemented through CustomForce.

Thanks and regards,
Sherm
cc @chrisdembia

chrisdembia:
That might have been my push. The hope is that the force would be part of Simbody proper, and my understanding was that custom forces are more for third-party plugins, not for forces part of `SimTKsimbody`.

chrisdembia:
> Can we pick a different name for this? The proliferation of HuntCrossley's in Simbody is confusing. I would propose something like "SmoothSphereHalfplaneContact" or "DifferentiableSphereHalfplaneContact" or some other name that reflects what is unique about this particular force element and distinguishes it from other models in Simbody. The implementation details are secondary and can appear in the theory section of the comments.

The other contact force models have names ending in "Force," not "Contact." Should we try to be consistent here?

antoinefalisse:
We had changed to "SmoothSphereHalfplaneContact" but we can go for "SmoothSphereHalfplaneForce".

antoinefalisse:
@sherm1 and @chrisdembia
This PR is ready for a second review. Thanks in advance.
Antoine and @gilserrancoli

antoinefalisse:
Hi @sherm1, if you have a chance to take a look at the PR in the coming days that would be awesome. I will be in Stanford in about 10 days for 5 weeks to work on @chrisdembia's library for optimal control (Moco). The goal is to reproduce some of the predictive simulations of walking I have performed. That's why I would like to use the contact model. No rush though. Thanks very much, Antoine.

sherm1:
Will do asap (on vacation for a few more days).

Simbody/src/SmoothSphereHalfplaneForce.cpp
```diff
@@ -0,0 +1,265 @@
...
+    pe += Real(2./5.)*fH*indentation;
+    // Calculate the Hunt-Crossley force.
+    const Real c = dissipation;
+    const Real fHd = fH*(1.+(3./2.)*c*indentationVel);
+    const Real fn = fHd*(1./2.+(1./2.)*std::tanh(bd*indentation))*
+        (1./2.+(1./2.)*std::tanh(bv*(indentationVel+(2./(3.*c)))));
+    Vec3 force = fn*normal;
+    // Calculate the friction force.
+    const Real aux = pow(vtangent[0],2) +
+        pow(vtangent[1],2)+pow(vtangent[2],2)+eps;
```
chrisdembia: Should be able to use `vtangent.normSqr()` or something like that.

Simbody/include/simbody/internal/SmoothSphereHalfplaneForce.h
```diff
@@ -0,0 +1,126 @@
...
+ * available in Simbody. The proposed implementation was designed for use
+ * with gradient-based optimization algorithms and algorithmic/automatic
+ * differentiation. To this aim, conditional if statements were approximated by
+ * using hyperbolic tangent functions. For example, the following if statement:
+ * <pre>     y = 0, if x < d </pre>
+ * <pre>     y = a, if x >= d </pre>
+ * can be approximated by:
+ * <pre>     f = 0.5 + 0.5 tanh(b(x-d)) </pre>
+ * <pre>     y = a f </pre>
+ * where b is a parameter determining the smoothness of the transition.
```
chrisdembia: This class also applies force when the sphere is not in contact, correct?

It would be great to add a plot to the doxygen showing the force as a function of depth. I could help with this.

sherm1:
Thanks for doing that analysis, @gilserrancoli!

I definitely don't think we should attempt to numerically integrate the PE. That creates too many difficulties for too little gain. Instead, an analytic approximation should be good enough provided it is well documented that the PE is approximate.

However, I am concerned if the forces begin before contact but the PE calculation does not. Zero is not a good approximation in that case! Can we have an analytic approximation that includes all of the force-generating regime?

sherm1:
Merged -- thanks @antoinefalisse @gilserrancoli and @chrisdembia !

chrisdembia:
Yay! Thank you, @sherm1. This will help us continue to make good progress this summer.

antoinefalisse:
Thanks a lot @sherm1 and @chrisdembia for your reviews and willingness to teach me good practice!

PR #670, opened 2019-07-05
Fix custom acceleration-only constraint to call calcAccelerationErrors().

### LINKED ISSUE #669: Constraint::Custom::Implementation override wrong function when implementing acceleration constraint
I define a acceleration constraint implementation inherited from Constraint::Custom::Implementation
I am sure I initialize Implementation(matter, 0, 0, 1) and override the calcAccelerationErrors function
But it still ask me to override the calcVelocityDotErrors because in runtime it is called and virtual function was called.
After I override calcVelocityDotErrors in the acceleration constraint implementation, it works.
[PrescribedAcc.txt](https://github.com/simbody/simbody/files/3341072/PrescribedAcc.txt)

sherm1:
Per issue #669, a custom acceleration-only constraint invoked the wrong virtual method. Should call calcAccelerationErrors() rather than calcVelocityDotErrors().

Fixes #669.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/670)

### DIFF
--- a/Simbody/src/ConstraintImpl.h
+++ b/Simbody/src/ConstraintImpl.h
@@ -3034,7 +3034,7 @@ void calcAccelerationErrorsVirtual
     const Array_<SpatialVec,ConstrainedBodyIndex>&  A_AB, 
     const Array_<Real,      ConstrainedUIndex>&     constrainedUDot,
     Array_<Real>&                                   aerr) const override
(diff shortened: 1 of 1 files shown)


dpkong:
I've verified the stackcall and the calcAccelerationErrors is coorrectly called.
Also, a test confirms that the acceleration is correctly defined.
This should work well.

PR #671, opened 2019-07-07
Add Constraint::PrescribedVelocity

dpkong:
This class is similar to Constraint::PrescribedMotion. 
It constrain one u of a MobilizedBody.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/671)

### DIFF
--- a/Simbody/include/simbody/internal/Constraint.h
+++ b/Simbody/include/simbody/internal/Constraint.h
@@ -475,6 +475,7 @@ class Custom;
 class CoordinateCoupler;
 class SpeedCoupler;
 class PrescribedMotion;
+class PrescribedVelocity;
 class PointOnPlaneContact; 
 class SphereOnPlaneContact; // ball in contact with plane (sliding or rolling)
 class SphereOnSphereContact; // ball in contact with ball (sliding or rolling)
@@ -496,6 +497,7 @@ class CustomImpl;
 class CoordinateCouplerImpl;
 class SpeedCouplerImpl;
 class PrescribedMotionImpl;
+class PrescribedVelocityImpl;
 class PointOnPlaneContactImpl; 
 class SphereOnPlaneContactImpl;
 class SphereOnSphereContactImpl;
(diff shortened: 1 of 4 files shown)


sherm1 (changes requested):
Thanks for writing this! I reviewed it and it looks good with a few comments below. Also, this class needs a unit test to verify that it is working and that no one breaks it in the future.

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/671#-:-Lm26a_VB0s0JSLYg_wy:bxgvu5z)* status: all files reviewed, 4 unresolved discussions (waiting on @dpkong)

___
*[Simbody/include/simbody/internal/Constraint.h, line 2211 at r2](https://reviewable.io/reviews/simbody/simbody/671#-Lm24WBU2DDaxipj03-U:-Lm24WBVDkk8KYAyQvFi:bjjm94j) ([raw file](https://github.com/simbody/simbody/blob/f0e21fe49aca9567cc079a81618417e5704765a8/Simbody/include/simbody/internal/Constraint.h#L2211)):*
> ```objc
> provide a Function which takes the current time as its argument and returns the
> required speed of the generalized coordinate. It also must support derivatives
> up to second order. **/
> ```

The above comment needs to be rewritten to be correct for prescribed velocity. It appears to have been copied from prescribed motion, which prescribed the position of a generalized coordinate q as a function of time q=q(t). Here we are prescribing the position of a generalized speed u=u(t) (_not_ a generalized coordinate). 

In general the indexes for generalized coordinates q are different than those for generalized speeds u, and it is not always true that the generalized speeds are time derivatives of the generalized coordinates. So it is important not to mix them up -- please be precise here.
___
*[Simbody/include/simbody/internal/Constraint.h, line 2233 at r2](https://reviewable.io/reviews/simbody/simbody/671#-Lm23AA041LJuEg5L5Iu:-Lm23AA041LJuEg5L5Iv:bb4g0tp) ([raw file](https://github.com/simbody/simbody/blob/f0e21fe49aca9567cc079a81618417e5704765a8/Simbody/include/simbody/internal/Constraint.h#L2233)):*
> ```objc
> 		       const Function*            function,
> 		       MobilizedBodyIndex         speedBody,
> 		       MobilizerUIndex            speedIndex);
> ```

This file contains tabs. Those are not permitted by the [Simbody style guide](https://github.com/simbody/simbody/blob/master/CONTRIBUTING.md#coding-conventions).
___
*[Simbody/src/Constraint.cpp, line 2745 at r2](https://reviewable.io/reviews/simbody/simbody/671#-Lm25aAn0Ii3UbZ6-PVV:-Lm25aAo44FUz1NR8-Lc:b-wtvgo4) ([raw file](https://github.com/simbody/simbody/blob/f0e21fe49aca9567cc079a81618417e5704765a8/Simbody/src/Constraint.cpp#L2745)):*
> ```cpp
> 	const Function*             function,
> 	MobilizedBodyIndex          speedBody,
> 	MobilizerUIndex             speedIndex)
> ```

Tabs are not allowed.

sherm1:
Is this PR still active?

dpkong:
hi Mr. Sherman,
Sorry i left this behind.
I didn't create unit test for this branch. 
Allow me several days plz.

adamkewley:

PR #689, opened 2020-05-15
Fixed bounds warning (gcc 8.3.0)

adamkewley:
- Compiler warns: ‘strncpy’ specified bound depends on the length of the source argument
- This commit is a proposed refactoring of the function to simplify it, rather than a
  like-for-like fix to the implementation

gcc warns about this method when building OpenSim with a superbuild. When I looked into it, I thought that the `szCat` implementation is a bit complicated for what it's doing, so this also throws a refactor in (revert it if you disagree).

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/689)

### DIFF
--- a/SimTKmath/Optimizers/src/c-cmaes/cmaes.c
+++ b/SimTKmath/Optimizers/src/c-cmaes/cmaes.c
@@ -2476,10 +2476,6 @@ double cmaes_random_Uniform( cmaes_random_t *t)
   return (double)(t->aktrand)/(2.147483647e9);
 }
 
-static char *
-szCat(const char *sz1, const char*sz2, 
-      const char *sz3, const char *sz4);
-
 /* --------------------------------------------------------- */
 /* -------------- Functions: cmaes_readpara_t -------------- */
 /* --------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


chrisdembia:
Hey @adamkewley ; we've resisted making changes to this file because it's third-party code. Ideally, the changes would be made upstream first (https://github.com/CMA-ES/c-cmaes) to avoid issues when updating the file from upstream in the future.

Another option is to tell CMake to reduce warnings for just this file.

A third option is to have CMake apply a patch to this when configuring the project.

adamkewley:
Sure, I'll see if I can upstream it.

adamkewley:
PRed upstream:

https://github.com/CMA-ES/c-cmaes/pull/25

adamkewley:
@nikohansen from upstream is going to take a look at my patch, make some suggestions + modifications if necessary, etc.

Once it's upstreamed I'll close this off and open a separate PR for merging in the newer version so that we can clear that warning message

adamkewley:
Upstream is now patched in `master`:

https://github.com/CMA-ES/c-cmaes/commit/0bd2e9e8ec69e9dcedc2b78716a5db5c2edf4e13

I'll close this off and make a separate PR for merging the latest [c-maes](https://github.com/CMA-ES/c-cmaes) `master` branch into Simbody. Thanks for @nikohansen for looking into it so quickly.

PR #690, opened 2020-05-20
Updated vendored c-maes code from upstream

adamkewley:
- Upstream commit 0bd2e9e8ec69e9dcedc2b78716a5db5c2edf4e13
- Should remove compiler warning:
    "‘strncpy’ specified bound depends on the length of the source argument"

This PR aims to update the vendored c-maes code in the repo to the latest upstream version.

The significance of this work is that the current (older) c-maes code currently produces these kinds of warnings when compiled in later (>v8) gcc version:

```
simbody/SimTKmath/Optimizers/src/c-cmaes/cmaes.c: In function ‘szCat’:
simbody/SimTKmath/Optimizers/src/c-cmaes/cmaes.c:3118:3: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
   strncpy ((char *)szBuf, sz1, (unsigned)intMin( (int)strlen(sz1), 698));
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Previous discussions about this fix are in [a previous PR](https://github.com/simbody/simbody/pull/689). Ultimately, the resolution was to [upstream a patch to c-maes](https://github.com/CMA-ES/c-cmaes/pull/25), which was merged in [this](https://github.com/CMA-ES/c-cmaes/commit/0bd2e9e8ec69e9dcedc2b78716a5db5c2edf4e13) upstream commit. This PR essentially tracks the work necessary to put the updated source files into Simbody, which should eliminate the warning.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/690)

### DIFF
--- a/SimTKmath/Optimizers/src/c-cmaes/cmaes.c
+++ b/SimTKmath/Optimizers/src/c-cmaes/cmaes.c
@@ -99,9 +99,13 @@
   14/10/18: splitted cmaes_init() into cmaes_init_para() and cmaes_init_final(),
             such that parameters can be set also comparatively safely within the
             code, namely after calling cmaes_init_para(). The order of parameters
+<<<<<<< HEAD
             of readpara_init() has changed to be the same as for cmaes_init().
   14/11/25  fix warnings from Microsoft C compiler (sherm)
+=======
+            of readpara_init() has changed to be the same as for cmaes_init(). 
   14/11/26: renamed exported symbols so they begin with a cmaes_prefix.
+>>>>>>> bd5aeb2a21fbc437f1b322610022f7cb1ba6a9db
   
   Wish List
     o make signals_filename part of cmaes_t using assign_string()
@@ -217,7 +221,6 @@ static double rgdouMax( const double *rgd, int len);
 static double rgdouMin( const double *rgd, int len);
 static double douMax( double d1, double d2);
 static double douMin( double d1, double d2);
-static int    intMin( int i, int j);
 static int    MaxIdx( const double *rgd, int len);
 static int    MinIdx( const double *rgd, int len);
 static double myhypot(double a, double b);
@@ -282,18 +285,18 @@ cmaes_init_final(cmaes_t *t /* "this" */)
   if (t->version == NULL) {
         ERRORMESSAGE("cmaes_init_final called (probably) without calling cmaes_init_para first",
                      ", which will likely lead to unexpected results",0,0);
-        printf("Warning: cmaes_init_final called (probably) without calling cmaes_init_para first\n");
+        printf("Error: cmaes_init_final called (probably) without calling cmaes_init_para first\n");
   }
   if (strcmp(c_cmaes_version, t->version) != 0) {
         ERRORMESSAGE("cmaes_init_final called twice, which will lead to a memory leak",
                      "; use cmaes_exit() first",0,0);
-        printf("Warning: cmaes_init_final called twice, which will lead to a memory leak; use cmaes_exit first\n");
+        printf("Error: cmaes_init_final called twice, which will lead to a memory leak; use cmaes_exit first\n");
   }
   /* assign_string(&t->signalsFilename, "cmaes_signals.par"); */
 
   if (!t->sp.flgsupplemented) {
     cmaes_readpara_SupplementDefaults(&t->sp);
-    if (!isNoneStr(t->sp.filename))
+    if (!isNoneStr(t->sp.filename)) /* TODO: should this be done in readpara_SupplementDefaults? */
       cmaes_readpara_WriteToFile(&t->sp, "actparcmaes.par");
   }
      
@@ -414,7 +417,6 @@ cmaes_init(cmaes_t *t, /* "this" */
 /* --------------------------------------------------------- */
 
 #ifdef __GNUC__
-    // These macros generate warnings in Visual Studio.
     #pragma GCC diagnostic push
     #pragma GCC diagnostic ignored "-Wunused-result"
 #endif
@@ -532,6 +534,8 @@ cmaes_resume_distribution(cmaes_t *t, char *filename)
   if (res != (t->sp.N*t->sp.N+t->sp.N)/2)
     FATAL("cmaes_resume_distribution(): C: dimensions differ",0,0,0); 
    
+  fclose(fp);
+  
   t->flgIniphase = 0;
   t->flgEigensysIsUptodate = 0;
   t->flgresumedone = 1;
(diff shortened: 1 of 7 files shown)


SimTKmath/Optimizers/src/c-cmaes/cmaes.c
```diff
@@ -99,9 +99,13 @@
   14/10/18: splitted cmaes_init() into cmaes_init_para() and cmaes_init_final(),
             such that parameters can be set also comparatively safely within the
             code, namely after calling cmaes_init_para(). The order of parameters
+<<<<<<< HEAD
             of readpara_init() has changed to be the same as for cmaes_init().
   14/11/25  fix warnings from Microsoft C compiler (sherm)
+=======
+            of readpara_init() has changed to be the same as for cmaes_init(). 
   14/11/26: renamed exported symbols so they begin with a cmaes_prefix.
+>>>>>>> bd5aeb2a21fbc437f1b322610022f7cb1ba6a9db
```
chrisdembia, line 108: Is this merge conflict in the c-cmaes codebase?

chrisdembia:
The Travis-CI test is failing because the c-cmaes files contain tab characters. In my opinion, replacing the tabs with spaces does not need to be upstreamed. We could also update the `grep` command we run on Travis-CI to omit the `c-cmaes` directory.

SimTKmath/Optimizers/src/c-cmaes/cmaes.c
adamkewley: Yeah: https://github.com/CMA-ES/c-cmaes/blob/master/src/cmaes.c#L108

adamkewley: Seems to have been pulled in ~6 years ago:

```
ffdc1025 (Nikolaus Hansen    2014-12-10 19:49:29 +0100  102) <<<<<<< HEAD
97ada5f3 (Michael Sherman    2014-11-25 16:23:50 -0800  103)             of readpara_init() has changed to be the same as for cmaes_init().
97ada5f3 (Michael Sherman    2014-11-25 16:23:50 -0800  104)   14/11/25  fix warnings from Microsoft C compiler (sherm)
ffdc1025 (Nikolaus Hansen    2014-12-10 19:49:29 +0100  105) =======
a7adbb3e (Nikolaus Hansen    2014-10-19 15:16:39 +0200  106)             of readpara_init() has changed to be the same as for cmaes_init(). 
23d8ffc1 (Christopher Dembia 2014-11-26 14:38:32 -0800  107)   14/11/26: renamed exported symbols so they begin with a cmaes_prefix.
ffdc1025 (Nikolaus Hansen    2014-12-10 19:49:29 +0100  108) >>>>>>> bd5aeb2a21fbc437f1b322610022f7cb1ba6a9db
```

adamkewley:
Sorry, i went on to working on other things without checking travis: I'm on holiday today/tomorrow, so might not have the time to change the grep in the next few days but can look into it myself. Unless it's a quick thing you can throw on this PR as a review change.

The tabs aren't from the warning patch, but are some other upstream changes that have been pulled in. I guess blacklisting upstream from the grep is an alright bandaid for us, but I might also additionally do a `whitespace-cleanup` on upstream because there's a fair amount of dangling whitespace etc. in the sources.

adamkewley:
That change might sort it

adamkewley:
@sherm1 what are we doing with this code, in particular? Copying c-maes then modifying our own clone, or pulling in an unmodified version of a third-party library? 

If we're pulling a library without modifying it, then I'll have to make *another* PR with the comment fix, get that merged, update this PR, get that merged. If we're keeping our own fork, then I can modify it here and it's done.

I ask because dependency management is something that I've seen kill a few larger software projects. It tends to exchange the time spent coding/maintaining a massive in-tree codebase (the "Linux" approach) with the time spent hopping between projects to get changes merged (the "why on earth does `npm install` pull 7831 dependencies" approach). I've spent several hours hopping around on this--frankly--extremely boring change.

sherm1:
Hi, Adam -- first I want to say I very much appreciate your help getting this sorted out!

I'm not sure I completely understand the question you're asking. Simbody's 3rd party strategy has always been to copy 3rd party code into Simbody's repo -- not necessarily a good idea but does have the advantage of allowing local fixes and making the repo self contained.

For this cmaes fix, ideally we would just copy in the current version of cmaes that incorporates your upstream change and use it unchanged. But I don't know whether there have been local changes to cmaes in Simbody's version, so it does seem safer to merge it in as you did here. (Maybe @chrisdembia knows?) My comment in the review was just that I saw unresolved merge arrows that I assumed were due to the current PR. If I misinterpreted that and they were already there, I don't object to leaving them though it would be tidier to clean them up now since they likely weren't ever noticed before.

Please let me know if that wasn't an actual answer to the question you were asking!

chrisdembia:
> it does seem safer to merge it in as you did here. (Maybe @chrisdembia knows?)

This is the confusing bit...the merge conflict is taken verbatim from the master branch of the upstream `c-cmaes` repository.

chrisdembia:
...likely introduced by me 😬

sherm1:
Looks like @adamkewley just fixed it upstream: https://github.com/CMA-ES/c-cmaes/pull/26

adamkewley:
Yeah, I threw it upstream. Luckily, the maintainer's very quick to respond to PRs, so it's not a huge hold up for this - I'll copy the (fixed) upstream into this PR so that we can roll this out without having to do something smelly like edit our in-tree code (tomorrow: busy right now).

@sherm1 for your answer, its good. What I'm venting (above) is that there is an not-insignificant frustration associated with managing dependencies when the dependency tree of a project deepens - especially if the "dependencies" are also in-tree and linted+reviewed as such (rather than pulled in as a black-box). From my perspective, to fix something in OpenSim, I can potentially end up recursing into 2 or 3 other projects, each requiring a PR to pull the update, and each PR having some time overhead associated with it (turnaround, discussion, review, etc. etc.).

Unfortunately, there isn't a magic bullet for dealing with that frustration, apart from trying to either minimize dependency use (something I'm trying to do on my more recent work, after having my fingers burnt on earlier projects), or just accept that the lead time between a change in a "leaf" dependency and a change at the root can end up growing with depth. No idea if c-cmaes is suitable for interning (rather than copying), though.

adamkewley:
Alright, I copied over the updated `cmaes.c` from 26cc7818 (upstream master), which removed the merge conflict. This changeset is a bit bigger than the warning fix, because upstream has also updated some warning messages etc. 

The overall changeset is looking innocent enough, with the main functional changes being that upstream seems to have fixed an descriptor leak (there's an additional `fclose` in there), and the `szCat` function has been replaced with the updated code for printing the warning/error messages (which now doesn't produce a compiler warning in gcc).

Issue #688, opened 2020-04-26
Ubuntu 16.04 simbody-visualizer ExamplePendulum error


freeyyc:
Ubuntu 16.04
builded successful

when run ./ExamplePendulum. get a Error:
yyc@yyc-HP-Z4-G4-Workstation:~/simbody-source/build$ ./ExamplePendulum
Reactions @M: ~[~[~[24.7279,30.7279,-0.2],~[-33.76,33.56,-12]] ~[~[6.36396,9.36396,0],~[-16.88,16.78,-6]] ~[~[6.66134e-16,3,0],~[-7.92,5.94,-3]] ~[~[-1.18329e-30,-7.88861e-31,-4.44089e-16],~[0,0,0]] ~[~[-1.32625e-30,1.31634e-30,1.23252e-15],~[1.77636e-17,1.23252e-15,-1.03042e-30]]]
Reactions @F: ~[~[~[-24.7279,-30.7279,0.2],~[33.76,-33.56,12]] ~[~[-6.36396,-9.36396,-0],~[16.88,-16.78,6]] ~[~[-6.66134e-16,-3,-0],~[7.92,-5.94,3]] ~[~[1.18329e-30,7.88861e-31,4.44089e-16],~[-0,-0,-0]] ~[~[1.32625e-30,-1.31634e-30,-1.23252e-15],~[-1.77636e-17,-1.23252e-15,1.03042e-30]]]
norm of difference: 0
Pin mobilizer reaction forces:
FB_G=~[~[2.12132,5.12132,0.0707107],~[-16.88,16.78,-6]] ~[~[0,0,-5.94],~[-7.92,5.94,-3]]
Constraint reaction forces (should be the same):
FC_G=~[~[2.12132,5.12132,0.0707107],~[-16.88,16.78,-6]] ~[~[-2.66454e-15,3.10862e-15,-5.94],~[-7.92,5.94,-3]]
Hit ENTER to run a short simulation ...simbody-visualizer listenerThread: unrecoverable error:
SimTK Exception thrown at simbody-visualizer.cpp:2048:
  Error detected by Simbody method simbody-visualizer: An attempt to read() 1 bytes from pipe 3 failed with errno=4 (Interrupted system call).
  (Required condition 'retval!=-1' was not met.)

[...]

PR #691, opened 2020-05-29
simple retry loop for ReadDataFromPipe

jjblum:
@sherm1 , for issue #688

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/691)

### DIFF
--- a/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
+++ b/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
@@ -2038,6 +2038,7 @@ class ReadingInterrupted : public std::exception {};
 
 // Read a particular number of bytes from srcPipe to the given buffer.
 // This will hang until the expected number of bytes has been received.
+// Retries read calls up to 5 times if a call is interrupted.
 // Throws ReadingInterrupted if the srcPipe is closed.
 static void readDataFromPipe(int srcPipe, unsigned char* buffer, int bytes) {
     int totalRead = 0;
--- a/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
+++ b/Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
@@ -2042,11 +2042,11 @@ class ReadingInterrupted : public std::exception {};
 static void readDataFromPipe(int srcPipe, unsigned char* buffer, int bytes) {
     int totalRead = 0;
     while (totalRead < bytes) {
-        auto retval = READ(srcPipe, buffer + totalRead, bytes - totalRead);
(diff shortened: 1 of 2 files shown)


Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
```diff
@@ -2043,8 +2043,12 @@ static void readDataFromPipe(int srcPipe, unsigned char* buffer, int bytes) {
     int totalRead = 0;
     while (totalRead < bytes) {
         auto retval = READ(srcPipe, buffer + totalRead, bytes - totalRead);
+        int attempts = 0;
+        while (retval == -1 && errno == 4 && ++attempts < 5) {
```
chrisdembia, line 2047: Would it be better to use the error name here? I think errno == 4 is `EINTR`. I wonder if `errno == EINTR` would work.

https://man7.org/linux/man-pages/man3/errno.3.html:

> The error numbers that correspond to each symbolic name vary across
       UNIX systems, and even across different architectures on Linux.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/errno-doserrno-sys-errlist-and-sys-nerr?view=vs-2019

Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
```diff
@@ -2043,8 +2043,12 @@ static void readDataFromPipe(int srcPipe, unsigned char* buffer, int bytes) {
...
+          retval = READ(srcPipe, buffer + totalRead, bytes - totalRead);
```
chrisdembia, line 2048: Nitpick: we might be able to avoid duplicating this line using a `do-while` loop:

```cpp
int attempts = 0;
int retval = -1;
do {
    retval = READ(srcPipe, buffer + totalRead, bytes - totalRead);
} while (retval == -1 && errno == 4 && ++attempts < 5);
```

Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
jjblum: agree, it works, pushed this change

Simbody/Visualizer/simbody-visualizer/simbody-visualizer.cpp
jjblum: agree, pushed this change

chrisdembia:
@sherm1 can I merge this?

jjblum:
@sherm1 I added an extra line to the comment. I wasn't sure how to word it - I can change it if you have something in particular in mind.

PR #704, opened 2020-09-10
Fix TLS issues in ParallelExecutor

MadCatX:
Current implementation of `ParallelExecutor` is allowed to switch between singlethreaded and multithreaded operation based on the size of the problem. This causes issues in `Parallel2DExecutor`. `Parallel2DExecutor` splits its calculation into multiple substeps which call `ParallelExecutor::execute()`. However, the associated task and its TLS is initialized only once at the beginning of the calculation. If `ParallelExecutor` decides to switch from ST to MT (or vice versa) mode in the middle `Parallel2DExecutor`'s calculation, the associated task may try to access invalid TLS.

First patch in the PR removes the problem-size-based logic and uses only the specified maximum number of threads to decide whether to run in ST or MT mode. This seems to have unmasked a potential race condition in `CalcForcesNonParallelTask`. This task does not use TLS and therefore is not thread-safe. `GeneralForcesSubsystem` does not limit the maximum number of threads to 1 when `CalcForcesNonParallelTask` is used, which leads to data races. This was not an issue before because `ParallelExecutor` ran in ST mode for `times` value of 1. With the size-based logic removed, we need to set the number of threads to 1 manually for thread-unsafe tasks.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/704)

### DIFF
--- a/SimTKcommon/src/ParallelExecutor.cpp
+++ b/SimTKcommon/src/ParallelExecutor.cpp
@@ -70,7 +70,7 @@ ParallelExecutorImpl* ParallelExecutorImpl::clone() const {
     return new ParallelExecutorImpl(numMaxThreads);
 }
 void ParallelExecutorImpl::execute(ParallelExecutor::Task& task, int times) {
-  if (min(times, numMaxThreads) == 1) {
+  if (numMaxThreads < 2) {
       //(1) NON-PARALLEL CASE:
       // Nothing is actually going to get done in parallel, so we might as well
       // just execute the task directly and save the threading overhead.
--- a/SimTKcommon/tests/TestNoMovingTasks.cpp
+++ b/SimTKcommon/tests/TestNoMovingTasks.cpp
@@ -1,0 +1,99 @@
+/* -------------------------------------------------------------------------- *
+ *                       Simbody(tm): SimTKcommon                             *
+ * -------------------------------------------------------------------------- *
+ * This is part of the SimTK biosimulation toolkit originating from           *
+ * Simbios, the NIH National Center for Physics-Based Simulation of           *
+ * Biological Structures at Stanford, funded under the NIH Roadmap for        *
+ * Medical Research, grant U54 GM072970. See https://simtk.org/home/simbody.  *
+ *                                                                            *
+ * Portions copyright (c) 2008-20 Stanford University and the Authors.        *
+ * Authors: Michal Malý                                                       *
+ * Contributors:                                                              *
+ *                                                                            *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may    *
+ * not use this file except in compliance with the License. You may obtain a  *
+ * copy of the License at http://www.apache.org/licenses/LICENSE-2.0.         *
+ *                                                                            *
+ * Unless required by applicable law or agreed to in writing, software        *
+ * distributed under the License is distributed on an "AS IS" BASIS,          *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.   *
+ * See the License for the specific language governing permissions and        *
+ * limitations under the License.                                             *
+ * -------------------------------------------------------------------------- */
+
+
+/* -------------------------------------------------------------------------- *
+ * Current ParallelExecutor design relies on TLS to store task data. If       *
+ * an initialized task moves between threads, the TLS will point to invalid   *
+ * data. Test that the task division logic does not move ParallelTasks        *
+ * between threads.                                                           *
+ * -------------------------------------------------------------------------- */
+
+#include "SimTKcommon.h"
+
+#include <iostream>
+#include <thread>
+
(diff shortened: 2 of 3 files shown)


sherm1 (approved):
Hi, Michal. Sorry for the delay in getting to look at this -- thanks for doing it! One thing I would like to see is an additional test case that would have revealed the problem before and can verify correct operation now. I'm not sure if that's reasonable -- can you think of a way to do that? There are regression tests TestParallelExecutor TestParallel2DExecutor and TestParallelForces. Please have a look at those to see if there is a way to add a meaningful test in one or more of those. Regards, Sherm

The Ubuntu failure looks like a flake but we'll need to retry.

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/704#-:-MHiC5rR0gJg_k6kYJiP:b-pu1cvt)* status: all files reviewed, 1 unresolved discussion (waiting on @MadCatX)

___
*[Simbody/src/GeneralForceSubsystem.cpp, line 681 at r1](https://reviewable.io/reviews/simbody/simbody/704#-MHiCw4GALAyBLJmPJCu:-MHiCw4HAZuL5f9UMf59:b-5dxbn) ([raw file](https://github.com/simbody/simbody/blob/d0d9b49d96c22d195b1bf9656d86a71ba0906afb/Simbody/src/GeneralForceSubsystem.cpp#L681)):*
> ```cpp
>         else {
>             calcForcesTask = new CalcForcesNonParallelTask();
>             calcForcesExecutor = new ParallelExecutor(1); // NonParallelTask is not thread-safe
> ```

nit: please keep lines to 80 characters

MadCatX:
I fixed the line-length limit and added a test case for unintended task migration between threads.

samuelflores:
Hi Sherm,

It looks like this passed all its tests. It would be nice if it could be reviewed now? We are chasing down seg faults and this fixes at least one of them. 

Thanks

Sam

sherm1 (approved):
Sorry Sam, lost track of this one. Thanks for adding the tests Michal. I am having trouble understanding how they test for the original problem. Can you add some comments to the test case explaining what's being tested? Also, are you sure these tests fail if you remove the new fix?

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/704#-:-MJ2iXiQEV02_1g4OOTM:b-lyyy3j)* status: all files reviewed, 3 unresolved discussions (waiting on @MadCatX)

___
*[SimTKcommon/tests/TestNoMovingTasks.cpp, line 10 at r3](https://reviewable.io/reviews/simbody/simbody/704#-MJ2h4WX6QDoWDWe4wIz:-MJ2h4WX6QDoWDWe4wJ-:b-kq17bw) ([raw file](https://github.com/simbody/simbody/blob/25ca7ea32e5d43b91f92a2f4cc895137942ab6d5/SimTKcommon/tests/TestNoMovingTasks.cpp#L10)):*
> ```cpp
>  *                                                                            *
>  * Portions copyright (c) 2008-12 Stanford University and the Authors.        *
>  * Authors: Peter Eastman                                                     *
> ```

nit: since this is a new file, copyright should be 2020 and the author should be you.
___
*[SimTKcommon/tests/TestNoMovingTasks.cpp, line 66 at r3](https://reviewable.io/reviews/simbody/simbody/704#-MJ2hjmx3Bgncy1Ju50B:-MJ2hjmx3Bgncy1Ju50C:b-w0mf4y) ([raw file](https://github.com/simbody/simbody/blob/25ca7ea32e5d43b91f92a2f4cc895137942ab6d5/SimTKcommon/tests/TestNoMovingTasks.cpp#L66)):*
> ```cpp
> 
> void testFixedMultiple()
> {
> ```

nit: "{" should be on previous line for consistency
___
*[SimTKcommon/tests/TestNoMovingTasks.cpp, line 76 at r3](https://reviewable.io/reviews/simbody/simbody/704#-MJ2hpJgEjjZP6vlXsoe:-MJ2hpJgEjjZP6vlXsof:b-w0mf4y) ([raw file](https://github.com/simbody/simbody/blob/25ca7ea32e5d43b91f92a2f4cc895137942ab6d5/SimTKcommon/tests/TestNoMovingTasks.cpp#L76)):*
> ```cpp
> 
> void testFixedSingle()
> {
> ```

nit: "{" should be on previous line for consistency

MadCatX:
The issue it fixes was a bit more of a corner case than I originally thought but the test does reproduce it and fails without the fix applied. I am sorry about the coding style mismatch, habits die hard.

sherm1 (approved):
Great, thanks! :lgtm:

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/704#-:-MJ8RB613M4PLC57BwGJ:b-napfio)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #705, opened 2020-09-10
Replace loop with explicit formula

MadCatX:
Iterative calculation of the number of bins in `Parallel2DExecutor` can be replaced by an explicit formula.

Test code:
```c++
#include <cmath>
#include <iostream>

int main()
{
        for (int nCpus = 2; nCpus < 513; nCpus++) {
                int levels = 1;
                while (1 << levels < nCpus)
                        levels++;
                levels++;
                int bins = 1 << levels;

                int levelsNew = std::ceil(std::log2(nCpus)) + 1;
                int binsNew = 1 << levelsNew;
                if (bins != binsNew)
                        throw std::runtime_error(std::string("MISMATCH ") + std::to_string(bins) + " " + std::to_string(binsNew));
                std::cout << nCpus << "(" << levels << ", " << bins << ") - " << "(" << levelsNew << ", " << binsNew << ")" << std::endl;
        }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/705)

### DIFF
--- a/SimTKcommon/src/Parallel2DExecutor.cpp
+++ b/SimTKcommon/src/Parallel2DExecutor.cpp
@@ -24,6 +24,7 @@
 #include "Parallel2DExecutorImpl.h"
 #include "SimTKcommon/internal/ParallelExecutor.h"
 #include <utility>
+#include <cmath>
 
 using std::pair;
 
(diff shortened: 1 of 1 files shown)


adamkewley:
This PR has been marked for closure because it has become stale. If there are no major objections, and a willingness to update/review/maintain the PR, by 1st October 2026 (2 weeks) then it will be closed.

PR #714, opened 2021-03-17
Hardened SimTK::Spline_ against thread races

adamkewley:
This is something that I noticed when quickly scanning through **potential** causes of [this](https://github.com/opensim-org/opensim-core/issues/2958) OpenSim issue. It does not necessarily fix that particular issue (I haven't had the time to fully investigate that), but I figured it would be a good idea to PR this in case it's not known to upstream. 

I may have misunderstood the assumptions of this class, or the usage semantics of simbody as a whole (e.g. if there's a higher-level convention in place that explains when/where thread safety is reasonable) - not a problem if you think this is an unnecessarily pedantic PR.

---

(commit notes):

- Because a spline does not mutate its data, SimTK::Spline_
  internally uses a shareable reference-counted SplineImpl to
  hold its data. This makes copying a Spline_ cheap (just increment
  the counter, rather than copying the data)

- However, the reference count was a standard (non-atomic) int. This
  opens up the possibility of a race when multiple threads concurrently
  copy a Spline_ object. Example: if two threads concurrently increment
  the reference counter then the result could be +1 or +2, depending on
  timing - this could result in a double-destruction event when each
  thread is calling Spline's destructor

- Typical C++ developers are likely to expect that copy-construction is
  threadsafe, so the not-threadsafe behavior or SimTK::Spline_ may be
  surprising to downstream users. Worse, thread-safety may be expected
  from an object that indirectly composes a SimTK::Spline_ in its
  (potentially, large) hierarchy. e.g. a user tries to copy-construct
  large model objects that contain SimTK::Spline_s from multiple threads

- This patch lifts the (intrusive) reference counter in SplineImpl into
  a shared_ptr in Spline_. The semantics of a shared_ptr (cheap aliased
  copies, reference counting, single destruction) closely match the original
  semantics, but shared_ptr is designed to be threadsafe

- see: https://en.cppreference.com/w/cpp/memory/shared_ptr

- Excerpt:

    > All member functions (including copy constructor and copy assignment)
    > can be called by multiple threads on different instances of shared_ptr
    > without additional synchronization even if these instances are copies
    > and share ownership of the same object.

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/714)

### DIFF
--- a/SimTKmath/Geometry/include/simmath/internal/Spline.h
+++ b/SimTKmath/Geometry/include/simmath/internal/Spline.h
@@ -29,6 +29,7 @@
 #include "simmath/internal/GCVSPLUtil.h"
 
 #include <limits>
+#include <memory>
 
 namespace SimTK {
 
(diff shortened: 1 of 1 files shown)


tgeijten:
Looks solid to me!

sherm1 (commented):
+@sherm1 for review

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/714#-:-MW-nCOxBPDBI73XKEQN:blr65vz)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sherm1)

PR #732, opened 2022-05-10
Update continuous_integration.yml

aymanhab:
Use VisualStudio 2022 Generator

---
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/732)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -33,7 +33,7 @@ jobs:
           chdir $env:GITHUB_WORKSPACE\\build
           # Set the CXXFLAGS environment variable to turn warnings into errors.
           # TODO: Can remove /WX when we use that in CMakeLists.txt.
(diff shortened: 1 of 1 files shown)


aymanhab:
@sherm1 CI was failing on windows while trying to build with VisualStudio 2019 on windows-latest platform which uses VisualStudio 2022. This PR upgrades the build command to use VisualStudio 2022 generator, not sure if there would be downstream effects of that change to simbody users. I have another PR that uses the older platorm  "windows-2019" which is what we use for OpenSim anyway. Let me know if you have a preference (upgrade generator vs. downgrade platform) so we can have a successful build of simbody on ci. Thanks.

sherm1 (approved):
Oh, sorry Ayman I merged the previous PR before seeing this one. I prefer using the most recent version that works because it is hard for users to get old versions of VisualStudio once the new ones come out. So I'll merge this one too!

Thanks and regards, Sherm

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/732#-:-N1k9U1u-POlnkLjMrTh:b-dgjw7b)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @aymanhab)

PR #745, opened 2022-09-28
Corrected formatting in ExponentialSpringForce.cpp

fcanderson:
Indents were automatically added, I think by Visual Studio, for inline methods in class ExponentialSpringForceImpl.  As a consequence, many lines were extended beyond the 80 character limit.  In addition, the automated formatting resulted in some awkwardly placed indentations and new lines.

I kept the indents for the inline methods but went through and made sure lines did not extend beyond the 80 char limit.  I also tried to make the code more readable by altering indentations and new-line placements in a number of cases, and I tightened up a few comments.

Only formatting and a few non-user-facing comments are changed in this PR.  No substantive changes were made to the code. The code runs the same.

### DIFF
--- a/Simbody/src/ExponentialSpringForce.cpp
+++ b/Simbody/src/ExponentialSpringForce.cpp
@@ -141,16 +141,16 @@ public:
         this->params = params;
     }
 
-    //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


fcanderson:
Thank you, Sherm!

PR #759, opened 2023-08-28
Removed trailing whitespace in String.h and Xml.h. Changed sprintf() to snprintf().

fcanderson:
Removal of trailing whitespace precedes planned changes in String.h and possibly Xml.h.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/759)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/String.h
+++ b/SimTKcommon/include/SimTKcommon/internal/String.h
@@ -118,22 +118,22 @@ Inf, -Inf). **/
 /*@{*/
 /** Format an int as a printable %String. **/
 explicit String(int i, const char* fmt="%d")
-{   const int n=32; char buf[n]; snprintf(buf, n, fmt,i); (*this)=buf; }
+{   const int n=32; char buf[n]; snprintf(buf,n,fmt,i); (*this)=buf; }
 /** Format a long as a printable %String. **/
 explicit String(long i, const char* fmt="%ld")
-{   const int n=64; char buf[n]; snprintf(buf, n, fmt,i); (*this)=buf; }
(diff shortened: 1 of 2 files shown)


fcanderson:
Regarding the failure of the continuous-integration test on Mac...

It seems the culprit is the use of deprecated ```sprintf()``` instead of ```snprintf()```.  I don't know why removing trailing whitespace would result in the emergence of this error.  I imagine there are instances of sprintf() elsewhere in the Simbody code.  Has some compiler directive been corrupted because a trailing space has been removed?  Hmmmmm.  This bridge must have been crossed already.

A simple enough fix is to transition from ```sprintf()``` to ```snprintf()```.  There are only 6 instances of ```sprintf()``` in String.h.

sherm1:
The Mac build has been whining about that for some time -- you didn't cause it. But if you want to fix that while you're in here, please do!

sherm1 (approved):
:lgtm: we can merge this with or without the Mac fix -- LMK

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/759#-:-NcyT8Ha9rJXQXLmunJv:b-zbh7k9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fcanderson)

fcanderson:
I just pushed the changes from sprintf() to snprintf() in String.h.

fcanderson:
Mac build still fails, but this time the culprit is not String.h but Exception.h.

I will go through Exception.h (and other culprits if needed) to fix this sprintf() issue.  But that is probably better done in another PR so this PR doesn't spiral into something big.  If you agree, go ahead and merge this PR.  Thanks!

sherm1:
OK, thanks Clay - merged.

PR #762, opened 2023-09-02
Precision set to LosslessNumDigitsReal in SimTK::String::String(const T&)

fcanderson:
The templatized constructor ```SimTK::String::String(const T& value)``` can be used to convert a value of type ```T``` to a ```String``` provided class ```T``` has implemented ```operator<<()```.

Previously, conversion was done in ```String::String(const T& value)``` using the default precision of ```std::ostringstream```, which is typically ~6.  While an output precision of 6 is sufficient for many applications, it is insufficient when lossless conversion is needed, such as during serialization.

To achieve lossless conversion, the precision is now explicitly set to ```SimTK::LosslessNumDigitsReal```, which is typically ~20.  This change in precision only applies to the local ```ostringstream``` in ```String::String(const T&)```; other instances of ```ostringstream``` in SimTK are unaffected.

This change will affect ```SimTK::Xml::setValueAs<T>(const T& value)```, as this method calls ```String::String(const T&)```. Now, XML elements representing doubles will have  up to ~20 significant digits instead of ~6, rendering class ```SimTK::Xml``` a viable framework for lossless serialization.

The Doxygen compliant comments for ```String::String(const T&)``` have been revised to reflect the change in precision.

Note that a nice feature would be to add a ```precision``` argument that would allow a non-default number of significant digits to be specified:
```
SimTK::String::String(const T& value, int precision=SimTK::LosslessNumDigitsReal)
```
However, in the absence of a clear need for variable precision, the current solution, which avoids a change in the API, is adopted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/762)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/String.h
+++ b/SimTKcommon/include/SimTKcommon/internal/String.h
@@ -27,6 +27,7 @@
 #include "SimTKcommon/internal/common.h"
 #include "SimTKcommon/internal/ExceptionMacros.h"
 
+#include <iomanip>
 #include <cstdio>
 #include <string>
 #include <limits>
--- a/SimTKcommon/include/SimTKcommon/internal/String.h
+++ b/SimTKcommon/include/SimTKcommon/internal/String.h
@@ -41,6 +41,8 @@
 
 namespace SimTK {
 
+extern SimTK_SimTKCOMMON_EXPORT const int LosslessNumDigitsReal;
+
 template <class N> class negator;
 template <class R> class conjugate;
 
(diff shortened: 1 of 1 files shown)


sherm1 (approved):
:lgtm:, thanks Clay

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/762#-:-NdRwjnFD6GgggSqhnI6:b-f70pf1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @fcanderson)

PR #763, opened 2023-09-05
Added access to State indices for ExponentialSpringForce

fcanderson:
Class ExponentialSpringForce has 4 states:
- Static coefficient of friction (discrete)
- Kinetic coefficient of friction (discrete)
- Sliding state (auto-update discrete)
- Elastic anchor point (auto-update discrete)

A 'get' method for the State index of each of the above states has been added:
- ```ExponentialSpringForce::getMuStaticStateIndex()```
- ```ExponentialSpringForce::getMuKineticStateIndex()```
- ```ExponentialSpringForce::getSlidingStateIndex()```
- ```ExponentialSpringForce::getAnchorPointStateIndex()```

A typical user will not need to call these methods.

Access to the indices is provided so that modeling frameworks written on top of Simbody, such as OpenSim, can get the values of the states in a generic way.  For example, a discrete state value can be obtained by calling the following method if its state index is know: 
```
SimTK::Subsystem::getDiscreteVariable(const State& s, DiscreteVariableIndex index)
```
Tests for these methods have been added to the unit test for class ExponentialSpringForce (see TestExponentialSpring.cpp). Documentation for these new methods has also been added.

A few trailing white spaces were also removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/763)

### DIFF
--- a/Simbody/include/simbody/internal/ExponentialSpringForce.h
+++ b/Simbody/include/simbody/internal/ExponentialSpringForce.h
@@ -116,7 +116,7 @@ publication:
 
 The current class makes several improvements to that contact model, most
 notably including 1) the ability to rotate and translate the contact plane and
-2) the ability to specify both a static and a kinetic coefficient of friction. 
+2) the ability to specify both a static and a kinetic coefficient of friction.
 The computational details of the contact model implemented by this class
 follow below.
 
@@ -405,6 +405,13 @@ public:
     at Stage::Topology is streamlined. */
     const ExponentialSpringParameters& getParameters() const;
 
+    /** Get the State index of the static coefficient of friction (μₛ) for
+    this exponential spring. Typically, you will not need to use this index.
+    It is exposed to allow low-level, generic access to the state by other
+    modeling frameworks writen on top of Simbody, such as OpenSim.
+    @return State index for μₛ. */
+    DiscreteVariableIndex getMuStaticStateIndex() const;
+
     /** Set the static coefficient of friction (μₛ) for this exponential
     spring. The value of μₛ is held in the System's State object. Unlike the
     parameters managed by ExponentialSpringParameters, μₛ can be set at any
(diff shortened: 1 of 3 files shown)


sherm1 (approved):
Looks good -- I have a suggestion for reorganizing the documentation to eliminate duplication, PTAL.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/763#-:-NeFsAfuCo0Wz6H8GWhM:b-awabq9)* status: all files reviewed, 2 unresolved discussions (waiting on @fcanderson)

___
*[`Simbody/include/simbody/internal/ExponentialSpringForce.h` line 408 at r1](https://reviewable.io/reviews/simbody/simbody/763#-NeFpGbq7r1eGvx1djBF:-NeFpGbq7r1eGvx1djBG:b2cvp1p) ([raw file](https://github.com/simbody/simbody/blob/99a7a8efbb1387d274a5476dbdc162a73612369c/Simbody/include/simbody/internal/ExponentialSpringForce.h#L408)):*
> ```objc
>     const ExponentialSpringParameters& getParameters() const;
> 
>     /** Get the State index of the static coefficient of friction (μₛ) for
> ```

BTW consider moving these more-obscure expert user functions to the end of the list rather than having them show up in the middle with the commonly-used functions. Also, consider beginning each comment with "(Advanced)", e.g.
`/** (Advanced) Get the State index ...`

That will help ordinary users know not to bother looking at these.

As an alternative (probably better) you could use an `@name` grouping to collect all of these index-access methods so they appear together in Doxygen and share common documentation rather than repeating the same text for each function.
___
*[`Simbody/include/simbody/internal/ExponentialSpringForce.h` line 430 at r1](https://reviewable.io/reviews/simbody/simbody/763#-NeFpv8A4nykpnoOlxBT:-NeFpv8A4nykpnoOlxBU:bgo2228) ([raw file](https://github.com/simbody/simbody/blob/99a7a8efbb1387d274a5476dbdc162a73612369c/Simbody/include/simbody/internal/ExponentialSpringForce.h#L430)):*
> ```objc
>     Real getMuStatic(const State& state) const;
> 
>     /** Get the State index assinged to the kinetic coefficient of friction
> ```

typo: assinged (same typo several places below)

Also, consider using the same wording style for similar functions to avoid implying that there is something different here. In the style of the MuStatic method, this would begin "Get the State index of the kinetic ..." (or use "assigned to" in both places)

Better, avoid duplication altogether using `@name` group

fcanderson:
Thanks for the comments, Sherm.  I'll explore the @name group to unify the documentation and correct the typo you found.

fcanderson:
The revamped documentation is ready to for a follow-up review. It's much improved I think.  Thanks for the suggestions.

PR #767, opened 2023-11-03
Fix MultibodyGraphMaker declaration order to be C++20 compatible (#766)

### LINKED ISSUE #766: Cannot compile in C++20 if a compilation unit uses `MultibodyGraphMaker.h`
(requires verification, but this is my written understanding of the issue)

The reason why is because, in C++20, `std::vector<T>` has many `constexpr` operations, and these enforce that `T` must be a complete type *when `std::vector<T>` is declared* (rather than when any of its member functions are called, which is the <C++20 behavior).

What this means is that code that forward-declares a class, e.g. like here in `MultiBodyGraphMaker`:

- https://github.com/simbody/simbody/blob/ebf2aa54385318cd7c9e460398fc9e329a8150cd/SimTKmath/include/simmath/MultibodyGraphMaker.h#L155

Can't instantiate a `std::vector<T>`, e.g. here:

- https://github.com/simbody/simbody/blob/ebf2aa54385318cd7c9e460398fc9e329a8150cd/SimTKmath/include/simmath/MultibodyGraphMaker.h#L385

Until `T` is fully defined, e.g. here:

- https://github.com/simbody/simbody/blob/ebf2aa54385318cd7c9e460398fc9e329a8150cd/SimTKmath/include/simmath/MultibodyGraphMaker.h#L401

It worked in previous versions of C++ because the constructor function for the containing class (`MultbodyGraphMaker`) was what was instantiating the data (by calling a member constructor). In previous C++es, it's at *that* point that the compiler may need to know the size of `T`.

The C++20 change is that `std::vector<T>` can now be used at compile-time:

- https://en.cppreference.com/w/cpp/container/vector/vector

But this feature forces the compiler to verify that the declared type is compile-time-constructible at class-declaration time.

adamkewley:
Fixes #766 

Steps taken

- Removed previous forward declarations
- Copied+pasted inner class definitions into the top of `MultibodyGraphMaker`
  - This should be behaviorally the same as the previous approach
  - But ensures they are fully defined *before* the `std::vector` members of `MultibodyGraphMaker`
- Indented definitions
- Removed large banners above the internal definitions (because they are now inside another class that already has its own banner?)

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/767)

### DIFF
--- a/SimTKmath/include/simmath/MultibodyGraphMaker.h
+++ b/SimTKmath/include/simmath/MultibodyGraphMaker.h
@@ -152,11 +152,227 @@ do not have a level.
 class SimTK_SIMMATH_EXPORT MultibodyGraphMaker {
 public:
     // Local classes.
-    class Body;
-    class Joint;
-    class JointType;
-    class Mobilizer;
-    class LoopConstraint;
+
+    /** Local class that collects information about bodies. **/
+    class Body {
+    public:
+        explicit Body(const std::string&    name, 
+                        double              mass, 
+                        bool                mustBeBaseBody,
+                        void*               userRef) 
+        :   name(name), mass(mass), mustBeBaseBody(mustBeBaseBody), 
+            userRef(userRef), level(-1), mobilizer(-1), master(-1) {}
+
+        void forgetGraph(MultibodyGraphMaker& graph);
+        int getNumFragments() const {return 1 + getNumSlaves();}
+        int getNumSlaves() const {return (int)slaves.size();}
+        int getNumJoints() const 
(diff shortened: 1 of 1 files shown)


sherm1 (approved):
Awesome! Thanks for the fix, Adam. :lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/767#-:-NiQARatB4t4FmShdDfc:b-6n6394)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #770, opened 2023-11-22
Replace std::iterator with std::iterator_traits-compatible equivalent

adamkewley:
Minor nit, but it would help with my downstream project.

In [opensim-creator](https://github.com/ComputationalBiomechanicsLab/opensim-creator) I'm compiling with `/W4`, `-pedantic`, C++20, with all warnings upgraded to errors etc. All of the simbody headers get through except for this one, which requires me to explicitly ignore the warning:

- https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/2ba887a1a3a828a81284c08fd18f9e9924d3c65f/src/OpenSimCreator/CMakeLists.txt#L343

It's because `std::iterator` was deprecated, and the modern convention is for `std::iterator_traits<UserType>` to query the object for the relevant typedefs. I went through `std::iterator` and ensured that the typedefs added by this patch are identical to what `std::iterator` used to add.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/770)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/Xml.h
+++ b/SimTKcommon/include/SimTKcommon/internal/Xml.h
@@ -646,9 +646,14 @@ inline std::ostream& operator<<(std::ostream& o, const  Attribute& attr) {
 //------------------------------------------------------------------------------
 /** This is a bidirectional iterator suitable for moving forward or backward
 within a list of Attributes within an Element, for writable access. **/
-class SimTK_SimTKCOMMON_EXPORT attribute_iterator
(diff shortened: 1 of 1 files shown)


sherm1 (approved):
:lgtm: -- thanks, Adam!

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/770#-:-NjsHbGP2MPIO6qzSSrq:b36ra04)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #771, opened 2023-11-22
Ensure simbody links transitive DLLs (e.g. libgfortran) 

adamkewley:
The simbody cmake install doesn't explicitly link all transitive DLLs that it may depend on, which means that downstream projects (e.g. [OpenSim Core](https://github.com/opensim-org/opensim-core), [OpenSim Creator](https://github.com/ComputationalBiomechanicsLab/opensim-creator)) have to manually copy DLL files etc. around when packaging downstream applications that depend on the libraries.

---

**Downstream Example**: [here](https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/2ba887a1a3a828a81284c08fd18f9e9924d3c65f/src/OpenSimCreator/CMakeLists.txt#L329)

In the example, I create fake `IMPORTED` targets for the `OpenSimCreator` library. The reason I do this, rather than manually copy the DLLs, is so that all downstream executables (e.g. the application, application packaging, unit tests, benchmarks, etc.) can generically copy runtime libraries next to the executable. This is necessary when (e.g.) running things from Visual Studio ([example usage](https://github.com/ComputationalBiomechanicsLab/opensim-creator/blob/2ba887a1a3a828a81284c08fd18f9e9924d3c65f/apps/osc/CMakeLists.txt#L62))

I'm aware that I could copy+dump all binaries into one directory, but the utility of doing it this way is that changing the dependencies of OSC automatically propagates any runtime file dependencies downstream - regardless of filesystem location.

---

This PR fixes the issue by telling CMake about the transitive dependency. **I'm not sure if I've done it correctly (I'm assuming these are all needed by BLAS and LAPACK)**. The reason why IMPLIB is `blas` is because the affected DLLs have no IMPLIB, but CMake's linking regieme gets a bit confused by that.

Automatic reproduction of the issue (Windows, git bash):

```bash
 # on Windows, in `git bash` terminal
 
 # configure simbody
cmake -S . -B simbody-build -DCMAKE_INSTALL_PREFIX=${PWD}/simbody-install

 # build + install simbody (locally)
cmake --build simbody-build -j$(nproc) --target install

# clean-generate downstream project
rm -rf downstream downstream-build
mkdir downstream
cat << EOF > downstream/main.cpp
#include <Simbody.h>

int main(char**,int)
{
    SimTK::Body b;  // ensure linkage
    return 0;
}
EOF

cat << EOF > downstream/CMakeLists.txt
cmake_minimum_required(VERSION 3.15)

project(example VERSION 0.0.1 LANGUAGES CXX)

find_package(Simbody)

add_executable(example main.cpp)
target_link_libraries(example SimTKcommon SimTKmath SimTKsimbody)

if (WIN32)
    add_custom_command(
        TARGET example
        PRE_BUILD
        COMMAND \${CMAKE_COMMAND} -E copy_if_different \$<TARGET_RUNTIME_DLLS:example> $<TARGET_FILE_DIR:example>
        COMMAND_EXPAND_LISTS
    )
endif()
EOF

# configure downstream project
cmake -S downstream -B downstream-build -DCMAKE_PREFIX_PATH=${PWD}/simbody-install

# build downstream project (works)
cmake --build downstream-build

# list libraries used by downstream project (should contain everything?)
ls downstream-build/Debug
```

Before this PR:

```bash
$ bash repro.sh
# build messages
SimTKcommon_d.dll  SimTKmath_d.dll  SimTKsimbody_d.dll  example.exe  example.pdb  libblas.dll  liblapack.dll
```

After this PR:

```bash
$ bash repro.sh
# build messages
SimTKcommon_d.dll  SimTKmath_d.dll  SimTKsimbody_d.dll  example.exe  example.pdb  libblas.dll  libgcc_s_sjlj-1.dll  libgfortran-3.dll  liblapack.dll  libquadmath-0.dll
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/771)

### DIFF
--- a/Platform/CMakeLists.txt
+++ b/Platform/CMakeLists.txt
@@ -195,7 +195,15 @@ endforeach()
 # All the platform libraries (including quadmath, etc.) still must be on the
 # PATH when running executables using Simbody.
 if(WIN32 AND NOT BUILD_USING_OTHER_LAPACK)
-    # transitive dependencies for blas/lapack
+    # declare transitive dependencies for blas/lapack (see simbody/simbody#771)
+    #
+    # the transitive dependencies don't, themselves, have IMPLIBs, because
+    # they have already been linked by blas/lapack. However, cmake can become
+    # confused when it sees a library with no IMPLIB, so we include blas's
+    # here to satisfy cmake
+    #
+    # (if you link to blas or lapack, you're going to link to the blas IMPLIB
+    # anyway, so it makes no difference)
     add_library(libgcc_s_sjlj-1 SHARED IMPORTED)
     set_target_properties(libgcc_s_sjlj-1 PROPERTIES
         IMPORTED_IMPLIB "${LIB_ABI_DIR}/libblas.lib"
@@ -218,7 +226,7 @@ if(WIN32 AND NOT BUILD_USING_OTHER_LAPACK)
     set_target_properties(blas PROPERTIES
         IMPORTED_IMPLIB "${LIB_ABI_DIR}/libblas.lib"
         )
-    # link transitive dependencies
+    # link to transitive dependencies
     target_link_libraries(blas INTERFACE libgcc_s_sjlj-1 libgfortran-3 libquadmath-0)
     
     add_library(lapack SHARED IMPORTED GLOBAL)
--- a/Platform/CMakeLists.txt
+++ b/Platform/CMakeLists.txt
@@ -217,9 +217,9 @@ if(WIN32 AND NOT BUILD_USING_OTHER_LAPACK)
     add_library(blas SHARED IMPORTED GLOBAL)
     set_target_properties(blas PROPERTIES
         IMPORTED_IMPLIB "${LIB_ABI_DIR}/libblas.lib"
-        # link transitive dependencies
-        INTERFACE_LINK_LIBRARIES libgcc_s_sjlj-1 libgfortran-3 libquadmath-0
         )
+    # link transitive dependencies
+    target_link_libraries(blas INTERFACE libgcc_s_sjlj-1 libgfortran-3 libquadmath-0)
     
     add_library(lapack SHARED IMPORTED GLOBAL)
     set_target_properties(lapack PROPERTIES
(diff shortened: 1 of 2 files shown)


sherm1 (approved):
I'm not enough of a CMake wizard to know if this is the best approach, but it seems very reasonable to want all the Simbody-supplied DLLs to be listed. :lgtm: pending addition of a short comment, please see below.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/771#-:-NjsLGc25qESTcOLuC2D:bdsdt1j)* status: all files reviewed, 1 unresolved discussion (waiting on @adamkewley)

___
*[`cmake/SimbodyConfig.cmake.in` line 83 at r1](https://reviewable.io/reviews/simbody/simbody/771#-NjsNVzw8oFFSiGWJOnr:-NjsNVzw8oFFSiGWJOns:bps18o1) ([raw file](https://github.com/simbody/simbody/blob/c984470690330ae84746c483452ebf7215e5cf18/cmake/SimbodyConfig.cmake.in#L83)):*
> ```
>     add_library(libgcc_s_sjlj-1 SHARED IMPORTED)
>     set_target_properties(libgcc_s_sjlj-1 PROPERTIES
>         IMPORTED_IMPLIB "@PACKAGE_CMAKE_INSTALL_LIBDIR@/libblas.lib"
> ```

minor: the mismatched IMPLIB and LOCATION will likely mystify later readers once the PR discussion is forgotten. Would be good to have a comment here explaining why you did it this way.

adamkewley:
I'll try and sort it, thanks @sherm1 !

sherm1 (approved):
Perfect, thanks Adam.

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/771#-:-Nk6n4S-91YUue94vzH_:b-7x2dbw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #773, opened 2023-12-11
Fix O(N^2) vector growth when reading mesh files

adamkewley:
Hi,

Random performance improvement related to loading mesh files.

I have a test suite that loads every available OpenSim model file, finalizes them, assembles them, and generates 3D decorations for them. It's a system-level sanity check method, but it is also one of the slowest tests in my suite. This is because it involves a lot of IO (esp. mesh loading).

If I profile the test suite in VS, I see all kinds of optimization candidates (e.g. 45 % of it's in OpenSim, with a lot of that being related to copying objects etc.) but one hotspot stood out as particularly suspicious:

image

The reason why is because I wouldn't expect 17.5 % of performance to be spent on a reallocations in that part of the code. When I investigated it, it turns out that it's because the vector that's used to read floats only grows one element at a time, which will have O(N^2) performance (allocate N-1, copy N-1, append Nth, etc. etc.) - this PR hopefully minimally patches the behavior

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/773)

### DIFF
--- a/SimTKcommon/BigMatrix/include/SimTKcommon/internal/BigMatrix.h
+++ b/SimTKcommon/BigMatrix/include/SimTKcommon/internal/BigMatrix.h
@@ -1131,13 +1131,17 @@ std::istream& readVectorFromStreamHelper
 
         // Now read in an element of type T.
         // The extractor T::operator>>() will ignore leading white space.
-        if (!isFixedSize)
-            out.resizeKeep(out.size()+1); // grow by one (default consructed)
+        if (!isFixedSize && nextIndex >= out.size())
+        {
+            // grow output vector geometrically (resized to actual size after loop)
+            out.resizeKeep(out.size() == 0 ? 1 : 2*out.size());
+        }
         in >> out[nextIndex]; if (in.fail()) break;
         ++nextIndex;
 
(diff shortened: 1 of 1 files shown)


sherm1 (approved):
Thanks, Adam. :lgtm:
Before merging, can you verify that the change had the desired performance effect?

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/773#-:-NlOYRakCtM6qVm2N-4U:b-cu7izr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

adamkewley:
Yep - can do, give me a little time!

adamkewley:
Screenshot 2023-12-11 202252

This is the new profile when running the same test against a custom build with this PR. The time gains are roughly in line with the percentage differences (around 10-15 %)

sherm1 (approved):
Nice! Thanks.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/773#-:-NlTbDns1CWMq4irUh1t:b-v6rzh8)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

PR #774, opened 2024-01-18
Fix RigidBodyNode API to allow returning locally-computed spatial vecs

adamkewley:
Fixes a warning that is emitted by newer versions of clang/gcc, e.g.:

```text
/home/adam/opensim-core/dependencies/simbody/Simbody/src/RigidBodyNode_LoneParticle.cpp: In member function ‘virtual const SpatialVec& RBNodeLoneParticle::getHCol(const SBTreePositionCache&, int) const’:
/home/adam/opensim-core/dependencies/simbody/Simbody/src/RigidBodyNode_LoneParticle.cpp:414:12: warning: function returns address of local variable [-Wreturn-local-addr]
  414 |     return col;
      |            ^~~
/home/adam/opensim-core/dependencies/simbody/Simbody/src/RigidBodyNode_LoneParticle.cpp:410:19: note: declared here
  410 |     Mat<2,3,Vec3> H = Mat<2,3,Vec3>::getAs(&pc.storageForH[2*uIndex]);
```

The API design for `RigidBodyNode` returns a `const SpatialVec&`, but that is incompatible with computing the `SpatialVec` on-the-fly (locally). To get around that, I modified the API to return a `SpatialVec` value. This is "more expensive" than just returning a reference (i.e. returning 1 integer vs. 6 floats) but, in practice, that kind of overhead rarely matters (because downstream users of the result are likely to want to immediately load the reference into registers to copy it elsewhere or use it in a calculation).

One other concern might be that changing an API from using a const-reference to a value may cause code like the following to be undefined:

```c++
const SpatialVec& v = node.getHCol(...);   // uh-oh: reference to a now-value-based temporary?
```

But that kind of downstream code will still be well-defined due to reference lifetime extension:

- https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary

However, code that (e.g.) takes references to subobjects of the `SpatialVec`, or passes the reference via other function calls, will be broken by this change. Whether this matters (because the top-level API appears to return `SpatialVec` values, and the original UB was specifically isolated to `LoneParticle`) - no idea

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/774)

### DIFF
--- a/Simbody/src/RigidBodyNode.h
+++ b/Simbody/src/RigidBodyNode.h
@@ -587,10 +587,10 @@ virtual void multiplyByMPass2Inward(
   { SimTK_THROW2(Exception::UnimplementedVirtualMethod, "RigidBodeNode", "multiplyByMPass2Inward"); }
 
 // Note that this requires columns of H to be packed like SpatialVec.
-virtual const SpatialVec& getHCol(const SBTreePositionCache&, int j) const 
+virtual SpatialVec getHCol(const SBTreePositionCache&, int j) const
 {SimTK_THROW2(Exception::UnimplementedVirtualMethod, "RigidBodeNode", "getHCol");}
 
-virtual const SpatialVec& getH_FMCol(const SBTreePositionCache&, int j) const 
+virtual SpatialVec getH_FMCol(const SBTreePositionCache&, int j) const
 {SimTK_THROW2(Exception::UnimplementedVirtualMethod, "RigidBodeNode", "getH_FMCol");}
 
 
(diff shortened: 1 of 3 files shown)


sherm1 (commented):
Thanks for looking at this, Adam! I was able to reproduce and took a look at the code. Both getHCol() and getH_FMCol() are buggy misguided attempts to generate the constant 6x3 H matrix with columns`(0,0,0,1,0,0)ᵀ (0,0,0,0,1,0)ᵀ (0,0,0,0,0,1)ᵀ`. The position cache already has storage for H which should have been initialized to that value in realizeInstance() with all the other initializations. Then those two functions should simply return a reference to one of those columns.

If you're willing, please take a look at fixing it that way. Since the LoneParticle class exists entirely for performance reasons I would like to avoid returning the 6x1 copy rather just a reference into the existing cache space (as all the other mobilizer types do!). 

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/774#-:-NoSjacREtOxspI8PnF6:byh1l5q)* status: 0 of 3 files reviewed, all discussions resolved

sherm1:
PR #786 replaces this PR using the implementation discussed above.

Issue #778, opened 2024-03-02
Specifying XML output precision with an optional argument causes a bug for variables of type bool


fcanderson:
With ```precision``` as an optional argument in the following methods (see #776) 

```
String::String(const T& t, int precision = DefaultOutputPrecision)
Xml::Element(const String& tagWord, const T& value, int precision = DefaultOutputPrecision)
Xml::setValueAs(const T& value, int precision = DefaultOutputPrecision)
```

variables of type ```bool``` are not being handled properly. In particular, calling either ```Xml::Element::Element()``` or ```Xml::Element::setValueAs<bool>()``` for a value of type bool was causing the specialized constructor ```SimTK::String::String(bool b)``` to be sidestepped. As a consequence, ```bool``` values are being written to an XML document as integers ```0``` or ```1``` as opposed to ```false``` or ```true```.

fcanderson:
#779 fixes this bug.

PR #776, opened 2024-02-06
XML String Precision

fcanderson:
While there are various constructors available in class ```SimTK::String``` that allowed for specification of output format, the templatized constructor for class ```String``` (i.e., ```String::String(const T& t)```) was fixed at the output precision set by ```std::ostream```. As a consequence, any ```String``` conversion for a type ```T``` for which there exists no matching constructor was fixed to an output precision of ```6```.

So that output precision (i.e., the number of significant figures used to represent a value) can be specified for such types, this commit adds an optional ```precision``` argument to the templatized ```String``` constructor and to the methods in class ```SimTK::Xml``` that call this templatized ```String``` constructor. In particular, the following methods now include the optional ```precision``` argument:
```
SimTK::String::String<T>(const T& t, int precision = String::DefaultOutputPrecision)

SimTK::Xml::Element::Element<T>(const String& tagWord, const T& value, int precision = String::DefaultOutputPrecision)

SimTK::Xml::Element::setValueAs<T>(const T& value, int precision = String::DefaultOutputPrecision)
```
Because the ```precision``` argument is optional, no API adjustments are needed in existing code that calls the above methods.

The static constant ```String::DefaultOutputPrecision``` was added to support this new functionality. Its value is chosen to be ```6``` to be consistent with the default precision of class ```std::ostream``` and so that previous behavior would be maintained.

The primary benefit of this commit is that a broad range of numerical value types in SimTK (e.g., ```Vec3```, ```Vec4```, etc.) can be serialized via class ```SimTK::Xml``` with a specified number of significant figures.  The changes revert default output precision to what it was prior to #762 but also make it possible to serialize ```SimTK::State```, among other quantities, in lossless fashion.

User facing Doxygen comments were updated and a subunit test (```testOutputPrecision```) was added to ```testXml.cpp```.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/776)

PR #779, opened 2024-03-02
XML String Precision Bug Fix

fcanderson:
This PR fixes issue #778.  The fix is to remove ```precision``` as an optional argument in the following methods, returning their signatures to their form prior to #776, and to add three new methods with ```precision``` as a non-optional argument.

---- Methods in #776
```
String::String(const T& t, int precision = String::DefaultOutputPrecision)
Xml::Element(const String& tagWord, const T& value, int precision = String::DefaultOutputPrecision)
Xml::setValueAs(const T& value, int precision = String::DefaultOutputPrecision)
```
---- Methods prior to #776 (original signatures)
```
String::String(const T& t)
Xml::Element(const String& tagWord, const T& value)
Xml::setValueAs(const T& value)
```

---- New methods (non-optional precision argument)
```
String::String(const T& t, int precision)
Xml::Element(const String& tagWord, const T& value, int precision)
Xml::setValueAs(const T& value, int precision)
```

With this approach, variables of type ```bool``` are handled properly.  When a demand for a number of significant figures other than the default ```std::ostream``` value (i.e., 6) arise, users should simply call one of the new added methods with the desired precision. An added benefit of this approach is that the constant ```String::DefaultOutputPrecision```, which was added in #776, could be (and was) removed.

The subunit test, ```Xml.cpp::TestXml::testOutputPrecision``` has been updated to account for the above changes. In particular, all references to ```String::DefaultOuputPrecision``` were removed, and ```SimTK_TEST()``` calls for variables of type bool were added.

User-facing documentation was also updated.

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/String.h
+++ b/SimTKcommon/include/SimTKcommon/internal/String.h
@@ -165,26 +165,14 @@ explicit String(std::complex<double> r, const char* fmt="%.17g")
 or "0" cast the bool to an int first. **/
 explicit String(bool b) : std::string(b?"true":"false") { }
 
-// ------------
-// Developer Notes Re: String::DefaultOutputPrecision
-// - This constant was added February 2024 to provide a consistent default
-//   value of the precision argument in the following methods:
-//   * String::String(const T& t, int precision = DefaultOutputPrecision)
-//   * Xml::Element::Element(const String& tagWord, const T& value,
-//         int precision = String::DefaultOutputPrecision)
-//   * Xml::Element::setValueAs(const T& value,
-//         int precision = String::DefaultOutputPrecision)
-// - It could be altered to be any integer without causing an error, but some
-//   care should be taken before making a change.
-//   * Simbody users (e.g., OpenSim devs) may have written software around a
-//     default output precision of 6 (e.g., model files, GUI displays).
-//   * The default output precision should be greater than or equal to 1 and
-//     less than or equal to SimTK::LosslessNumDigitsReal.
-// ------------
-/** The default output precision of the templatized string constructor is 6,
-which corresponds to the default output precision of std::ostream objects.
-See String::String(const T& t, int precision). */
-static const int DefaultOutputPrecision{6};
+/** For any type T for which there is no matching constructor, this templatized
+constructor will format an object of type T into a %String provided that there
+is either an available specialization or (as a last resort) a stream insertion
+operator<<() available for type T; a *runtime* error is thrown if neither is
+available.
+@param t %Value to be converted to a %String. **/
+template <class T> inline explicit
+String(const T& t);
 
 /** For any type T for which there is no matching constructor, this templatized
 constructor will format an object of type T into a %String provided that there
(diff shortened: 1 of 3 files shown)


sherm1 (approved):
Thanks, Clay -- this seems like a good solution. :lgtm:
Consider changing the date from Feb -> Mar (comment below)

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/779#-:-Ns5cXQH1tsxW1DUxJuF:b-1799qa)* status: all files reviewed, 1 unresolved discussion (waiting on @fcanderson)

___
*[`SimTKcommon/tests/TestXml.cpp` line 327 at r1](https://reviewable.io/reviews/simbody/simbody/779#-Ns5c2Ur2uflAiZHm1U8:-Ns5c2Ur2uflAiZHm1U9:b-thxwn5) ([raw file](https://github.com/simbody/simbody/blob/1f64af4d02a5220d6133bd15c8c5e2a6b47c366f/SimTKcommon/tests/TestXml.cpp#L327)):*
> ```cpp
> }
> 
> // February 2024
> ```

nit: March now

fcanderson (commented):
Thanks for the speedy review Sherm!
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/779#-:-Ns5n-Sg2Momz5wFmgQA:b-4qmz2)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @sherm1)

___
*[`SimTKcommon/tests/TestXml.cpp` line 327 at r1](https://reviewable.io/reviews/simbody/simbody/779#-Ns5c2Ur2uflAiZHm1U8:-Ns5mqhV39riEiUAejm0:bb7f2mf) ([raw file](https://github.com/simbody/simbody/blob/1f64af4d02a5220d6133bd15c8c5e2a6b47c366f/SimTKcommon/tests/TestXml.cpp#L327)):*

nit: March now
</blockquote></details>

done

sherm1 (approved):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/779)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @fcanderson)

Giopestri01:
Excuse me, I got this error when I try to execute the tests on Simbody:
 C:\Windows\System32>"C:\Users\acer\Desktop\SW_THESIS\simbody-master\build\Release\TestXml.exe"
Path of this executable: 'c:\Users\acer\Desktop\SW_THESIS\simbody-master\build\Release\TestXml.exe'
Executable directory: 'c:\Users\acer\Desktop\SW_THESIS\simbody-master\build\Release\'
Current working directory: 'c:\Windows\System32\'
Starting test TestXml ...
  testOutputPrecision ...
  testOutputPrecision done. real/CPU ms: 0.4 / 0.0
Test failed due to exception: SimTK Exception thrown at TestXml.cpp:396:
  Internal bug detected: Test condition failed.
  (Assertion '(output == expected[p])' failed).
  Please file an Issue at https://github.com/simbody/simbody/issues.
  Include the above information and anything else needed to reproduce the problem.

TestXml done. real/CPU ms: 1.1 / 0.0 
Is it related to this issues by any chance?

sherm1:
Thanks, @Giopestri01. What do you think, @fcanderson ?

fcanderson:
Yes, thank you, @Giopestri01.  I am looking into it.

Giopestri01:
[like]  Giovanni Pestrichella reacted to your message:
________________________________

Giopestri01:
[like]  Giovanni Pestrichella reacted to your message:
________________________________

Giopestri01:
Actually, I do not have a lot of experience with cpp and software developing. I need Simbody to make RTosim work (real-time OpenSim), I compile cmake to build the binaries from the simbody-master and then I build the projects using Release configuration. Then, when I built RUN-TESTS, the only falied test has been the xml test, with the error I showed above. This is the onlt things I have done. I tried multiple time, but I got the same error

Giopestri01:
I'm using Visual Studio 17 2022, cmake 3.29, Windows 11

fcanderson:
@Giopestri01 ok.  thx for the info.
I may have an idea about what's happening.  OpenSim does an automatic install of Simbody, but it is not currently set to build the Simbody master.  If you are also building the Simbody master, it is possible that those two builds are stepping on each other.

Let's continue this dialog via email if that is ok with you.  Email me at fcandergit@gmail.com.

fcanderson:
@Giopestri01 Thank you for your assistance in identifying the cause of this test failure.
I created a new issue reporting the details of the failure.  See #780.  I will implement a fix, which I anticipate will be fairly straightforward.  Thanks again for your help.

Giopestri01:
Thank you too @fcanderson, it has been a pleasure for me!

PR #783, opened 2024-03-09
Add methods for computing the Jacobian `PV = [P; V]` to SimbodyMatterSubsystem

### LINKED ISSUE #782: Expose `SimbodyMatterSubsystemRep::multiplyByPVATranspose` to `SimbodyMatterSubsystem`
I am currently working a new method for enforcing kinematic constraints in OpenSim Moco (opensim-org/opensim-core#3587). This method relies requires a projection of the coordinate states in a manner analoguous to Simbody's approach where a vector is multiplied by components of the kinematic constraint Jacobian `G = [P; V; A]`. 

Like Simbody (or as far as I understand what Simbody is doing), I project the coordinate values and coordinate speeds separately. I use the `SimbodyMatterSubsystem::multiplyByPqTranspose` for the `q` projections, but there is currently no similar method available for projecting only the coordinate speeds. `SimbodyMatterSubsystem::multiplyByGTranspose` is available, but includes slots for acceleration constraints, which I don't need.

I recently found `SimbodyMatterSubsystemRep::multiplyByPVATranspose`, which seems perfect for this (I can configure it to give me a `[P; V]` transpose multiplication), but it is not available in `SimbodyMatterSubsystem`.

@sherm1 what do you think about exposing this functionality? Seems pretty straightforward at first glance. I can start working on this locally and, if it makes sense, open a PR soon.

nickbianco:
Fixes #782.

This PR adds the desired method `multiplyByPVTranspose()` to `SimbodyMatterSubsystem`. It seemed logical to add the additional methods `multiplyByPV()` (both signatures), `calcBiasForMultiplyByPV()`, `calcPV()`, and `calcPVTranspose()` similar to the methods for `Pq` and `G`. I've added relevant tests to `TestConstraints.cpp`. 

Minor: I fixed a couple of typos in comment strings and also updated the comment for `multiplyByPVATranspose()` to more clearly describe the size requirement for the input argument `lambda`.

I'll leave a comment or two in the changes below with some other questions I add.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/783)

### DIFF
--- a/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
+++ b/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
@@ -1535,10 +1535,10 @@ functions with zero input to determine the bias term for use in
 multiplyByG(). Body quantities and generalized quantities are supplied to each 
 of the m active constraints' (constant time) error methods to calculate
 <pre>
-   pverr(t,q,u;ulike)=G*ulike - c(t,q)    (holonomic) 
+   pverr(t,q,u;ulike)=G*ulike - c(t,q)    (holonomic)
 or aerr(t,q,u;ulike)=G*ulike - b(t,q,u)   (nonholonomic or acceleration-only)
 </pre>
-with ulike=0, giving the bias term in O(m) time. 
+with ulike=0, giving the bias term in O(m) time.
 
 If you want the acceleration-level bias terms b for all the constraints, even
 if they are holonomic, use calcBiasForAccelerationConstraints(). **/
@@ -1679,7 +1679,7 @@ void multiplyByGTranspose(const State&  state,
     
 /** This O(nm) operator explicitly calculates the n X m transpose of the 
 acceleration-level constraint Jacobian G = [P;V;A] which appears in the system 
-equations of motion. This method generates ~G columnwise use the constraint 
+equations of motion. This method generates ~G columnwise using the constraint
 force generating methods which map constraint multipliers to constraint forces.
 To within numerical error, this should be identical to the transpose of
 the matrix returned by calcG() which uses a different method. Consider using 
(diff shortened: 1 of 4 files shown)


Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
```diff
@@ -1895,6 +1895,146 @@ void calcP(const State& state, Matrix& P) const;
...
+    constraints.
+
+@pre \a state realized to Velocity stage
+@par Implementation
+This method uses either velocity- or acceleration- level constraint error
+functions with zero input to determine the bias term for use in
+multiplyByPV(). Body quantities and generalized quantities are supplied to each
+of the m active constraints' (constant time) error methods to calculate
+<pre>
+   pverr(t,q,u;ulike)=P*ulike - c(t,q)     (holonomic)
```
nickbianco, line 1961: I based this equation on the docs above for `calcBiasForMultiplyByG` above, but per the Simbody theory manual it seems to me that this should be either
```
pverr(t,q;qlike)=Pq*qlike - c(t,q)
```
or
```
pverr(t,q,u;ulike)=P*ulike - b_p(t,q,u)
```

Or are they interchangeable if `ulike = 0` (or `qlike = 0`)?

sherm1 (approved):
+@sherm1 for review

:lgtm_strong: really great to see this enhancement! Thanks, Nick. Just a few minor requests below then ready to merge.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/783#-:-NseWuly68DFladCH-g5:b-uc3iny)* status: all files reviewed, 3 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/include/simbody/internal/SimbodyMatterSubsystem.h` line 1961 at r1](https://reviewable.io/reviews/simbody/simbody/783#gh-1518454796:-NsePnzK4PvJxfplKPxP:b1vyvt6) ([raw file](https://github.com/simbody/simbody/blob/c31397c863d8a52140388558763800371b4b707b/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h#L1961)):*

I based this equation on the docs above for `calcBiasForMultiplyByG` above, but per the Simbody theory manual it seems to me that this should be either
```
pverr(t,q;qlike)=Pq*qlike - c(t,q)
```
or
```
pverr(t,q,u;ulike)=P*ulike - b_p(t,q,u)
```

Or are they interchangeable if `ulike = 0` (or `qlike = 0`)?
</blockquote></details>

The difference is whether we use the first time derivative of the holonomic (position) constraints (pverr) or the second time derivative (paerr). If we use the pverr equation, then we want just the time dependent "bias" term c(t,q); if we use paerr then we need the more complex bias term b_p(t,q,u). Because the pverr calculation is simpler (and still involves P), it is preferable to use that if you can for computing produces of P. It's more uniform to do everything with the acceleration-level constraint errors paerr, vaerr, aerr but it isn't the most efficient. In addition to being simpler, pverr has no velocity dependence so we don't need to realize velocities to calculate it.

(Pq is just PN⁻¹ and ulike=N⁻¹qlike so Pq*qlike = P*ulike so that part doesn't mean much, but definitely c ≠ b_p.)

So I think this is right as you've written it.
___
*[`Simbody/include/simbody/internal/SimbodyMatterSubsystem.h` line 1975 at r2](https://reviewable.io/reviews/simbody/simbody/783#-NseS6_B851bhfT5qZNP:-NseS6_B851bhfT5qZNQ:b1jqk0u) ([raw file](https://github.com/simbody/simbody/blob/770c5108988dc5067658ab0c56472a60213b3473/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h#L1975)):*
> ```objc
> constraint Jacobian G) which appears in the system equations of motion. Consider
> using the multiplyByPV() method instead of this one, which forms the matrix-
> vector product PV*v in O(mp+mv+n) time without explicitly forming PV.
> ```

nit: `PV*u` to keep the same notation (in retrospect I wish I had used v for generalized velocities but we're stuck with u in Simbody for historical reasons).
___
*[`Simbody/include/simbody/internal/SimbodyMatterSubsystem.h` line 1979 at r2](https://reviewable.io/reviews/simbody/simbody/783#-NseSiNe6Q_QZM2rwzaZ:-NseSiNfBjD7uafpRup6:b-85lh07) ([raw file](https://github.com/simbody/simbody/blob/770c5108988dc5067658ab0c56472a60213b3473/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h#L1979)):*
> ```objc
> @par Implementation
> This method generates PV columnwise using repeated calls to multiplyByPV(),
> which makes use of the constraint error methods to perform a PV*v product
> ```

nit: `PV*u`
___
*[`Simbody/src/SimbodyMatterSubsystemRep.h` line 815 at r2](https://reviewable.io/reviews/simbody/simbody/783#-NseVbjrA9bvfHP5mwSk:-NseVbjrA9bvfHP5mwSl:b-r8fum7) ([raw file](https://github.com/simbody/simbody/blob/770c5108988dc5067658ab0c56472a60213b3473/Simbody/src/SimbodyMatterSubsystemRep.h#L815)):*
> ```objc
>     //    fu = [ ~P ~V ~A ] * lambda
>     // with all or a subset of P,V,A included. The multiplier-like vector
>     // must have length equal to the total number of kinematic constraints for
> ```

nit: number of _active_ kinematic constraints

nickbianco (commented):
Thanks @sherm1! Let me know if the failing tests are an issue.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/783#-:-NsigcW53deuGwVC6mIN:bsrpb90)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @sherm1)

___
*[`Simbody/include/simbody/internal/SimbodyMatterSubsystem.h` line 1961 at r1](https://reviewable.io/reviews/simbody/simbody/783#gh-1518454796:-NsiftFmE4i5BLu8kcPi:bovwwqo) ([raw file](https://github.com/simbody/simbody/blob/c31397c863d8a52140388558763800371b4b707b/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h#L1961)):*

The difference is whether we use the first time derivative of the holonomic (position) constraints (pverr) or the second time derivative (paerr). If we use the pverr equation, then we want just the time dependent "bias" term c(t,q); if we use paerr then we need the more complex bias term b_p(t,q,u). Because the pverr calculation is simpler (and still involves P), it is preferable to use that if you can for computing produces of P. It's more uniform to do everything with the acceleration-level constraint errors paerr, vaerr, aerr but it isn't the most efficient. In addition to being simpler, pverr has no velocity dependence so we don't need to realize velocities to calculate it.

(Pq is just PN⁻¹ and ulike=N⁻¹qlike so Pq*qlike = P*ulike so that part doesn't mean much, but definitely c ≠ b_p.)

So I think this is right as you've written it.
</blockquote></details>

Great. Thanks for the clarification!

nickbianco:
Oh, the Ubuntu tests were failing on my previous commit. Not sure why they are hanging in GitHub Actions now.

sherm1 (approved):
I built and ran your branch on Ubuntu 22.04 (Jammy). I got two compiler warnings in LoneParticle (unrelated to your change). All 110 tests passed cleanly. So we're good to go here!

The Ubuntu CI needs something; I'm not sure why it isn't running automatically.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/783#-:-Nsk5S2N7Q3Fp8XYTfjg:b-woxuaj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

PR #786, opened 2024-03-23
Fixed LoneParticle bugs in getHCol() and getH_FMCol()

sherm1:
The LoneParticle mobilizer implementation was incorrectly returning references to expired local variables when asked for columns of H_PB_G or H_FM. 

Since these matrices never change for a LoneParticle (F=P=G and M=B by construction) this PR instead precalculates H_PB_G and H_FM in the position cache (in realizeInstance()) and then just returns the right column when asked. I added a missing test case to TestLoneParticle.cpp that failed with the original implementation and works now.

There is some minor cleanup here also:
- I renamed storageForH to storageForH_PB_G to be more precise.
- I wrapped some long lines in TestLoneParticle.cpp to meet the 80 character standard
- Added a few short comments
- Added a .clang-format configuration file that matches Simbody's style

This replaces PR #774.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/786)

### DIFF
--- a/Simbody/src/RigidBodyNode_LoneParticle.cpp
+++ b/Simbody/src/RigidBodyNode_LoneParticle.cpp
@@ -26,15 +26,17 @@
 #include "MobilizedBodyImpl.h"
 #include "RigidBodyNodeSpec_Translation.h"
 
-/* This is a specialized class used for MobilizedBody::Translation objects that 
-satisfy all of the following requirements:
+/* This is a specialized class used for MobilizedBody::Translation mobilizers
+that satisfy _all_ of the following requirements:
  -  The body has no children.
  -  The body's parent is ground.
  -  The inboard and outboard transforms are both identities.
  -  The body is not reversed.
 
- These assumptions allow lots of routines to be implemented simpler and faster.
- */
+These assumptions allow lots of routines to be implemented simpler and faster.
+If any of them is not met, we use the more general RBNodeTranslate to
+implement the Translation mobilizer.
+*/
 class RBNodeLoneParticle : public RigidBodyNode {
 public:
 RBNodeLoneParticle(const MassProperties& mProps_B,
@@ -148,6 +150,19 @@ void realizeInstance(const SBStateDigest& sbs) const override {
     SBTreeAccelerationCache& ac = sbs.updTreeAccelerationCache();
     Transform& X_FM = toB(pc.bodyJointInParentJointFrame);
     X_FM.updR().setRotationToIdentityMatrix();
+
+    // Precalculate H_FM_F and H_PB_G here since they never change for a
+    // LoneParticle (the rotational part is zero and the translational part
+    // is identity). And since we require F=P=G here, these are the same.
+    Mat<2,3,Vec3>& H_PB_G = // a.k.a. just "H"
+        Mat<2,3,Vec3>::updAs(&pc.storageForH_PB_G[2 * uIndex]);
+    Mat<2, 3, Vec3>& H_FM_F =
+        Mat<2,3,Vec3>::updAs(&pc.storageForH_FM[2 * uIndex]);
+    H_PB_G.setToZero();
+    for (int j = 0; j < 3; ++j)
+      H_PB_G(1, j)[j] = 1;
+    H_FM_F = H_PB_G;
+
     updV_FM(vc)[0] = Vec3(0);
     updV_PB_G(vc)[0] = Vec3(0);
     updVD_PB_G(vc)[0] = Vec3(0);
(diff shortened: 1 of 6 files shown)


sherm1:
@adamkewley can I get you to review this, please? It implements the suggestion I made in your PR #774.

Simbody/src/RigidBodyNode_LoneParticle.cpp
```diff
@@ -148,6 +150,19 @@ void realizeInstance(const SBStateDigest& sbs) const override {
...
+    // LoneParticle (the rotational part is zero and the translational part
+    // is identity). And since we require F=P=G here, these are the same.
+    Mat<2,3,Vec3>& H_PB_G = // a.k.a. just "H"
+        Mat<2,3,Vec3>::updAs(&pc.storageForH_PB_G[2 * uIndex]);
+    Mat<2, 3, Vec3>& H_FM_F =
+        Mat<2,3,Vec3>::updAs(&pc.storageForH_FM[2 * uIndex]);
+    H_PB_G.setToZero();
+    for (int j = 0; j < 3; ++j)
+      H_PB_G(1, j)[j] = 1;
+    H_FM_F = H_PB_G;
```
adamkewley, line 164: `Mat<2,3,Vec3>::updAs(&pc.storageForH_FM[2 * uIndex]) = H_PB_G` if you want to drop a local that's unused for most of this logic

adamkewley (approved):
Thanks for looking into this @sherm1 - I couldn't see any problems, my only comments are drive-bys that don't matter (i.e. just merge this as-is)

sherm1 (commented):
Thanks, Adam! Responses below.

When using Reviewable for reviews (a really great tool, BTW!) there are magic keywords you can use to express the level of strictness of your comments. If you just comment as you did below, the comment is blocking and you have to unblock it after the author responds (fine if that is your intention). For comments where you want to inform the author of a minor problem or just a thought but don't require a re-review afterwards, start your comment with one of these words:
- _minor_: must be fixed, but I don't need to look at it again
- _typo_: obvious
- _nit_: something trivial but annoying, e.g. missing punctuation, bad spacing
- _BTW_: a thought or ignorable suggestion

Also there is a pull-down menu with these same options if you forget the magic keywords. (You'll need that to unblock your blocking comments when satisfied.)

When you feel comfortable that the PR is ready except for minor things you trust the author to fix, you should stamp your comment in the discussion area here with `:lgtm:` :lgtm: to unblock overall.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/786#-:-NtqFVHj3O5fBPAaAiZn:b-81ctj5)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @adamkewley)

___
*[`Simbody/src/RigidBodyNode_LoneParticle.cpp` line 164 at r2](https://reviewable.io/reviews/simbody/simbody/786#gh-1536789672:-NtqE7Qt3K5xxaQUkldL:bk4izc9) ([raw file](https://github.com/simbody/simbody/blob/2e133341c71e5b53eb9e8e5680660e837dbae0b4/Simbody/src/RigidBodyNode_LoneParticle.cpp#L164)):*

`Mat<2,3,Vec3>::updAs(&pc.storageForH_FM[2 * uIndex]) = H_PB_G` if you want to drop a local that's unused for most of this logic
</blockquote></details>

A reasonable alternative, thanks. I like the clarity of getting the ugly accessing over with then using nicer names so I'll leave it this way.
___
*[`Simbody/src/SimbodyTreeState.h` line 675 at r2](https://reviewable.io/reviews/simbody/simbody/786#gh-1536789221:-NtqEVLh-yl63PgvhLQ6:bdyiv9o) ([raw file](https://github.com/simbody/simbody/blob/2e133341c71e5b53eb9e8e5680660e837dbae0b4/Simbody/src/SimbodyTreeState.h#L675)):*

Not relevant for this particular PR, but it might be worth considering encoding the `nx` shape of this data at the type level (e.g. as `std::pair<Vec3, Vec3>`, `struct H { Vec3 F; Vec3 M; Vec3 PB; Vec3 G; }`) so that callers aren't having to stride their accesses (with `2*uIndex`) - unless there's a locality argument for striping going on (which `2*uIndex` would give but `uIndex].second` wouldn't)
</blockquote></details>

Good idea. I added a TODO for now -- I'll need to investigate why I wrote it this way in the first place -- seems like it would have been obviously better to use a SpatialVec (which is two packed Vec3s) rather than individual Vec3s.

For performance reasons (due to memory access patterns) it is generally better to have a "struct of arrays" than "array of structs" so I wouldn't mix H_PB and H_FM here, but making them arrays of SpatialVecs seems much better.

sherm1 (commented):
Merging this now.

> Dismissed @adamkewley from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/786#-:-Nu-QJb11OGUDrmvQLwA:bram0kq)* status: 0 of 6 files reviewed, all discussions resolved (waiting on @adamkewley)

Issue #787, opened 2024-03-25
Efficiently calculating a subset of a system's body and mobility forces


nickbianco:
I am working on a new feature for OpenSim Moco that would allow users to track a reference set of generalized coordinate forces in a trajectory optimization problem. The OpenSim `InverseDynamicsTool` uses `SimbodyMatterSubsystem::calcResidualForce()` to compute generalized coordinate forces by internally making a copy of the model and disabling forces that should be excluded from the anaylsis (usually all forces except external loads, e.g., from ground reaction force data). 

In Moco, it would be preferable to not require a copy of the model since all classes that implement cost function terms already have a reference to the original model. Instead, we ideally should only need to retrieve the set of body and mobility forces needed to compute the generalized forces from inverse dynamics. [`Force::calcForceContribution`](https://github.com/simbody/simbody/blob/bd171213534e524da279d823a786f9c5c143ca2c/Simbody/include/simbody/internal/Force.h#L128) has the desired functionality, but, as noted in the documentation, it requires resizing and zeroing out passed in body and mobility force arrays and is therefore not the best choice for use in optimization. 

## Desired workflow
The desired workflow to compute generalized forces within Moco's trajectory optimization framework would look as follows:
```c++
// Non-modifiable inputs from the trajectory optimization framework.
const State& state = input.state;
const OpenSim::Model& model = input.model;

// Get the generalized accelerations.
model.realizeAcceleration(state);
const Vector& udot = state.getUDot();

// Get a subset of the applied body and mobility forces in the model. 
Vector subsetAppliedMobilityForces = ?
Vector_<SpatialVec> subsetAppliedBodyForces = ?

// Compute the residual mobility forces (i.e., the inverse dynamics generalized forces).
const SimbodyMatterSubsystem& matter = model.getMatterSubsystem();
Vector residualMobilityForces(state.getNU());
matter.calcResidualForceIgnoringConstraints(state, 
        subsetAppliedMobilityForces, subsetAppliedBodyForces, 
        udot, residualMobilityForces);

```

## Proposed method
One potential option is to add a method to `GeneralForceSubsystemRep` to calculate the subset of desired forces based on a list of `ForceIndex`es:
```c++
GeneralForceSubsystemRep::calcForces(const State& state, 
        const Array_<ForceIndex>& forceIndexes, 
        Vector_<SpatialVec>& bodyForces, 
        Vector& mobilityForces) {

    // appropriate asserts/checks here

    Vector_<Vec3> particleForces; // unused
[...]

PR #788, opened 2024-03-28
Add the method `calcForces` to `GeneralForceSubsystem`

nickbianco:
Addresses #787.

## Summary

This changes adds the method `calcForces` to `GeneralForceSubsystem`, which accepts an array of `ForceIndex` values and calculates the sum of each `Force`'s contribution to the system body and mobility forces. `calcForces` provides similar functionality as `Force::calcForceContribution()`, but only requires zeroing the input force vectors once before calculating the contributions from multiple forces, which should be more efficient when repeated calls are necessary.

## Looking for feedback on

1. Method naming. As drafted, I'm using the generic `calcForces`, but something like `calcForceContributions` or `calcSummedForceContributions`  might be more descriptive.
2. A safe and proper way to skip the resizing/zeroing steps if possible, since the intent here is for speed. Resizing is implicitly skipped if the vectors are the correct size, but not zeroing (I think). A flag to the method would be one option, but stylistically avoiding that would probably be preferred. If no solution is optimal then that's also fine since this is already an improvement for my application.
3. Should I leverage `CalcForcesTask` somehow? This would open up the possibility for parallelization, but I don't quite fully understand everything related to caching forces in there yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/788)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -126,7 +126,7 @@ jobs:
   ubuntu:
     name: Ubuntu
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     strategy:
       matrix:
@@ -185,7 +185,7 @@ jobs:
   style:
     name: Style
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
--- a/Simbody/include/simbody/internal/Force.h
+++ b/Simbody/include/simbody/internal/Force.h
@@ -124,7 +124,9 @@ public:
         hint here. 
     @note This method must zero out the passed in arrays, and in most cases
     almost all returned entries will be zero, so this is \e not the most
-    efficent way to calculate forces; use it sparingly. **/
+    efficent way to calculate forces; use it sparingly. For a more efficent way
+    to calculate force contributions from multiple force elements, see
+    GeneralForceSubsystem::calcForces(). **/
     void calcForceContribution(const State&          state,
                                Vector_<SpatialVec>&  bodyForces,
                                Vector_<Vec3>&        particleForces,
--- a/Simbody/include/simbody/internal/Force.h
+++ b/Simbody/include/simbody/internal/Force.h
@@ -126,7 +126,7 @@ public:
     almost all returned entries will be zero, so this is \e not the most
     efficent way to calculate forces; use it sparingly. For a more efficent way
     to calculate force contributions from multiple force elements, see
-    GeneralForceSubsystem::calcForces(). **/
+    GeneralForceSubsystem::calcForceContributionsSum(). **/
     void calcForceContribution(const State&          state,
                                Vector_<SpatialVec>&  bodyForces,
                                Vector_<Vec3>&        particleForces,
--- a/Simbody/include/simbody/internal/GeneralForceSubsystem.h
+++ b/Simbody/include/simbody/internal/GeneralForceSubsystem.h
@@ -107,8 +107,8 @@ public:
     if the given \a state were realized to Dynamics stage. This sizes the given
     arrays if necessary, zeroes them, and then calls each force element's
     calcForce() method which adds its force contributions if any to the
-    appropriate array elements for bodies and mobilities. Note that in general 
-    we have no idea what elements of the system are affected by a force element, 
+    appropriate array elements for bodies and mobilities. Note that in general
+    we have no idea what elements of the system are affected by a force element,
     and in fact that can change based on state and time (consider contact 
     forces, for example). A disabled force element will return all zeroes 
     without invoking calcForce(), since that method may depend on earlier 
@@ -139,7 +139,7 @@ public:
         hint here.  **/
     void calcForceContributionsSum(
         const State& s, const Array_<ForceIndex>& forceIndexes, 
-        Vector_<SpatialVec>& rigidBodyForces, Vector& mobilityForces) const;
+        Vector_<SpatialVec>& bodyForces, Vector& mobilityForces) const;
 
     /** Every Subsystem is owned by a System; a GeneralForceSubsystem expects
     to be owned by a MultibodySystem. This method returns a const reference
--- a/Simbody/include/simbody/internal/GeneralForceSubsystem.h
+++ b/Simbody/include/simbody/internal/GeneralForceSubsystem.h
@@ -102,6 +102,44 @@ public:
     @return Maximum number of threads GeneralForceSubsystem can use for force
     computations**/
     int getNumberOfThreads() const;
+    
+    /** Calculate the sum of forces that would be applied by the force elements 
+    if the given \a state were realized to Dynamics stage. This sizes the given
+    arrays if necessary, zeroes them, and then calls each force element's
+    calcForce() method which adds its force contributions if any to the
+    appropriate array elements for bodies and mobilities. Note that in general 
+    we have no idea what elements of the system are affected by a force element, 
+    and in fact that can change based on state and time (consider contact 
+    forces, for example). A disabled force element will return all zeroes 
+    without invoking calcForce(), since that method may depend on earlier 
+    computations which may not have been performed in that case.
+    @param[in]      state
+        The State containing information to be used by the force elements to
+        calculate the current sum of forces. This must have already been 
+        realized to a high enough stage for each force element to get what it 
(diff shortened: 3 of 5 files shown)


tkuchida:
If `calcForces()` is like `calcForceContribution()` but for a plurality of forces, perhaps then `calcForceContributionSum()` instead so the methods smell the same and are next to each other in Doxygen/autocomplete? 🌹

nickbianco:
@tkuchida good point, it would be good to have these appear next to each other in Doxygen. One tweak on your suggestion: `calcForceContributionsSum` to connote multiple contributions being summed together.

sherm1 (approved):
@tkuchida great to see you out here in Simbody land!

@nickbianco I like this new method. To answer your questions above:
- Tom's idea for the name looks great to me
- My first thought about zeroing would be to leave it as-is and try at the call site to list all the forces you're interested in at one shot. If you must make multiple calls and measure zeroing as an unacceptable cost, you could make an even-more-advanced method that has the same signature but assumes the arrays have been properly sized and zeroed (calcForceContributionSumAdvanced() or some such). Then calcForceContributionSum() would size & zero and then call the advanced method to do the work. But I'd stick with just the current method unless you can determine that the advanced method would make a significant difference.
- I would not attempt to use the force parallelization stuff unless really desperate!

Regarding the PR, it looks great except for one thing -- there is no unit test for the new method. I'd like to see a test showing that it works properly for several forces now and to make sure it continues to work forever. (I always picture some future overconfident teenage hacker messing with things years from now without understanding the big picture -- we need a unit test that will blow up when that happens.)  

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/788#-:-NuBSG0-6aDXlw4dvjIw:b-xpwum)* status: all files reviewed, 4 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/include/simbody/internal/Force.h` line 129 at r2](https://reviewable.io/reviews/simbody/simbody/788#-NuBQ3wrBV9ktrEtlMx9:-NuBQ3wrBV9ktrEtlMxA:bvq0d4s) ([raw file](https://github.com/simbody/simbody/blob/8c36a6b10081547da95edb8acfc6adce536f3a2f/Simbody/include/simbody/internal/Force.h#L129)):*
> ```objc
>     efficent way to calculate forces; use it sparingly. For a more efficent way
>     to calculate force contributions from multiple force elements, see
>     GeneralForceSubsystem::calcForces(). **/
> ```

minor: calcForces() -> calcForceContributionsSum()
___
*[`Simbody/include/simbody/internal/GeneralForceSubsystem.h` line 111 at r2](https://reviewable.io/reviews/simbody/simbody/788#-NuBQe1z8l9H3ly9fYQB:-NuBQe1z8l9H3ly9fYQC:bx9u6o8) ([raw file](https://github.com/simbody/simbody/blob/8c36a6b10081547da95edb8acfc6adce536f3a2f/Simbody/include/simbody/internal/GeneralForceSubsystem.h#L111)):*
> ```objc
>     calcForce() method which adds its force contributions if any to the
>     appropriate array elements for bodies and mobilities. Note that in general 
>     we have no idea what elements of the system are affected by a force element, 
> ```

nit: spurious blank line
___
*[`Simbody/include/simbody/internal/GeneralForceSubsystem.h` line 142 at r2](https://reviewable.io/reviews/simbody/simbody/788#-NuBQNHl3r-XH8L9LEiX:-NuBQNHl3r-XH8L9LEiY:b-uruzf5) ([raw file](https://github.com/simbody/simbody/blob/8c36a6b10081547da95edb8acfc6adce536f3a2f/Simbody/include/simbody/internal/GeneralForceSubsystem.h#L142)):*
> ```objc
>     void calcForceContributionsSum(
>         const State& s, const Array_<ForceIndex>& forceIndexes, 
>         Vector_<SpatialVec>& rigidBodyForces, Vector& mobilityForces) const;
> ```

minor: rigidBodyForces -> bodyForces to match the documentation just above and for consistency with calcForceContribution().
___
*[`Simbody/src/GeneralForceSubsystem.cpp` line 604 at r2](https://reviewable.io/reviews/simbody/simbody/788#-NuBS6Zh5BJQGxPmBCuk:-NuBS6Zh5BJQGxPmBCul:bh5eln9) ([raw file](https://github.com/simbody/simbody/blob/8c36a6b10081547da95edb8acfc6adce536f3a2f/Simbody/src/GeneralForceSubsystem.cpp#L604)):*
> ```cpp
>     void calcForceContributionsSum(const State& s, 
>         const Array_<ForceIndex>& forceIndexes, 
>         Vector_<SpatialVec>& rigidBodyForces, Vector& mobilityForces) const 
> ```

nit: rigidBodyForces -> bodyForces for consistency

nickbianco (commented):
Thanks @sherm1! I've added a unit test for the new method and addressed your comments.

The current method works great as-is for my application, so I think there's no need for an extra advanced method.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/788#-:-NuBnKr-BcoxYKC_MYMz:byfmpzc)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @sherm1)

sherm1 (commented):
Great test, thanks Nick! :lgtm: pending a few nits, then this is ready to merge.

> Reviewed 3 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/788#-:-NuFkA_ECRAt5LEc9mpt:bwp3tmm)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/tests/TestForces.cpp` line 369 at r3](https://reviewable.io/reviews/simbody/simbody/788#-NuFiod-0YSp-2q83JiO:-NuFiod-0YSp-2q83JiP:b-afj1zp) ([raw file](https://github.com/simbody/simbody/blob/08f2511a0f8103cce227b0341e5eeae3799de793/Simbody/tests/TestForces.cpp#L369)):*
> ```cpp
>     for (int i = 0; i < NUM_BODIES; ++i) {
>         MobilizedBody& parent = matter.updMobilizedBody(MobilizedBodyIndex(matter.getNumBodies()-1));
>         MobilizedBody::Gimbal b(parent, Transform(Vec3(0)), body, Transform(Vec3(BOND_LENGTH, 0, 0)));
> ```

nit: per Simbody's CONTRIBUTING.md [style guide](https://github.com/simbody/simbody/blob/master/CONTRIBUTING.md#keep-line-width-to-80-characters), lines should be kept to 80 characters.

I realize existing code in this file violates that convention! But ideally we should get closer over time :)
___
*[`Simbody/tests/TestForces.cpp` line 401 at r3](https://reviewable.io/reviews/simbody/simbody/788#-NuFihX00Cx0UUcE37S_:-NuFihX00Cx0UUcE37Sa:b29fpf4) ([raw file](https://github.com/simbody/simbody/blob/08f2511a0f8103cce227b0341e5eeae3799de793/Simbody/tests/TestForces.cpp#L401)):*
> ```cpp
>     
>     Vector_<SpatialVec> bodyForces(matter.getNumBodies(), SpatialVec(0));
>     Vector_<Vec3> particleForces(0);
> ```

nit: unused
___
*[`Simbody/tests/TestForces.cpp` line 416 at r3](https://reviewable.io/reviews/simbody/simbody/788#-NuFik4EA4kk3GG0OX1F:-NuFik4EA4kk3GG0OX1G:b-cxqn24) ([raw file](https://github.com/simbody/simbody/blob/08f2511a0f8103cce227b0341e5eeae3799de793/Simbody/tests/TestForces.cpp#L416)):*
> ```cpp
>     forceIndexes.push_back(constantTorque.getForceIndex());
>     
>     // The GlobalDamper is disable and should not contribute any forces
> ```

typo: disable -> disabled

nickbianco:
I noitced that the Ubuntu builds and style check jobs in GitHub Actions are hanging because the `ubuntu-18.04` runner is deprecated. As a drive-by for this PR, I've updated the workflow to use the `ubuntu-20.04` runner.

sherm1 (approved):
Perfect -- and thanks for the CI fix!

> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/788#-:-NuGWw-h8JI0MP8OuRu-:b-59rkpy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

PR #792, opened 2024-05-29
Add CMakePresets and begin using in CI

halleysfifthinc:
This PR defines CMake presets (pre-defined build configurations) for common simbody build configs (initial main intended use is in CI).

The primary benefit of presets is that they make it much easier/more convenient to run the same build config that is used in CI, or to define standard build configurations (for users or devs) with different build options or settings (e.g. you can define a preferred compiler, which is not possible from within a CMakeLists.txt file).

The build configuration settings for the new CI build presets are exactly the same* as the previous command line options in the ["continuous-integration" workflow](.github/workflows/continuous_integration).

Two other presets were added: `default` and `dev-default`, intended to define standard build configs for users (people who want to build simbody for some reason, but don't intend to develop simbody), and for developers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/792)


CMakePresets.json
```diff
@@ -0,0 +1,307 @@
...
+            "description": "The right build configuration for most users",
+            "inherits": "ci-base",
+            "cacheVariables": {
+                "BUILD_TESTING": false,
+                "CMAKE_COMPILE_WARNING_AS_ERROR": false
+            },
+            "warnings": {
+                "dev": false,
+                "deprecated": false
+            }
```
halleysfifthinc, line 42: Warnings (CMake and compiler) in CI/dev-default vs default: Builds for regular users (who don't intend to develop/modify Simbody itself) should be quiet (no warnings); only hard errors should be shown. Compiling unfamiliar software is difficult and a wall of irrelevant text doesn't make anything easier. 

This doesn't silence compiler warnings yet; I'm considering adding a "DEVELOP_SIMBODY" CMake variable to use to turn on some dev defaults in a more cross platform way (instead of needing different presets to silence warnings correctly in different compilers).

CMakePresets.json
```diff
@@ -0,0 +1,307 @@
...
+            "hidden": true,
+            "name": "ci-base",
+            "displayName": "CI configuration",
+            "description": "Sets Ninja generator, build and install directory",
+            "generator": "Ninja",
+            "binaryDir": "${sourceDir}/out/build/${presetName}",
+            "installDir": "${sourceDir}/out/install/${presetName}",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release",
+                "BLA_VENDOR": "Generic"
```
halleysfifthinc, line 23: *This is a new-to-simbody configuration option [`BLA_VENDOR=Generic`](https://cmake.org/cmake/help/latest/module/FindBLAS.html#blas-lapack-vendors) added to the base preset (i.e. applies to all presets).

This setting should be functionally irrelevant for CI builds, which only have one BLAS/LAPACK installed, but should be more universal for user's with multiple installed BLAS/LAPACK's. Any manually specified cache variables (e.g. on the command line or via GUI) will override preset settings, so users who wish to use a specific BLAS/LAPACK (either by picking a different/specific BLA_VENDOR, or by setting BUILD_USING_OTHER_LAPACK). Defaulting to Generic BLAS/LAPACK allows more vendor-agnostic linking, such as allowing post-build swapping of the default BLAS/LAPACK's using update-alternatives, which I have tested does work.)

CMakePresets.json
```diff
@@ -0,0 +1,307 @@
...
+            }
+        },
+        {
+            "name": "dev-default",
+            "displayName": "Default build configuration for devs",
+            "description": "The right build configuration for most developers",
+            "inherits": "ci-base",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+                "CMAKE_CXX_FLAGS": "-Wall"
```
halleysfifthinc, line 51: `-Wall` for dev-default may be too opinionated. Happy to remove/change, but it did highlight a number of potential issues (unused variables, etc)

halleysfifthinc (commented):
Some additional context

halleysfifthinc:
I've pushed fixes for the previous build failure, and I will need another Actions run approval. Thanks!

halleysfifthinc:
All green! :tada:  Looking at the artifacts, the Debug artifacts are >.25Gb. If that's not wanted, I can change the artifact upload step to skip uploading Debug artifacts.

halleysfifthinc:
Bump.

sherm1:
Thanks -- I'll try to review this weekend.

sherm1 (approved):
:lgtm: Allen, thanks for doing this.

Q: do we need to update the README documentation so that people know what to do with this?  Or is it intended just for our own CI use and transparent? I for one am unclear how I would use presets in my development workflow. Can you elaborate?

> Reviewed 3 of 27 files at r1, 24 of 24 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/792#-:-O4fEmaTCml_yq0WfaQB:b817p4h)* status: all files reviewed, 2 unresolved discussions (waiting on @halleysfifthinc)

___
*[`CMakePresets.json` line 2 at r2](https://reviewable.io/reviews/simbody/simbody/792#-O4fDvl76xt4dv-Uc2YP:-O4fDvl76xt4dv-Uc2YQ:b5g78mr) ([raw file](https://github.com/simbody/simbody/blob/bb4369413b14d9d64fa5fbf1d7320e81e496da03/CMakePresets.json#L2)):*
> ```JSON
> {
>     "version": 3,
> ```

BTW The CMake documentation [here](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) says preset version 3 goes with CMake 3.21 (if I understood correctly). Below it looks like you are requiring only CMake 3.12. Should we require a higher level?

halleysfifthinc:
> Q: do we need to update the README documentation

I do plan on eventually updating the README for this and other planned build system updates, but I figured I would wait until the end of my planned PRs and update the README once.

The main benefit of presets is standardized build configs (primarily for CI) with less typing. As you can see in the updated CI, the configuration command is shorter and simpler since the build options are set in the preset instead of on the command line. Another benefits is that build/install directories are set by the preset, so you can quickly switch between presets (e.g. between release and debug build types), and each build is in a separate directory and won't be overwritten. (Of course that is doable by reconfiguring manually and changing the build directory, but the presets make that much more convenient.)

> preset version 3 goes with CMake 3.21

That is correct. The CMake presets are by nature an opt-in system, so people can continue using manual build configs with whatever version of CMake they have installed (including 3.12-<3.21). I decided that anyone who was interested in CMake presets would be interested enough to upgrade their CMake if necessary (although, presets being a feature of newer CMake's its likely that anyone who is familiar enough to be interested in CMake presets probably has a new enough CMake).

(I'm also bumping the minimum CMake version to 3.21 anyways in a forthcoming PR.)

sherm1:
OK, thanks. I'll merge this now.

PR #795, opened 2024-07-22
Check BLAS/LAPACK libraries at configure time

halleysfifthinc:
There are 4 ways simbody finds BLAS/LAPACK libraries:
  - Windows
    - Default (`WINDOWS_USE_EXTERNAL_LIBS=OFF AND BUILD_USING_OTHER_LAPACK=""`) is a set of simbody vendored BLAS/LAPACK libraries
    - If `WINDOWS_USE_EXTERNAL_LIBS=ON`, then the system BLAS/LAPACK is used (whatever is found by CMake FindBlas and FindLapack)
  - Unix default (`BUILD_USING_OTHER_LAPACK=""`) is the system BLAS/LAPACK (whatever is found by CMake FindBlas and FindLapack)
  - Set manually using `BUILD_USING_OTHER_LAPACK`

This could fail any number of ways (especially when using `BUILD_USING_OTHER_LAPACK`), so it
would be helpful to fail as early as possible if there is a problem. Currently, a build with
a poorly specified `BUILD_USING_OTHER_LAPACK` won't fail until linking the SimTKcommon
library, after building approximately 1/3 of Simbody.

This PR adds a `try_compile` check that try's to link to the chosen BLAS/LAPACK libraries.
If it fails, a helpful error message is printed by CMake:
```
CMake Error at CMakeLists.txt:537 (message):
  Failed to compile using the BLAS/LAPACK libraries -lnotablas.

          If BUILD_USING_OTHER_LAPACK was given; check that it was set correctly.
```

You can locally test this with either of these (valid) configurations (on Linux):
`cmake -B build -S . --fresh` 
`cmake -B build -S . --fresh -DBUILD_USING_OTHER_LAPACK="-lblas;-llapack"`

This should fail:
`cmake -B build -S . --fresh -DBUILD_USING_OTHER_LAPACK="-lnotablas"`

**Note:** This does not change any generated CMakefiles, and it won't break anything that isn't/wouldn't already be broken.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/795)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -483,6 +483,50 @@ add_compile_definitions(
    SimTK_SIMBODY_AUTHORS="${SIMBODY_AUTHORS}"
 )
 
+#
+# Allow automated build and dashboard.
+#
+
+if(BUILD_TESTING)
+    include(CTest) # automatically calls enable_testing()
+    ## When in Debug mode and running valgrind, some of the test
+    ## cases take longer than the default 1500 seconds.
+    set(CTEST_TESTING_TIMEOUT 7200)
+endif()
+
+# These are used in Doxyfile.in and SimbodyConfig.cmake.in.
+set(SIMBODY_INSTALL_DOXYGENDIR   "${CMAKE_INSTALL_DOCDIR}/api")
+set(SIMBODY_DOXYGEN_TAGFILE_NAME "SimbodyDoxygenTagfile")
+if( INSTALL_DOCS )
+    add_subdirectory(doc)
+endif()
+
+# Specify where visualizer should be installed. This needs to be in the
+# root CMakeLists.txt so the cmake config file can see this value.
+#
+# Also specify where include files are installed.
+if(WIN32)
+    # Install visualizer to bin, since it needs to be co-located with dll's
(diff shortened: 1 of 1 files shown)


halleysfifthinc:
Bump

sherm1 (commented):
Thanks again, Allen. This looks great and works as you said. I did notice on Linux that I get the message `No Package 'glut' found`. That's not due to your changes however; just a minor annoyance since it actually does find glut and use it successfully in the visualizer. In case you're in the mood for more CMake cleanup, it would likely avoid confusion if it didn't complain about something it actually found!

:lgtm:

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/795#-:-O55Yw6l3Bt_HjtTLTic:b-449x16)* status: 0 of 1 files reviewed, all discussions resolved

sherm1:
Mac CI fails now -- I _think_ it's related to this change. Can you tell?
https://github.com/simbody/simbody/actions/runs/10542062601/job/29208417524

halleysfifthinc:
Hmm looking at the CI logs for after the merge and the most recent run from this branch, it seems like there is an interaction between this branch and the CMake presets because the "found" BLAS/LAPACK are now different. I will investigate when I get back from vacation at the end of this week.

PR #798, opened 2024-09-17
Support polygonalmesh normals and texture coordinates

aymanhab:
This PR should allow us to read normals at vertices from mesh file formats that provide them. Some caveats:
- In cases like stl where a vertex can result from merging of loops, vertex normal is set to that of the first facet on which it appears.
- Nothing changed regarding the limitation of parsing (single solid/volume).
- The API change is made so that it's not breaking, if normals are available they maybe be used but any old code should work as is.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/798)


adamkewley:
Seems like a reasonable enough change!

The only thing I'd suggest is to get the normals via the face indirection, rather than vertex indices. I get that the data is almost always stored on a GPU as a 1:1 association in a vertex buffer, with a triangle-based index buffer, but the data file formats don't necessarily store them that way, and `SimTK::PolygonalMesh` is already assuming non-triangular (potentially, n-gon) buffers.

The API change suggested below should mean that everything here works basically how you want (you *can* store them the way you're storing them), but would also mean that later API changes are possible (someone later decides to store normals differently - no problem).

---

Just as an example, here's what exporting Blender's default cube to an `.obj` gives:

```txt
# Blender 4.2.1 LTS
# www.blender.org
mtllib untitled.mtl
o Cube
v 1.000000 1.000000 -1.000000
v 1.000000 -1.000000 -1.000000
v 1.000000 1.000000 1.000000
v 1.000000 -1.000000 1.000000
v -1.000000 1.000000 -1.000000
v -1.000000 -1.000000 -1.000000
v -1.000000 1.000000 1.000000
v -1.000000 -1.000000 1.000000
vn -0.0000 1.0000 -0.0000
vn -0.0000 -0.0000 1.0000
vn -1.0000 -0.0000 -0.0000
vn -0.0000 -1.0000 -0.0000
vn 1.0000 -0.0000 -0.0000
vn -0.0000 -0.0000 -1.0000
vt 0.625000 0.500000
vt 0.875000 0.500000
vt 0.875000 0.750000
vt 0.625000 0.750000
vt 0.375000 0.750000
vt 0.625000 1.000000
vt 0.375000 1.000000
vt 0.375000 0.000000
vt 0.625000 0.000000
vt 0.625000 0.250000
vt 0.375000 0.250000
vt 0.125000 0.500000
vt 0.375000 0.500000
vt 0.125000 0.750000
s 0
usemtl Material
f 1/1/1 5/2/1 7/3/1 3/4/1
f 4/5/2 3/4/2 7/6/2 8/7/2
f 8/8/3 7/9/3 5/10/3 6/11/3
f 6/12/4 2/13/4 4/5/4 8/14/4
f 2/13/5 1/1/5 3/4/5 4/5/5
f 6/11/6 5/10/6 1/1/6 2/13/6
```

Which is to say, the positions, normals, texture coordinates, and faces aren't all the same count. Also, from what I understand, OBJ allows for mixed faces (where some have normals, some don't, etc.).

Therefore, it might be better to rephrase the API a little bit. Maybe something like `bool hasCustomNormals() const`, to indicate to callers that at least one of (but not necessarily all of?) the normals coming out of this API aren't necessarily the same as they'd get if they computed them from the vertex positions using cross products etc.

I'd also suggest changing the API for getting normals to the `(face, index)` level - and to return the normal vector by value. E.g. the API could be `UnitVec3 getVertexNormal(int face, int index) const;`. The reason why is because `const UnitVec3& getVertexNormal(int index)` strongly couples vertex normal fetching to the same index as the vertex positions. This is not necessarily always true, given the design of the data files.

Returning the normal vector by value makes it possible to later compute the vertex normal on-the-fly. E.g. one could imagine that a later PR then adds a convenience feature like "if there's no custom normals, then go and compute a reasonable one from the positional data for the caller". This would help remove the currently-quite-unfortunate situation of Simbody Visualizer, OpenSim GUI, and OpenSim Creator all having their own normals-calculation methods.

aymanhab:
Awesome feedback @adamkewley thank you 👍 
Indeed I was biased by the threejs and gltf formats which store one normal per vertex so I wanted to simplify the data passing but the mesh file formats are a bit more general as you pointed out. Will update accordingly. Thanks again.

adamkewley:
No problem!

Note, i'm not married to either solution: there's merits in both (e.g. using vertex indices implies tying them together, which is actually a handy thing when working with graphics APIs like OSC's, threejs, etc.). Just pointing out that the underlying data files can be a little different!

aymanhab:
@adamkewley I took another pass and implemented the referencing by face/vertex which mimics the obj/stl format more closely. I can add a method to convert to visualization/gfx format where all vectors of vertices, normals and textures are the same but decided to leave this functionality on client side as I may defer the loading/compressing of the files to other tools. At a high level the polygonal mesh is now capable of representing all the relevant data in the file(s) and that seams like a reasonable end-point. Let me know what you think.

aymanhab:
@sherm1  Simbody build on windows fails due to Doxygen. We changed to "https://github.com/doxygen/doxygen/releases/download/Release_1_12_0/doxygen-1.12.0.windows.x64.bin.zip", for OpenSim, any thoughts on whether we upgrade doxygen version for simbody as well?

nickbianco:
FYI @aymanhab, I made the change in OpenSim based on @halleysfifthinc's pending PR (https://github.com/simbody/simbody/pull/801). So that change is coming to Simbody soon.

sherm1:
@aymanhab I have no attachment to any particular version of Doxygen -- feel free to change.
Re #801 just waiting for @halleysfifthinc to make a minor fix

adamkewley (changes requested):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798)* status: 0 of 4 files reviewed, 11 unresolved discussions (waiting on @aymanhab)

___
*[`SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h` line 234 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwt_T74NJOV1bFAVFC:-OAwmYwr6qlETXHU8zOS:bounufx) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h#L234)):*
> ```objc
>     int addFace(const Array_<int>& vertices);
> 
>     int addFaceWithNormals(const Array_<int>& vertices,
> ```

This API implies that the caller is able to retrieve whole normals vector in order to index into it. In this design, the size of the normals vector (and indices into it) seem like they could be disjoint from the size of the vertex indices. Because of that, the rest of the API has to reflect the fact that there could be a disjoint.
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 81 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwsTyV694KrVrftIES:-OAwsTyV694KrVrftIET:b-nkkcay) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L81)):*
> ```cpp
> const Vec2 PolygonalMesh::getVertexTextureCoordinate(int faceIndex,
>                                                      int vertexIndex) const {
>     assert(0 <= faceIndex && faceIndex < getNumFaces());
> ```

Assertions are disabled in release builds, and there's no way for a caller to check if a specific face+vertex tuple has a texture coordinate or normal (there's only asking whether the mesh has >0 normals/coords), so this API guarantees UB if (e.g.) an OBJ is loaded but has sparse coordinates/normals.

The preconditions either need to be:

- The file is loaded and `hasNormals` is `true`. In this case, ALL vertices will have a normal (even if defaulted). If `hasNormals` is false, any calls to `getVertexNormal` with any combination of face/vertex will fail with an exception

Or:

- The file is loaded and `hasNormals` is `true`. The API has a "tryGet" style API that returns a pointer to the normals, or nullptr if that particular normal isn't found. Alternatively, `std::optional`. Alternatively, `bool tryGet(int face, int vertex, UnitVec3& out)`
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 484 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwrQdJCFmWiBzgeoGo:-OAwrQdJCFmWiBzgeoGp:bry22gn) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L484)):*
> ```cpp
>     }
>     
>     std::cout << textureCoordinates << std::endl;
> ```

Debug printer?
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 503 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwqnjM4r7QCilM-67w:-OAwqnjM4r7QCilM-67x:b-hs52gt) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L503)):*
> ```cpp
>     // Now that we have the point coordinates, use them to create the vertices
>     // in our mesh.
>     // VTP format provides one normal per vertex
> ```

This is useful to keep in mind, given the sparsity-assuming datastructure design. I imagine that most file formats are going to be all-or-nothing on normals/texture coordinates, because sparsity implies the implementation has an adequate way of filling the gaps. I'd try out MeshLab/Blender etc. and see if any of them ever export OBJs with sparse normals/coords because, if not, you can just have one index buffer (all indices are the same).
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 740 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwqgoU-enne1UaBW4O:-OAwqgoU-enne1UaBW4P:bua1vdu) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L740)):*
> ```cpp
> 
>     Array_<int> vertices;
>     Array_<int> normals;
> ```

Are these `normals` or `normalIndices`?
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 762 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwq1GMAK5GFOpHdxl-:-OAwq1GMAK5GFOpHdxl0:bd7xqle) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L762)):*
> ```cpp
>                 normalsSeen = true;
>             } 
>             
> ```

Whitespace in added lines
___
*[`SimTKcommon/Geometry/src/PolygonalMeshImpl.h` line 53 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwozil3s0giPbXvm1X:-OAwozil3s0giPbXvm1Y:biybks0) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMeshImpl.h#L53)):*
> ```objc
>     Array_<int>     faceVertexStart;
>     Array_<UnitVec3> normals;
>     Array_<int> faceVertexNormalIndex;
> ```

The presence of these members implies there's an index mis-match between the indices of face vertices and the indices of normals/texture coordinates attached to the vertices. Given OBJ/VTP's specifications, that could be the case, but the current API design of `PolygonalMesh` implies that it's always possible to retrieve (e.g.) a normal for a given index. Otherwise, an exception is thrown. The caller is only capable of checking if >0 normals/coords are available.

If sparsity/mismatching is possible, then the API should reflect the fact that retrieving a normal/texture coordinate can fail. In C++17 that'd be `std::optional<UnitVec3>`. Here, it could be `const UnitVec3*`, if you believe the underlying mesh implementation will always store `UnitVec3`s and never materialize them. Otherwise it'd be `bool F(...params..., UnitVec3& out)`
___
*[`SimTKcommon/tests/TestPolygonalMesh.cpp` line 195 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwmrWh15ynhEoVFqFJ:-OAwmrWh15ynhEoVFqFK:b-5ekmmh) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/tests/TestPolygonalMesh.cpp#L195)):*
> ```cpp
>     PolygonalMesh mesh;
>     string fileContent;
>     fileContent += "<?xml version='1.0'?>";
> ```

C++ supports multiline strings, so this can be included as `fileContent = R"(content here)"` see "Raw string literals" in:

https://en.cppreference.com/w/cpp/language/string_literal
___
*[`SimTKcommon/tests/TestPolygonalMesh.cpp` line 275 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwnAGH2Udft4RjyQo9:-OAwnAGH2Udft4RjyQoA:b-15kewe) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/tests/TestPolygonalMesh.cpp#L275)):*
> ```cpp
>     fileContent += "</VTKFile>";
> 
>     std::ofstream filePtr("plane.vtp", std::ios::out);
> ```

If the test *must* use IO, then this is ok, but if it's acceptable to use a stream-based overload, it'd be preferred. Writing files as a side-effect of running tests--especially when not using temporary files--may have unusual consequences (usually, it's fine though).
___
*[`SimTKcommon/tests/TestPolygonalMesh.cpp` line 283 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwnZS2-n2ONyUBT7tQ:-OAwnZS2-n2ONyUBT7tR:b-3kof3f) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/tests/TestPolygonalMesh.cpp#L283)):*
> ```cpp
>     // Normals = 0 0 1, 0 0 1,0 0 1, 0 0 1
>     Vec3 verts[] ={
>         {-.5, -.5, 0}, {.5, -.5, 0}, {-.5, .5, 0}, {.5, .5, 0}};
> ```

The types of these init lists are `<double, double, int>`. The `int` is  (luckily) converted but high-warning-level compilers might cry that `int` was implicitly converted to `double` in an expression like this.
___
*[`SimTKcommon/tests/TestPolygonalMesh.cpp` line 286 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwnsL91fJejSWb5cMO:-OAwnsL91fJejSWb5cMP:b-eqiboy) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/tests/TestPolygonalMesh.cpp#L286)):*
> ```cpp
>     ASSERT(mesh.getNumVertices() == 4);
>     ASSERT(mesh.getNumFaces() == 1);
>     for (int v = 0; v < 4; ++v) {
> ```

Use `std::array<Vec3, 4> verts = {init expr};` so that this loop can use `verts.size()` instead of hard-coding the `4`. Alternatively, `std::size` works with traditional C arrays (but might not be available in C++11, unsure).

adamkewley:
@aymanhab I think I understand the intent, but my main comment would be that the API design doesn't match up with the underlying data structure. If the intent is that there's going to be three index buffers (vertices/normals/coords) and they can each be of different size/indices (i.e. there's sparsity between vertex positions and normals) then the API needs to provide the caller with sufficient querying methods, or fail-able getters, in order to be able to pull out what's necessary from the class.

From what I've seen of game engines etc., there's almost always one index buffer that indexes into either a striped vertex buffer, or separate vertex buffers that are of the same size and topology. E.g. most engine APIs will assume an all-or-nothing association between vertices and normals. That's obviously different from a file specification (OBJ does allow for sparsity), so the design either needs to reflect file specifications (by having fail-able lookups) or reflect graphics engine requirements (by validating the files *must* have all data, or defaulting missing values, or computing reasonable values for missing values).

aymanhab:
Thanks for the feedback @adamkewley 👍 I totally agree that more query methods would help to make sure clients avoid stepping on bombs 💣 

As I explained in my comment, my intention was to get the relevant data from the file(s) into in-memory structures, and leaving the rest to the client to use as they see fit. The sizes of the vectors are not always the same (true for vtp files I've seen, but false for obj, stl). I think a typical workflow for a client would be to cycle through faces, triangulate, compute normals if needed and produce vectors of equal length   of vertices, normals and texture coordinates if available.  My question above, do you think this belongs here or to the client? and if here would you do it in-place or pass in fresh vectors to populate and send to visualization?

adamkewley:
For filling in the blanks etc. I'd say that it's ok for downstream clients to handle it, because they will likely have their own datastructures with their own memory layout etc.

aymanhab (commented):
@adamkewley I decided to not allow gaps (ignore normals and/or textures if gaps exist primarily because the types of objects we parse are already limited to shells that can be used for contact and/or visualization rather than arbitrary topology).  The comments in the header should reflect that decision/limitation. I added query methods to check if normals are available at faces (as in obj, stl) or vertices (vtp) and added code in test case to convert the face/vertex layout to vertex-list layout as typically used for visualization and by gltf format.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798#-:-OBgLO-a9oKFbVqOh0sb:b-9fykn2)* status: 0 of 4 files reviewed, 9 unresolved discussions (waiting on @adamkewley)

___
*[`SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h` line 234 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwt_T74NJOV1bFAVFC:-OBh4ssM9GD0tIh8yIZo:b-fk0mer) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/include/SimTKcommon/internal/PolygonalMesh.h#L234)):*

This API implies that the caller is able to retrieve whole normals vector in order to index into it. In this design, the size of the normals vector (and indices into it) seem like they could be disjoint from the size of the vertex indices. Because of that, the rest of the API has to reflect the fact that there could be a disjoint.
</blockquote></details>

For normals they have to be at all faces or all vertices if neither then they are ignored and normals are wiped. For textures they are either per face/vertex or per vertex otherwise unused. There're query methods to check which one of these scenarios is the case.
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 81 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwsTyV694KrVrftIES:-OBh5Uqp9RqYWudFrXO8:blmbwwg) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L81)):*

Assertions are disabled in release builds, and there's no way for a caller to check if a specific face+vertex tuple has a texture coordinate or normal (there's only asking whether the mesh has >0 normals/coords), so this API guarantees UB if (e.g.) an OBJ is loaded but has sparse coordinates/normals.

The preconditions either need to be:

- The file is loaded and `hasNormals` is `true`. In this case, ALL vertices will have a normal (even if defaulted). If `hasNormals` is false, any calls to `getVertexNormal` with any combination of face/vertex will fail with an exception

Or:

- The file is loaded and `hasNormals` is `true`. The API has a "tryGet" style API that returns a pointer to the normals, or nullptr if that particular normal isn't found. Alternatively, `std::optional`. Alternatively, `bool tryGet(int face, int vertex, UnitVec3& out)`
</blockquote></details>

I chose not to use TryGet since it complicates the logic of the client and at any rate we don't know what to do/pass if texture coordinates for example are only partly specified.
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 484 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwrQdJCFmWiBzgeoGo:-OBk_OvS0iyfoZE3Ikd-:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L484)):*

Debug printer?
</blockquote></details>

Done.
___
*[`SimTKcommon/tests/TestPolygonalMesh.cpp` line 275 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwnAGH2Udft4RjyQo9:-OBkbZCP4PSaf9OG1VYu:b8hh65x) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/tests/TestPolygonalMesh.cpp#L275)):*

If the test *must* use IO, then this is ok, but if it's acceptable to use a stream-based overload, it'd be preferred. Writing files as a side-effect of running tests--especially when not using temporary files--may have unusual consequences (usually, it's fine though).
</blockquote></details>

It's not a must but the vtp file parser doesn't handle streams and I wanted to have the contents of the file baked into the test case instead of chasing where the file is located in the file system. If you feel strongly about it I can use the file directly though.

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @aymanhab)

___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 201 at r3](https://reviewable.io/reviews/simbody/simbody/798#-OC3Fiq5121vERVxcWSe:-OC3Fiq5121vERVxcWSf:biqnlwg) ([raw file](https://github.com/simbody/simbody/blob/59bdfb4591ad9e5f14ae0f9439ea19cdf8d73e17/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L201)):*
> ```cpp
>         for (int i = 0; i < (int) vertices.size(); i++)
>             vertices[i] = X_AM*vertices[i];
>         // What to do to normalsIndices if available?
> ```

In this case, the normal *values* also require transformation. Because `SimTK::Transform` doesn't include a scale, this probably just involves only applying the rotational component of the transform to the normals.
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 320 at r3](https://reviewable.io/reviews/simbody/simbody/798#-OC3I_8z8GGWqr04YGWV:-OC3I_8z8GGWqr04YGWW:b-80u8rg) ([raw file](https://github.com/simbody/simbody/blob/59bdfb4591ad9e5f14ae0f9439ea19cdf8d73e17/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L320)):*
> ```cpp
>                     tIndex--;
>                     textureIndices.push_back(tIndex);
>                     // 
> ```

Stray

aymanhab (commented):
Done with latest round, thanks @adamkewley 

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798#-:-OCA2rjW7oEtu7hgUZBq:b863fmj)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @adamkewley)

___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 503 at r2](https://reviewable.io/reviews/simbody/simbody/798#-OAwqnjM4r7QCilM-67w:-OCA24VK22cmlBoRciyf:bnonqje) ([raw file](https://github.com/simbody/simbody/blob/fa9b36b2fdbcc7d7c853a313628995345683dca9/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L503)):*

This is useful to keep in mind, given the sparsity-assuming datastructure design. I imagine that most file formats are going to be all-or-nothing on normals/texture coordinates, because sparsity implies the implementation has an adequate way of filling the gaps. I'd try out MeshLab/Blender etc. and see if any of them ever export OBJs with sparse normals/coords because, if not, you can just have one index buffer (all indices are the same).
</blockquote></details>

Done, thanks for catching before falling into the cracks :+1: 
___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 320 at r3](https://reviewable.io/reviews/simbody/simbody/798#-OC3I_8z8GGWqr04YGWV:-OCA21x08BHLkW6eWC_X:bb6un1j) ([raw file](https://github.com/simbody/simbody/blob/59bdfb4591ad9e5f14ae0f9439ea19cdf8d73e17/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L320)):*

Stray
</blockquote></details>

Done

adamkewley (commented):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @aymanhab)

___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 103 at r5](https://reviewable.io/reviews/simbody/simbody/798#-OCbe2ro6UbFj5UGXgmh:-OCbe2ro6UbFj5UGXgmi:b-qy6vnu) ([raw file](https://github.com/simbody/simbody/blob/cb6f369fe789749948a7dcfdfd18d46bc40232b6/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L103)):*
> ```cpp
> 
> bool PolygonalMesh::hasNormalsAtFaces() const {
>     return getNumFaces() == getImpl().normals.size();
> ```

Just an observation, but this logic breaks if the number of faces is equal to the number of vertices.

E.g. a three-sided pyramid, including the base - 4 points, 4 faces. Those normals would be different. Face-based normals would be orthogonal to the surface (face) of the pyramid. Vertex-based normals would (presumably) point outward from the core of the pyramid, such that taking the average of three of them around any face roughly reproduces a face normal.

aymanhab (commented):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @adamkewley)

___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 103 at r5](https://reviewable.io/reviews/simbody/simbody/798#-OCbe2ro6UbFj5UGXgmh:-OCjDYzz23aeWqq-0ud3:b-jvd8mo) ([raw file](https://github.com/simbody/simbody/blob/cb6f369fe789749948a7dcfdfd18d46bc40232b6/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L103)):*

Just an observation, but this logic breaks if the number of faces is equal to the number of vertices.

E.g. a three-sided pyramid, including the base - 4 points, 4 faces. Those normals would be different. Face-based normals would be orthogonal to the surface (face) of the pyramid. Vertex-based normals would (presumably) point outward from the core of the pyramid, such that taking the average of three of them around any face roughly reproduces a face normal.
</blockquote></details>

Good point that crossed my mind too but fell through the cracks.  Clients should test first if normals are at vertices (as these can be interpolated). Would documenting that be enough in your opinion?

aymanhab (commented):
Fixed a bug I introduced in parsing stl binary files, added comments to clarify the ambiguity. Mainly that the file format typically indicate what info is available. Hope this helps

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/798#-:-OD7l-y99I2UtAr73ewp:b79vvr1)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @adamkewley)

___
*[`SimTKcommon/Geometry/src/PolygonalMesh.cpp` line 103 at r5](https://reviewable.io/reviews/simbody/simbody/798#-OCbe2ro6UbFj5UGXgmh:-OD7kdxI1DZ6ZWDbkO0H:b1gvp59) ([raw file](https://github.com/simbody/simbody/blob/cb6f369fe789749948a7dcfdfd18d46bc40232b6/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L103)):*

I mean, whatever works, but it'd be ambiguous in that case too? The pyramid example would return true for both having normals at faces and having normals at vertices, because the logic of the check is only looking at quantities of collections
</blockquote></details>

I added a clarifying comment that may help. Let me know

aymanhab:
@adamkewley I believe I addressed your last round of feedback. Looking forward to moving this along as the viewer backend is working/tested now with this API and would be good to make available to others. Thank you 👍

adamkewley:
@aymanhab , sorry, I thought I concluded it etc -  I think it's all fine, nice work

aymanhab:
Awesome, thanks so much @adamkewley 🥇 you definitely made it better with your excellent review/feedback.

PR #807, opened 2024-12-11
Handle vtp files without normals

aymanhab:
The last PR that enabled reading normals and texture coordinates, mistakenly required vtp files to contain normals. While true for visualization files, this is not always the case. This PR fixes this assumption and adds a test case to verify proper handling

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/807)

### DIFF
--- a/SimTKcommon/tests/TestPolygonalMesh.cpp
+++ b/SimTKcommon/tests/TestPolygonalMesh.cpp
@@ -406,6 +406,100 @@ void testLoadVtpFile() {
     ASSERT(mesh.hasTextureCoordinatesAtVertices())
 }
 
+void testLoadVtpFileNoNormals() {
+    PolygonalMesh mesh;
+    string fileContent;
+    fileContent += "<?xml version='1.0'?>";
+    fileContent +=
+        "<VTKFile type='PolyData' version='0.1' byte_order='LittleEndian' "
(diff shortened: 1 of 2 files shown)


aymanhab:
@nickbianco

SimTKcommon/tests/TestPolygonalMesh.cpp
```diff
@@ -406,6 +406,100 @@ void testLoadVtpFile() {
...
+    fileContent += "</PolyData>";
+    fileContent += "</VTKFile>";
+
+    std::ofstream filePtr("planeNoNormals.vtp", std::ios::out);
+    filePtr << fileContent;
+    filePtr.close();
+
+    mesh.loadVtpFile("planeNoNormals.vtp");
+    // verts = -0.5 -0.5 0, 0.5 -0.5 0, -0.5 0.5 0,  0.5 0.5 0
+    // Normals = 0 0 1, 0 0 1,0 0 1, 0 0 1
```
nickbianco, line 492: Remove this comment since there are no normals in this file.

nickbianco (approved):
One minor suggestion to remove a comment in the next test for clarity, but otherwise LGTM.

aymanhab:
Thanks @nickbianco

PR #809, opened 2025-01-20
Ensure loadVtpFile works when the vtp file doesn't have a <PointData>…

adamkewley:
Related #807 #798

The latest vtp parser is unable to parse the VTP files contained in this directory:

- https://github.com/opensim-org/opensim-models/tree/master/Geometry

Because the files don't contain a `<PointData>` block, which is what the new parser now requires in order to process texture coordinates and normals.

@aymanhab @nickbianco

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/809)


aymanhab:
@adamkewley The change looks good, I may not have seen it because all the files I worked with had this element. Thanks for the find 👍

aymanhab (approved):

PR #811, opened 2025-01-23
Fix STL parser memory usage growing quadratically

adamkewley:
Found downstream:

- https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/987

The new additions to the STL parser for handling normals was incorrectly using `push_back` rather than assigning the normals over the same array. This results in `normals^2` memory usage, which means loading a model that uses STLs with the new parser can consume **a lot** of memory (e.g. jumped from 0.2 -> 14 GB on my machine). @aymanhab

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/811)


adamkewley:
Merging this because it's a oneliner

PR #814, opened 2025-02-21
CableSpan: A length minimizing algorithm

pepbos:
# Update to Cost Function used in a CableSpan
This PR changes the cost function used to compute the optimal path of a CableSpan such that it is a minimal length path.
Both the original algorithm as well as the new algorithm can be selected:

```cpp
cable.setAlgorithm(CableSpanAlgorithm::Scholz2015); // the original algorithm.
cable.setAlgorithm(CableSpanAlgorithm::MinimalLength); // this PR
```

By *original algorithm* or `CableSpanAlgorithm::Scholz2015` we mean the algorithm as published in:

Scholz, A., Sherman, M., Stavness, I. et al (2016). A fast multi-obstacle muscle wrapping method using natural geodesic variations. Multibody System Dynamics 36, 195–219.

## Issue with Current Algorithm

The `CableSpan` should compute the path that a cable under tension would take over smooth, frictionless obstacles: i.e. a shortest path. However, the current algorithm can not distinguish between a longest and a shortest path. To illustrate the problem, consider the following computed motion:

https://github.com/user-attachments/assets/5acf9dc9-ddaf-4334-894b-3e98a7de96ad

Firstly this path is not the shortest. Secondly, the motion appears unnatural (Perhaps try at home :)). 

By comparison, here is the shortest path:

https://github.com/user-attachments/assets/08a73872-57ca-475c-829e-a8eae9d77df0

The path starts the same, but any small motion causes the cable to snap off the surface. The resulting path might seem trivial, just a straight line, but you have to admit: It *is* shorter.

If we wish to visualize what happens in that first frame, we have to visualize the intermediate solver steps. Starting from the same initial condition, we see that the optimizer step direction really is different:

https://github.com/user-attachments/assets/1ff00dd4-5173-4f20-bbe1-80e0235817c8

In the above video we see two identical wrapping scenarios, with the intermediate solver steps visualized. The path is initialized by the chosen contact point on top of the obstacle. The left cable uses the original algorithm, whereas the right cable is minimizing the cable length. Note that the right cable detects liftoff at some point, and the path becomes a straight line.

The fact that the original algorithm can not distinguish between a maximum and a minimum cable length follows from the chosen cost function. The path errors, as defined in the paper, are zero for *any* cable length optimum.

**Ill conditioning of the path error Jacobian**

There are some robustness issues with the current algorithm as well: Below we compare wrapping over an ellipsoid, the original algorithm on the left, and a minimal length path on the right.

https://github.com/user-attachments/assets/69d8cd75-e17f-4664-a2fa-836aee998414

The path error Jacobian can loose rank, and we see this happening at around 10 seconds, where the original algorithm (left) becomes ill conditioned. Eventually it passes, but unfortunately it moves to the maximum length, and stays there.

**Robustness during initialization**

Finally another issue with the original algorithm is that there are two ways to align the straight line segments with the Frenet frame tangents: $e_{P,Q} = t_{P,Q}$ and $e_{p,Q} = -t_{P,Q}$. That is, the straight line segments pointing in opposite direction of the curve tangent is also a valid solution. This appears to affect the robustness during initialization, when the initial path errors are close to 90 degrees:

https://github.com/user-attachments/assets/a18964d7-4a29-45bc-b1c1-bc0da1830206

where again, on the left is the original algorithm, and the right a minimal length solution. The video shows how the optimizer steps from the initial path, just a point, to the optimal path. The original algorithm on the left is attempting to align the Frenet frame's tangent with the *negative* of the straight line segment's direction.

## A Minimal Length Cost Function

In the above examples the shortest length paths were obtained with the update of this PR. The main change is to the cost function used to compute the optimal path. The fundamental concepts proposed in the original paper are still the same: The path consists of straight line segments, connecting curve segments for each obstacle (geodesics), and the natural geodesic variations are the free variables of the problem.

**The original cost function**

In the original paper the defined path error captures the misalignment of the straight line segments with the curve segments, as angular discontinuities in the total cable path. A smooth path would have no discontinuity, and would be shortest.

The following image taken from the paper illustrates the path errors:

Scholz2015-dissert-img

with the the normal- and binormal path errors:
- Normal path error: $\epsilon_{N_i} = [ e^T_i N_{P_i} \quad e^T_{i+1} N_{Q_i} ]^T$
- Binormal path error: $\epsilon_{B_i} = [ e^T_i B_{P_i} \quad e^T_{i+1} B_{Q_i} ]^T$

and the total path error simply stacking these for all obstacles in a single vector $\epsilon$.  Driving $\epsilon\rightarrow 0$ results in the optimal path.
However, as illustrated in the given examples, there are some issues with using these path errors as the cost function.

**The new cost function**

This PR changes the cost function to be the total cable length itself. Using the Natural Geodesic Variations described in the original paper we can compute a second order approximation of the cable length:

$`
l(q) = l + g^T q + \frac12 q^T H q + ...
`$

where $l$ is the total cable length, $q$ is the vector of natural geodesic corrections, $g$ is the gradient of the cable length, and $H$ is the Hessian of the cable length to $q$.

In addition we impose the normal path errors (as defined above) as equality constraints. Using the Natural Geodesic Variations we can compute a first order approximation of these constraints:

$`
\epsilon_N(q) = \epsilon_N + J q + ...
`$

with $J$ the Jacobian of $\epsilon_N$ to $q$.

We arrive at the optimization problem:

**minimize $l(q)$  subject to $c(q) = 0$**

The similarity with the original algorithm is that the normal path errors are enforced as equality constraints, but that the binormal path error is replaced by the length as a cost function. Minimizing the length will drive the binormal path errors to zero.

**Minimizing the new cost function**

We solve the problem as a Quadratic Program using the Lagrangian formulation:

$`
\begin{bmatrix} Q & J^T \\\ J & 0 \end{bmatrix}
\begin{bmatrix} q \\\ \lambda \end{bmatrix} =
\begin{bmatrix} -g \\\ - \epsilon_n \end{bmatrix}
`$

where $Q$ is a positive definite matrix, and $\lambda$ are the Lagrange multipliers.

A natural choice for the matrix $Q$ would be the computed Hessian $H$. However, the Hessian of the length is not always positive (semi) definite. Near the maximum cable length some eigenvalues are in fact negative. We therefore take $Q$ as the Hessian, but with any negative eigenvalue flipped, such that it is positive definite. This can be achieved as follows:
1. Take the Singular Value Decomposition: $\frac12\left(H + H^T\right) = U\Sigma V^*$
2. Compute $Q$ as: $Q = U\Sigma U^T$
Note that $Q$ has the same eigenvectors as $\frac12 \left(H+H^T\right)$, as well as the same eigenvalues up to a sign difference. Clearly $Q$ is positive definite, which means that solving the posed Lagrangian will meet the KKT conditions.

The stopping condition of the path solver remains the same: The original path error correctly detects an *optimum* in the cable length.

## Testing

**New Unit Tests**
Two new unit tests are added to `TestCableSpan.cpp`:
- `testSolverOptimum` verifies that the original algorithm converges to the maximum cable length, and that the minimal length algorithm converges to the minimal cable length, for the same initial conditions.
- `testRobustnessInitialPath` verifies that the minimal length algorithm correctly initializes the cable path, even if the initial path error is close to 90 degrees.

**Modified Unit Tests**
A number of unit tests were using a maximum length path for testing, and have been changed to test a minimal length path. For example `TestCableSpan.cpp::testCableSpanSimple` was using a maximum length path:

TestCableSpan_simple_oriiginal

by replacing the sphere with a torus, and elongating the ellipsoid, the same path was tested, but now as a minimal length path:

TestCableSpan_simple_modified

**Gradient and Hessian tests**
The computed gradient and Hessian of the cable length are added to the perturbation test of `CableSubsystemTestHelper`. They are numerically verified along with the path error Jacobian in the unit tests.

**Existing tests**
With the above modifications and additions the changes in this PR pass the existing unit tests. These tests include spanning over multiple obstacles, spanning over different obstacle kinds, detecting touchdown and liftoff, etc.

## Summary of Changes

Public changes to `Simbody/include/simbody/internal/CableSpan.h:
- The enum `CableSpanAlgorithm` for choosing the algorithm is added
- A new function is added to select the algorithm `CableSpan::setAlgorithm` 

Changes to `Simbody/src/CableSpan.cpp`:
- The cached `MatrixWorkspace` now contains the matrices required for the new algorithm.
- The function `CableSpan::calcSolverStep` computes the path corrections, and contains the switch between both algorithms
- The functions `calcLengthGradient` and `calcLengthHessian` compute the gradient and Hessian
- The function `CableSubsystemTestHelper::Impl::runPerturbationTest` is expanded to also numerically validate the computed gradient and Hessian.

New unit tests:
- `Simbody/tests/TestCableSpan.cpp::testSolverOptimum`: New test verifying that the original algorithm computes the maximum length path, and the new algorithm the minimal length path.
- `Simbody/tests/TestCableSpan.cpp::testRobustInitialPath`: New test verifying convergence from initial conditions that have a large path error (close to 90 degrees).

The following unit tests were slightly modified to obtain a minimal length path (without changing the purpose of the test):
- `Simbody/tests/TestCableSpan.cpp::testSimpleCable`
- `Simbody/tests/TestCableSpan_Forces.cpp::testCableForceAngMomentOnSingleObstacle`
- `Simbody/tests/TestCableSpan_LengthDot.cpp::testCableLengthDotWithSingleObstacle`

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/814)

### DIFF
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -35,6 +35,85 @@ This is a unique integer type for quickly identifying specific cables for fast
 lookup purposes. These begin at zero for each CableSubsystem. **/
 SimTK_DEFINE_UNIQUE_INDEX_TYPE(CableSpanIndex);
 
+/** @enum SimTK::CableSpanAlgorithm
+These are the different solver options for computing the optimal path of a
+CableSpan. These options change the cost function and the computed descending
+direction for reaching the optimal path.
+see CableSpan::setAlgorithm **/
+enum class CableSpanAlgorithm
+{
+    /** This is the original algorithm as described in:
+
+    Scholz, A., Sherman, M., Stavness, I. et al (2015). A fast multi-obstacle
+    muscle wrapping method using natural geodesic variations. Multibody System
+    Dynamics 36, 195–219, DOI https://doi.org/10.1007/s11044-015-9451-1.
+
+    The path error vector captures the misalignment of the straight line and
+    curved segments at the contact points. The computation of the path is then
+    done by driving this path error to zero (up to tolerance), such that all
+    segments are smoothly connected, and the optimization step size converges
+    to zero.
+
+    A drawback of this algorithm is that it converges to any cable length
+    optimum. This means that the cable might be of maximum length or minimal
+    length. **/
+    Scholz2015,
+    /** The Minimal length algorithm finds the optimal path by minimizing the
+    total cable length directly, while enforcing the contour constraints.
+
+    The first step is to describe the total cable length using a second order
+    approximation:
+    <pre>
+    l(q) = l + ~g * q + ~q * H * q / 2 + ...
+    </pre>
+    where l is the total cable length, q is the vector of natural geodesic
+    corrections (see Scholz2015 paper), g and H are the gradient and Hessian of
+    l to q.
+
+    For the constraints we use the normal path errors described in the
+    Scholz2015 paper:
+    <pre>
+    c_i(q) = [ ~e_P * n_P ]
+             [ ~e_Q * n_Q ]
+    </pre>
+    where c is the vector of constraints, c_i are the two constraint elements
+    associated with the i-th curve segment, e_P is the direction of the
+    straight line segment at the P frame, n_P is normal direction at the
+    P-frame, and similarly for e_Q and n_Q.
+    Taking the first order approximation of these constraints gives:
+    <pre>
+    c(q) = c + J * q + ...
+    </pre>
+
+    The optimization problem that is solved is then posed as:
+
+    minimize l(q) subject to c(q) = 0
+
+    This problem is solved by repeatedly solving a Quadratic Program given by:
+    <pre>
+    [ Q   J^T ]   [ q ]   [ -g ]
+    [ J   0   ] * [ λ ] = [ -c ]
+    </pre>
+    where Q is a symmetric positive definite approximation of H, and λ are the
+    lagrange multipliers.
+
+    The approximation Q is used because H is itself not symmetric positive
+    definite. Q is obtained from H by flipping the sign of any negative
+    eigenvalues of H, to get a positive definite approximation. Consider the
+    following eigen decomposition:
+    <pre>
+    (H + ~H)/2 = ~P D P
+    </pre>
+    Then Q is given as:
+    <pre>
+    Q = ~P abs(D) P
+    </pre>
+
+    For a comparison between Algorithm::Scholz2015 and this
+    Algorithm::MinimumLength see https://github.com/simbody/simbody/pull/814. **/
+    MinimumLength,
+};
+
 class MultibodySystem;
 class CableSubsystem;
 class CableSubsystemTestHelper;
@@ -302,6 +381,9 @@ public:
     See CableSpan::getSmoothnessTolerance.
     State must be realized to Stage::Position. **/
     Real getSmoothness(const State& state) const;
+
+    /** Set the algorithm used to compute the optimal path. **/
+    void setAlgorithm(CableSpanAlgorithm algorithm);
 
     ///@}
 
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -2616,7 +2616,7 @@ void calcLengthHessian(
             const Real aDot = dataInst.jacobiDot_Q[0];
             const Real rDot = dataInst.jacobiDot_Q[1];
 
-            // If we assume the variation of the jaobi fields is zero, then the
+            // If we assume the variation of the jacobi fields is zero, then the
             // columns of v_Q have a constant norm. Then, the Hessian of the
             // straight line segment at the Q-frame is computed analogously as
             // done at the P-frame:
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -118,36 +118,48 @@ matrices need to be computed. This struct provides the required matrices with
 appropriate dimensions. After computing the path this data is no longer needed
 by the CableSpan. */
 struct MatrixWorkspace {
-    // Number of path error constraints per curve segment.
-    static constexpr int c_NumPathErrorConstraints = 4;
-    // Total number of constraints per curve segment.
-    static constexpr int c_NumConstraints = c_NumPathErrorConstraints + 1;
-
     // Given the number of CurveSegments that are in contact with their
     // respective obstacle's surface, contruct a MatrixWorkspace of correct
     // dimensions.
-    explicit MatrixWorkspace(int problemSize) : nObstaclesInContact(problemSize)
+    explicit MatrixWorkspace(int problemSize)
     {
-        static constexpr int Q = c_GeodesicDOF;
-        static constexpr int C = c_NumConstraints;
-        const int n            = problemSize;
+        // Total number of obstacles in contact with cable.
+        const int n = problemSize;
+        // Dimension of the free variable vector: all geodesic corrections.
+        const int nQ = n * c_GeodesicDOF;
+        // Total number of constraints (2 per obstacle).
+        const int nC = n * 2;
 
         lineSegments.resize(n + 1);
-        pathErrorJacobian = Matrix(C * n, Q * n, 0.);
-        pathCorrection    = Vector(Q * n, 0.);
-        pathError         = Vector(C * n, 0.);
+        pathCorrection = Vector(nQ, 0.);
+
+        normalPathError   = Vector(nC, 0.);
+        binormalPathError = Vector(nC, 0.);
+
+        normalPathErrorJacobian = Matrix(nC, nQ, 0.);
+
+        A      = Matrix(nC, nC, 0.);
+        b      = Vector(nC, 0.);
+        lambda = Vector(nC, 0.);
+
+        lengthGradient               = Vector(nQ, 0.);
+        lengthHessian                = Matrix(nQ, nQ, 0.);
+        lengthHessianInverseEstimate = Matrix(nQ, nQ, 0.);
+
+        singularValues      = Vector(nQ, 0.);
+        leftSingularValues  = Matrix(nQ, nQ, 0.);
+        rightSingularValues = Matrix(nQ, nQ, 0.);
     }
 
     // Return the NaturalGeodesicCorrection for the curve segment at the
     // "active" curve index, where "active" means counting those that are in
     // contact with the obstacle.
-    NaturalGeodesicCorrection getCurveCorrection(int activeCurveIx) const
+    static NaturalGeodesicCorrection getCurveCorrection(
+        const Vector& pathCorrection,
+        int activeCurveIx)
     {
         SimTK_ASSERT(
-            activeCurveIx < nObstaclesInContact,
-            "Index of curve (counting active only) exceeds number of segments in contact with obstacles");
-        SimTK_ASSERT(
-            pathCorrection.size() == nObstaclesInContact * c_GeodesicDOF,
+            pathCorrection.size() <= (activeCurveIx + 1) * c_GeodesicDOF,
             "Invalid size of pathCorrection vector.");
         const int eltIx = activeCurveIx * c_GeodesicDOF;
         return {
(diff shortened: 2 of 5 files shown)


sherm1:
This looks like a major improvement! The exposition above is compelling. I hope you have plans to publish this.

Pepijn, this PR is currently set to "Draft". When you're ready to have it reviewed you should change it to "Ready for review" (button at the bottom). Do you have a proposal for a first reviewer? I'm happy to do a final review pass and merge.

Thanks and regards, Sherm

sherm1:
BTW this failed on Windows CI due to this error:
```
D:\a\simbody\simbody\Simbody\src\CableSpan.cpp(2835,64): error C2220: the following warning is treated as an error [D:\a\simbody\simbody\out\build\ci-msvc-windows-release\Simbody\sharedTarget\SimTKsimbody.vcxproj]
D:\a\simbody\simbody\Simbody\src\CableSpan.cpp(2835,64): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data [D:\a\simbody\simbody\out\build\ci-msvc-windows-release\Simbody\sharedTarget\SimTKsimbody.vcxproj]
D:\a\simbody\simbody\Simbody\src\CableSpan.cpp(2835,37): warning C4267: 'initializing': conversion from 'size_t' to 'const int', possible loss of data [D:\a\simbody\simbody\out\build\ci-msvc-windows-release\Simbody\sharedTarget\SimTKsimbody.vcxproj]
```

adamkewley:
@pepbos fyi, the fix for MSVC is as simple as (confirmed on my machine):

```diff
diff --git a/Simbody/src/CableSpan.cpp b/Simbody/src/CableSpan.cpp
index 9d715e77..44ecb50f 100644
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -2832,7 +2832,7 @@ void CableSpan::Impl::calcSolverStep(
 
     // SOLVER STEP 2: Compute the path correction step that reduces the path
     // errors.
-    const int numObstaclesInContact = data.lineSegments.size() - 1;
+    const int numObstaclesInContact = static_cast<int>(data.lineSegments.size()) - 1;
     SimTK_ASSERT_ALWAYS(
         numObstaclesInContact > 0,
         "No obstacles in contact with the cable: unable to compute path corrections");
```

adamkewley:
As a quick first pass, the changes seem logical enough.

I only spotted minor things:

- The `switch` for the algorithm (`CableSpan.cpp:2841`) contains some quite big blocks that could be cut down a bit.
- There's some `std::vector` uses that might benefit from a `reserve` before filling them (e.g. `CableSpan.cpp:3234/3301`). Same goes for switching out a value-initialized initialized allocation with a blank `.reserve`d one (e.g. `CableSpan.cpp:3179`) because later code `clear`s and uses `push/emplace_back` anyway.
- The comment strings might benefit from also including a link to this issue, and a DOI link to Scholz's original paper, so that future devs can just Ctrl+Click the link and get the full context (e.g. of the screenshots/videos in this PR)
- Any `TODO`s that aren't going to realistically be handled by the integration should be cleared from the codebase (e.g. `CableSpan.cpp::464`

None of those observations are worth blocking or holding back this PR.

It must be noted that I have no idea about the underlying maths/algorithm being implemented, so my reviews over this code is just a quick engineering vibe check. I imagine @sherm1  knows more about it algorithmically, but because there's a lot of code here, I imagine the process will be closer to "lets see how it behaves in prod" than "lets formally verify these algorithms" rollout - the video/image explanations help a lot in this regard, which is why I recommend linking to this PR from the relevant comment string (e.g. `CableSpan.h:60` or so).

I would also consider squashing the final version of this PR into a smaller number of commits (one, even).  This is not a blocking suggestion, it's just that it's unlikely that the intermediate commits are going to be useful in the future. For example, it's unlikely that a future developer will want to `git bisect` into an intermediate commit in order to diagnose a bug in prod (esp. if an intermediate commit isn't compiling on the target platform, or similar).

Nice contribution @pepbos 🎉

adamkewley:
Note: the test suite fails under MSVC with the build profile `ci-msvc-windows-release`:

```text
ci-msvc-windows-release (CableSpan_MinimalLengthCostFunction)
$ ctest -R "TestCableSpan*" --output-on-failure
Test project C:/Users/adamk/Desktop/simbody/out/build/ci-msvc-windows-release
    Start 89: TestCableSpan
1/3 Test #89: TestCableSpan ....................   Passed   15.76 sec
    Start 90: TestCableSpan_Forces
2/3 Test #90: TestCableSpan_Forces .............Exit code 0xc0000409
***Exception:   0.11 sec

    Start 91: TestCableSpan_LengthDot
3/3 Test #91: TestCableSpan_LengthDot ..........   Passed    0.01 sec

67% tests passed, 1 tests failed out of 3

Total Test time (real) =  15.90 sec

The following tests FAILED:
         90 - TestCableSpan_Forces (Exit code 0xc0000409
)
Errors while running CTest
```

But it passes in `msvc-dev-default`:

```text
msvc-dev-default (CableSpan_MinimalLengthCostFunction)
$ ctest -R "TestCableSpan*" --output-on-failure
Test project C:/Users/adamk/Desktop/simbody/out/build/msvc-dev-default
    Start 89: TestCableSpan
1/3 Test #89: TestCableSpan ....................   Passed   20.13 sec
    Start 90: TestCableSpan_Forces
2/3 Test #90: TestCableSpan_Forces .............   Passed    0.01 sec
    Start 91: TestCableSpan_LengthDot
3/3 Test #91: TestCableSpan_LengthDot ..........   Passed    0.02 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =  20.19 sec
```

adamkewley:
@pepbos It's a little bit tricky to play around with the failing build because it doesn't have debug symbols, but a little bit of `__debugbreak()` bisecting indicates that the exception occurs in `testCableForceOnSingleObstacle()` when it calls `assertForces`. I guess that the issue is that the implementation isn't getting the expected forces. I wrote some debug printing into the test suite and this is what it looks like:

```text
[RUN ] testCableForceOnSameBody
    [OK  ] force = ~[~[0,0,0],~[0,0,0]] error = 0
[OK  ] testCableForceOnSameBody
[RUN ] testCableForceBetweenTwoBodies
    [OK  ] force = ~[~[0,0,0],~[1,0,0]] error = 0
    [OK  ] force = ~[~[0,0,0],~[-1,0,0]] error = 0
[OK  ] testCableForceBetweenTwoBodies
[RUN ] testCableForceAndMomentBetweenTwoBodies
    [OK  ] force = ~[~[0,0,0],~[1,0,0]] error = 0
    [OK  ] force = ~[~[0,0,2],~[-1,0,0]] error = 0
[OK  ] testCableForceAndMomentBetweenTwoBodies
[RUN ] testCableForceOnSingleObstacle
    [OK  ] force = ~[~[-1.10063e-22,6.57277e-23,0],~[1.41421,0,-1.58681e-22]] error = 2.03994e-22
    [FAIL] force = ~[~[2.35099e-38,-2.35099e-38,0],~[-1.41421,7.80699e-11,6.57277e-23]] error = 1
```

adamkewley:
Ah, after printing the working version, I think I have found the problem.

```diff
-forcesExpected[1] = SpatialVec(Vec3{0., 0., 0.}, Vec3{-2. * cos(angle), 0.});
+forcesExpected[1] = SpatialVec(Vec3{0., 0., 0.}, Vec3{-2. * cos(angle), 0., 0.});
```

Because `Vec` is uninitialized in simbody in release builds

pepbos:
A classic! And still hard! Thanks @adamkewley for tracking it down!

sherm1 (approved):
Thanks @adamkewley for an in-depth review. I followed with a shallow review. Did find a typo! Otherwise, :lgtm_strong:
Awesome functionality, @pepbos -- thanks so much!

It occurs to me that we really ought to have an Example program that shows how to use this and does something compelling. We have the old ExampleCableSpan but it doesn't use the new algorithm. Consider modifying that (in a separate PR) or making a new example that shows off the new algorithm more.

@adamkewley @pepbos LMK when this is ready to merge.

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/814#-:-OKOYc0rCZlyYWbqn1zT:b-gk3vtz)* status: all files reviewed, 1 unresolved discussion (waiting on @pepbos)

___
*[`Simbody/src/CableSpan.cpp` line 2619 at r1](https://reviewable.io/reviews/simbody/simbody/814#-OKOX2fAAKXsZDIE2sZz:-OKOX2fAAKXsZDIE2s_-:b-f2nvbr) ([raw file](https://github.com/simbody/simbody/blob/f8346824fe2849f25b6034ffb1f4010d464233af/Simbody/src/CableSpan.cpp#L2619)):*
> ```cpp
>             const Real rDot = dataInst.jacobiDot_Q[1];
> 
>             // If we assume the variation of the jaobi fields is zero, then the
> ```

typo: jaobi -> jacobi

adamkewley:
@sherm1 , this should now be ready to merge

PR #815, opened 2025-03-05
Remove legacy operators that are causing problems to swig

aymanhab:
These operators are within ifdef that suggest all modern compilers don't need them, so this likely has no effect on simbody clients, except that using these headers to generate bindings in swig causes problems as swig 4.2+ generates new types and code that fails to resolve these operators. 
I will remove the ifdef block completely before merging if approved.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/815)

### DIFF
--- a/SimTKcommon/include/SimTKcommon/internal/common.h
+++ b/SimTKcommon/include/SimTKcommon/internal/common.h
@@ -635,6 +635,7 @@ struct Segment {
 
 // With compiler support for <=> operator these four methods below don't
 // need to be manually defined. (from gcc 10, clang 10 & msvc 19.22)
+// also exclude if SWIG_PYTHON is defined to build python bindings in OpenSim
 #if !defined(__cpp_lib_three_way_comparison) && !defined SWIG_PYTHON
 // These next four methods supply the missing relational operators for any
 // types L and R where L==R and L<R have been defined. This is like the
--- a/SimTKcommon/include/SimTKcommon/internal/common.h
+++ b/SimTKcommon/include/SimTKcommon/internal/common.h
@@ -635,7 +635,7 @@ struct Segment {
 
 // With compiler support for <=> operator these four methods below don't
 // need to be manually defined. (from gcc 10, clang 10 & msvc 19.22)
-#if 0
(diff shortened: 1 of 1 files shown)


sherm1 (commented):
Hi, Ayman. `__cpp_lib_three_way_comparison` was first defined for C++20, so I suspect these operators are needed to compile Simbody on anything earlier than that. I don't think we've required people to be at that level yet for Simbody.

As an alternative, doesn't SWIG define a macro that allows for excluding some code from its purview? We could wrap the current ifdef with that so SWIG doesn't have to deal with those operators, but leave them in for old compilers? WDYT?

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/815#-:-OKcEQN006djcCowOi8T:bvfoldt)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @adamkewley)

adamkewley:
Obviously, I love removing unnecessary code ❤️ , but @sherm1 is correct: if you dump this change through an older compiler that hasn't set the macro then any code in the `SimTK` namespace (or a compilation unit with `using namespace SimTK;`) will probably fail to compile with a "no overload > available"-type message.

I don't know for certain whether C++ feature flags are set in the case where the language level is low (e.g. you configured the compiler to use C++11) but the compiler version is high (e.g. it could support C++20).

aymanhab:
Thanks @adamkewley and @sherm1 I restored the code block and added another condition that we use on OpenSim side so that building python bindings in OpenSim succeeds. The error was showing when compiling swig generated code due to ambiguity resolving the operators for some swig generated classes.

adamkewley:
Seems reasonable enough 👍

sherm1 (approved):
:lgtm:, thanks Ayman!

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/815#-:-OLROKLs9uwElnUIKmsi:bt8m4dt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

aymanhab:
Thanks much @sherm1 and @adamkewley

PR #828, opened 2025-07-09
CableSpan: Via points

### LINKED ISSUE #827: Add via points to `CableSpan`
### Summary

Support for `CableSpan` should be incorporated into OpenSim as an option to model muscle geometry, but many OpenSim models rely on via points which `CableSpan` currently lacks. One option would be to connect a series of `CableSpan` elements together in OpenSim, but it seems more appropriate (and likely more efficient) to add via point support directly to `CableSpan`.  

The following represents a proposed implementation for via point support in `CableSpan`.

### Via point definition

Via points would be handled based on the order that they are added to the `CableSpan`. Any obstacles present in the path would respect via points added before or after the obstacle during path calculations. This means that, for example, the following two paths with two obstacles and one via point would have different wrapping behavior:

- Path 1: origin - via point - obstacle - obstacle - termination
- Path 2: origin - obstacle - via point - obstacle - termination

The key difference between these two paths is that Path 1 would have off diagonal entries in the path-error Jacobian representing coupling matrices due to adjacent geodesics, whereas Path 2 interrupts the two obstacles which means the coupling matrices would be zero. Path 2 would essentially treat the via point as the origin for the latter grouping of segments in the `CableSpan`. Therefore, similarly to obstacles, new line segments must be added to the `CableSpan` for each added via point.

### Interface

A via point is a station rigidly attached to `MobilizedBody` that a `CableSpan` may freely slide through. The interface for adding a via point to `CableSpan` could mirror that of adding an obstacle:

```c++
/** Add a via point to the cable's path.
@param viaPointBody The body that the via point is rigidly attached to.
@param point The via point's position in body fixed coordinates.
@return The index of the added via point in this cable. **/
CableSpanViaPointIndex addViaPoint(
    MobilizedBodyIndex viaPointBody, 
    const Vec3& point_B);

/** Get the number of via points added to the path. **/
int getNumViaPoints() const;
```

Accessors for setting and getting the via point properties could also mirror that of obstacles:

```c++
  /** @name Via point configuration */
  ///@{

  /** Get the index of the mobilized body that the via point is attached to.
  @param ix The index of the via point in this CableSpan.
  @return The index of the mobilized body that the via point is attached to. **/
  const MobilizedBodyIndex& getViaPointMobilizedBodyIndex(
      CableSpanViaPointIndex ix) const;

  /** Set the index of the mobilized body that the via point is attached to.
  @param ix The index of the via point in this CableSpan.
  @param body The index of the mobilized body that the via point is attached
  to. **/
  void setViaPointMobilizedBodyIndex(
      CableSpanViaPointIndex ix,
      MobilizedBodyIndex body);

  /** Get the station of the via point in body fixed coordinates.
  @param ix The index of the via point in this CableSpan.
  @return The station of the via point in body fixed coordinates. **/
  Vec3 getViaPointStation(CableSpanViaPointIndex ix) const;

  /** Set the station of the via point in body fixed coordinates.
  @param ix The index of the via point in this CableSpan.
  @param station The station of the via point in body fixed coordinates. **/
  void setViaPointStation(
      CableSpanViaPointIndex ix,
      const Vec3& station_B);

  ///@}
```

### Implementation details

We should be able to add support for via points with minimal changes to the existing wrapping calculations. The main challenge is bookkeeping, as we must maintain ordered lists of the current obstacles and via points in the `CableSpan` and a master list of indices to both obstacles and via points in the order that they existing in the path. 

To that end, we can introduce both `CableSpanSegmentIndex` and `CableSpanViaPointIndex` to supplement the existing `CableSpanObstacleIndex`. `CableSpanObstacleIndex` and `CableSpanViaPointIndex` can be used to index into the list of obstacles and via points, respectively, and `CableSpanSegmentIndex` can be used to iterate through all obstacles and via points in the path, in order. The following enum and helper struct can be used to access either the internal `ViaPoint` or `CurveSegment` (associated with an obstacle) at the current `CableSpanSegmentIndex`.

```c++
enum class SegmentType {
    CurveSegment,
    ViaPoint      
};

struct SegmentInfo {
public:
    SegmentInfo(SegmentType t, SegmentIndex i, ObstacleIndex o) 
        : m_type(t), m_index(i), m_obstacleIndex(o) {
            SimTK_ERRCHK_ALWAYS(
                t == SegmentType::CurveSegment,
                "SegmentInfo::SegmentInfo()",
                "Obstacle index must be provided for obstacle segments.");
            SimTK_ERRCHK_ALWAYS(
                m_viaPointIndex == ViaPointIndex::Invalid(),
                "SegmentInfo::SegmentInfo()",
                "SegmentInfo already assigned a via point index.");
        }

    SegmentInfo(SegmentType t, SegmentIndex i, ViaPointIndex v) 
        : m_type(t), m_index(i), m_viaPointIndex(v) {
            SimTK_ERRCHK_ALWAYS(
                t == SegmentType::ViaPoint,
                "SegmentInfo::SegmentInfo()",
                "Via point index must be provided for via point segments.");
            SimTK_ERRCHK_ALWAYS(
                m_obstacleIndex == ObstacleIndex::Invalid(),
                "SegmentInfo::SegmentInfo()",
                "SegmentInfo already assigned an obstacle index.");
        }

    SegmentType getType() const { return m_type; }
    const SegmentIndex& getIndex() const { return m_index; }
    const ObstacleIndex& getObstacleIndex() const { return m_obstacleIndex; }
    const ViaPointIndex& getViaPointIndex() const { return m_viaPointIndex; }

private:
    SegmentType m_type;
    SegmentIndex m_index;
    ObstacleIndex m_obstacleIndex = ObstacleIndex::Invalid();
    ViaPointIndex m_viaPointIndex = ViaPointIndex::Invalid();
};

```

Note that "segment" is not an ideal term to group via points in since they have zero length, but for lack of a better term, I will use this for now.

With the new indices and helper data structures, the following modifications to `CableSpan` can be made. Note that this is not a complete list, but an overview of the key changes that will need to made.

- The helper `calcLineSegments` should be updated to add line segments according to via points in the path.
- The helper `calcPathErrorVector` should be updated to account for via points and use the correct adjacent line segments when calculating path errors.
- Similarly, the helpers`calcPathErrorJacobian`, `calcLengthGradient`, and `calcLengthHessian` should be updated to account for via points and use the correct line segments where appropriate.
- There are a few instances in`CableSpan::Impl::calcSolverStep()` where the number of line segments is used to determine whether or not path errors are calculated. This should be updated to use the number of active obstacles since via points can independently add line segments.
- `CurveSegment::findPrevObstacleInContactWithCable()` and `CurveSegment::findNextObstacleInContactWithCable()` are used to determine if off-diagonal path-error matrices are needed. Therefore, these functions should be updated to return an invalid index if a via point exists before the next obstacle in the path.
- Similar, `CurveSegment::findPrevPathPoint_G` and `CurveSegment::findNextPathPoint_G` should return the via point station if a via point exists before the next obstacle in the path.
- `CableSpan::Impl::calcDataVel()` should be updated to incorporate via points in path speed calculations. 
- The helpers `applyBodyForces` and `calcCablePower` should be updated to account for via points when computing forces, powers, etc.

### Draft and feedback

I am working on a draft PR that I will post once it is functional to facilitate discussion. All feedback on the proposed changes are welcome!

nickbianco:
Fixes #827 

## Summary

This change adds support for including via points in a CableSpan's path. Via points can be added to a CablePath by specifying a mobilized body and a body-fixed station that define the position of the via point in the multibody system. Similarly to obstacles, via points are handled in the order that they are added to the CableSpan. Internally, via points divide the cable into segments, and a separate path optimization is performed for each segment.

Here is the result of `ExampleCableSpan.cpp`, now updated to include two via points (shown in light blue) that modify the` `CableSpan`'s path. 

https://github.com/user-attachments/assets/e9dfc198-110c-4b69-aa20-cc3d31d0706e

## Interface

`CableSpan::addViaPoint()` can be used to add a via point to the cable by specifying a `MobilizedBodyIndex` and the position of the via point on the mobilized body in body-fixed coordinates. A `CableSpanViaPointIndex` is returned, providing a unique identifier for the via point in the `CableSpan`.

```c++
// Construct a new cable.
CableSpan cable(
    cables,
    cableOriginBody,
    Vec3{0.},
    cableTerminationBody,
    Vec3{0.});

// Mobilizer for the path via point.
MobilizedBody::Translation cableViaPointBody(
    matter.Ground(),
    Transform(Vec3(0., 0.9, 0.5)),
    aBody,
    Transform());

// Add the via point.
CableSpanViaPointIndex ix = cable.addViaPoint(cableViaPointBody, Vec3{0.});
```

Accessors have been added to enable retrieving or modifying the mobilized body or via point location, and the convenience method `CableSpan::calcViaPointLocation()` is added for calculating the location of a via point in the ground frame.

## Implementation

Internally, two new classes are added to support via points in a `CableSpan` path: `ViaPoint` and `CableSegment`. `ViaPoint` caches computations relevant to via points and provides convenience methods similar to `CurveSegment` for path obstacles. `CableSegment` is added to manage groups of obstacles in the cable that are separated by via points. Within `CableSpan.cpp`, `CableSegmentIndex` provides a unique index to `CableSegement`s, and `CableSpanViaPointIndex` can be used to index into `ViaPoint`s.

A key aspect of this implementation is that the path segments associated with each `CableSegment` are optimized independently, since obstacles separated by via points have no interaction when natural geodesic corrections are applied (i.e., the path error Jacobian off-diagonal elements are structurally zero). This approach is advantageous because path segments may converge in a different number of solver iterations and any invalidated segments will not force unchanged paths to re-run the path optimization unnecessarily. 

## Testing

`TestCableSpan.cpp` has been updated to add the new subtest `testViaPoints()` and include via points in the subtest `testAllSurfaceKinds()`. `testViaPoints()` mimics `testSimpleCable()` by defining a `CableSpan` comprised only of via points from which it is easy to verify both the path geometry and applied body forces:

TestCableSpanViaPoints

Additionally, tests have been added to `TestCableSpan_Forces.cpp` and `TestCableSpan_LengthDot.cpp` mimicking the existing obstacle tests but instead using via points.

### CableSubsystem accessors

To help support a wrapping implementation in OpenSim based on `CableSpan`, accessors to `CableSubsystem` have been added to `MultibodySystem`. `ModelComponent` in OpenSim would now have access to `CableSubsystem` (and subsequently any underlying `CableSpan` elements) through it's stored reference to `MultibodySystem`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/828)

### DIFF
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -6,7 +6,7 @@
 -------------------------------------------------------------------------------
  Copyright (c) 2024 Authors.
  Authors: Pepijn van den Bos
- Contributors:
+ Contributors: Nicholas Bianco
 
  Licensed under the Apache License, Version 2.0 (the "License"); you may
  not use this file except in compliance with the License. You may obtain a
@@ -26,9 +26,14 @@ namespace SimTK
 {
 
 /** @class SimTK::CableSpanObstacleIndex
-This is a unique integer type for identifying obstacles comprising a particular
+This is a unique integer type for identifying obstacles within a particular
 CableSpan. These begin at zero for each CableSpan. **/
 SimTK_DEFINE_UNIQUE_INDEX_TYPE(CableSpanObstacleIndex);
+
+/** @class SimTK::CableSpanViaPointIndex
+This is a unique integer type for identifying via points within a particular
+CableSpan. These begin at zero for each CableSpan. **/
+SimTK_DEFINE_UNIQUE_INDEX_TYPE(CableSpanViaPointIndex);
 
 /** @class SimTK::CableSpanIndex
 This is a unique integer type for quickly identifying specific cables for fast
@@ -120,17 +125,18 @@ class CableSubsystemTestHelper;
 
 //==============================================================================
 /** This class represents the path of a frictionless cable from an origin point
-fixed to a body, over geometric obstacles fixed to other bodies, to a final
-termination point.
+fixed to a body, over geometric obstacles and through via points fixed to other 
+bodies, to a final termination point.
 
 The CableSpan's path can be seen as consisting of straight line segments and
 curved line segments: A curved segment over each obstacle, and straight segments
-connecting them to each other and to the end points. Each curved segment is
-computed as a geodesic to give (in some sense) a shortest path over the surface.
-During a simulation the cable can slide freely over the obstacle surfaces. It
-can lose contact with a surface and miss that obstacle in a straight line.
-Similarly the cable can touchdown on the obstacle if the surface obstructs the
-straight line segment again.
+connecting them to each other and to via points and the end points. Each curved 
+segment is computed as a geodesic to give (in some sense) a shortest path over 
+the surface. During a simulation the cable can slide freely over the obstacle 
+surfaces. It can lose contact with a surface and miss that obstacle in a 
+straight line. Similarly the cable can touchdown on the obstacle if the surface 
+obstructs the straight line segment again. The cable will always slide freely 
+through any via points present in the path.
 
 The path is computed as an optimization problem using the previous optimal path
 as the warm start. This is done by computing natural geodesic corrections for
@@ -156,6 +162,13 @@ added to a cable, then, the cable can wrap over the first, then the second, and
 then the third. If the first obstacle spatially collides with the path twice it
 will not actually wrap over it twice. Use CablePath if this is not the desired
 behavior.
+
+When via points are present in the cable, the cable is internally divided into 
+cable segments separated by the via points. Obstacles separated by via points 
+have no interaction when geodesic corrections are applied during optimization,
+and therefore the path for each cable segment is solved independently. This 
+approach is advantageous since cable segments with different sets of wrap 
+obstacles may require a different number of optimization iterations to converge.
 
 Note that a CableSpan is a geometric object, not a force or constraint element.
 That is, a CableSpan alone will not influence the behavior of a simulation.
@@ -232,6 +245,20 @@ public:
     /** Get the number of obstacles added to the path. **/
     int getNumObstacles() const;
 
+    /** Add a via point that the cable's path must pass through.
+    @param viaPointBody The body that the via point is rigidly attached to.
+    @param station_B The via point's station in body fixed coordinates.
+    @return The index of the added via point in this cable. **/
+    CableSpanViaPointIndex addViaPoint(
+        MobilizedBodyIndex viaPointBody, 
+        const Vec3& station_B);
+
+    /** Get the number of via points added to the path. **/
+    int getNumViaPoints() const;
+
+    /** Get the cable's index in the CableSubsystem. **/
+    CableSpanIndex getIndex() const;
+
     ///@}
 
     /** @name End points configuration */
@@ -259,13 +286,36 @@ public:
     /** Set the cable's origin point defined in body fixed coordinates. **/
     void setOriginStation(const Vec3& originStation);
 
-    /** Get the cable's termination point defined in body fixed coordinates.
-    **/
+    /** Get the cable's termination point defined in body fixed coordinates. **/
     Vec3 getTerminationStation() const;
 
-    /** Set the cable's termination point defined in body fixed coordinates.
-    **/
+    /** Set the cable's termination point defined in body fixed coordinates. **/
     void setTerminationStation(const Vec3& terminationStation);
+
+    ///@}
+
+    /** @name End point computations */
+    ///@{
+
+    /** Calculate the unit force exerted by the cable at the cable origin (in
+    ground frame). The force can be obtained by multiplying the result by the
+    cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    @param[out] unitForce_G The resulting unit spatial force in ground frame. 
+    **/
+    void calcOriginUnitForce(
+        const State& state,
+        SpatialVec& unitForce_G) const;
+
+    /** Calculate the unit force exerted by the cable at the cable termination
+    (in ground frame). The force can be obtained by multiplying the result by
+    the cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    @param[out] unitForce_G The resulting unit spatial force in ground frame. 
+    **/
+    void calcTerminationUnitForce(
+        const State& state,
+        SpatialVec& unitForce_G) const;
 
     ///@}
 
@@ -335,6 +385,38 @@ public:
 
     ///@}
 
+    /** @name Via point configuration */
+    ///@{
+
+    /** Get the index of the mobilized body that the via point is attached to.
+    @param ix The index of the via point in this CableSpan.
+    @return The index of the mobilized body that the via point is attached to. 
+    **/
+    const MobilizedBodyIndex& getViaPointMobilizedBodyIndex(
+        CableSpanViaPointIndex ix) const;
+
+    /** Set the index of the mobilized body that the via point is attached to.
+    @param ix The index of the via point in this CableSpan.
+    @param body The index of the mobilized body that the via point is attached
+    to. **/
+    void setViaPointMobilizedBodyIndex(
+        CableSpanViaPointIndex ix,
+        MobilizedBodyIndex body);
+
+    /** Get the station of the via point in body fixed coordinates.
+    @param ix The index of the via point in this CableSpan.
+    @return The station of the via point in body fixed coordinates. **/
+    Vec3 getViaPointStation(CableSpanViaPointIndex ix) const;
+
+    /** Set the station of the via point in body fixed coordinates.
+    @param ix The index of the via point in this CableSpan.
+    @param station_B The station of the via point in body fixed coordinates. **/
+    void setViaPointStation(
+        CableSpanViaPointIndex ix,
+        const Vec3& station_B); 
+
+    ///@}
+
     /** @name Solver configuration */
     ///@{
 
@@ -359,7 +441,9 @@ public:
     void setSolverMaxIterations(int maxIterations);
 
     /** Number of solver iterations used to compute the current cable's path.
-    State must be realized to Stage::Position.
+    State must be realized to Stage::Position. If via points are present in the
+    cable, the sum of the iterations used to compute the path for each cable
+    segment is returned.
     @param state System State.
     @return Number of solver iterations used. **/
     int getNumSolverIterations(const State& state) const;
@@ -378,6 +462,8 @@ public:
     void setSmoothnessTolerance(Real tolerance);
 
     /** The smoothness of the current cable's path.
+    If via points are present in the cable, the maximum smoothness error across
+    all cable segments is returned.
     See CableSpan::getSmoothnessTolerance.
     State must be realized to Stage::Position. **/
     Real getSmoothness(const State& state) const;
@@ -504,6 +590,41 @@ public:
         int numSamples,
         const std::function<void(Vec3 point_G)>& sink) const;
 
+    /** Calculate the unit force exerted by the cable at the specified obstacle
+    (in ground frame). The force can be obtained by multiplying the result by
+    the cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    @param ix The index of the obstacle in this CableSpan.
+    @param[out] unitForce_G The resulting unit spatial force in ground frame. **/
+    void calcCurveSegmentUnitForce(
+        const State& state,
+        CableSpanObstacleIndex ix,
+        SpatialVec& unitForce_G) const;
+
+
+    /** @name Via point computations */
+    ///@{
+
+    /** Compute the location of a via point in the ground frame.
+    State must be realized to Stage::Position.
+    @param state State of the system.
+    @param ix The index of the via point in this CableSpan.
+    @return The location of the via point in the ground frame. **/
+    Vec3 calcViaPointLocation(
+        const State& state, 
+        CableSpanViaPointIndex ix) const;
+
+    /** Calculate the unit force exerted by the cable at the specified via point
+    (in ground frame). The force can be obtained by multiplying the result by
+    the cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    @param ix The index of the via point in this CableSpan.
+    @param[out] unitForce_G The resulting unit spatial force in ground frame. **/
+    void calcViaPointUnitForce(
+        const State& state,
+        CableSpanViaPointIndex ix,
+        SpatialVec& unitForce_G) const;
+
     ///@}
 
     /** @cond **/ // Hide from Doxygen.
@@ -607,9 +728,8 @@ public:
     ~CableSubsystemTestHelper() noexcept;
     CableSubsystemTestHelper(const CableSubsystemTestHelper&);
     CableSubsystemTestHelper& operator=(const CableSubsystemTestHelper&);
-    CableSubsystemTestHelper(CableSubsystemTestHelper&&) noexcept = default;
-    CableSubsystemTestHelper& operator=(CableSubsystemTestHelper&&) noexcept =
-        default;
+    CableSubsystemTestHelper(CableSubsystemTestHelper&&) noexcept;
+    CableSubsystemTestHelper& operator=(CableSubsystemTestHelper&&) noexcept;
 
     /** Test the paths of all CableSpan registered in a CableSubsystem.
     State must be realized to Stage::Position.
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -125,17 +125,17 @@ class CableSubsystemTestHelper;
 
 //==============================================================================
 /** This class represents the path of a frictionless cable from an origin point
-fixed to a body, over geometric obstacles and through via points fixed to other 
+fixed to a body, over geometric obstacles and through via points fixed to other
 bodies, to a final termination point.
 
 The CableSpan's path can be seen as consisting of straight line segments and
 curved line segments: A curved segment over each obstacle, and straight segments
-connecting them to each other and to via points and the end points. Each curved 
-segment is computed as a geodesic to give (in some sense) a shortest path over 
-the surface. During a simulation the cable can slide freely over the obstacle 
-surfaces. It can lose contact with a surface and miss that obstacle in a 
-straight line. Similarly the cable can touchdown on the obstacle if the surface 
-obstructs the straight line segment again. The cable will always slide freely 
+connecting them to each other and to via points and the end points. Each curved
+segment is computed as a geodesic to give (in some sense) a shortest path over
+the surface. During a simulation the cable can slide freely over the obstacle
+surfaces. It can lose contact with a surface and miss that obstacle in a
+straight line. Similarly the cable can touchdown on the obstacle if the surface
+obstructs the straight line segment again. The cable will always slide freely
 through any via points present in the path.
 
 The path is computed as an optimization problem using the previous optimal path
@@ -163,11 +163,11 @@ then the third. If the first obstacle spatially collides with the path twice it
 will not actually wrap over it twice. Use CablePath if this is not the desired
 behavior.
 
-When via points are present in the cable, the cable is internally divided into 
-cable segments separated by the via points. Obstacles separated by via points 
+When via points are present in the cable, the cable is internally divided into
+cable segments separated by the via points. Obstacles separated by via points
 have no interaction when geodesic corrections are applied during optimization,
-and therefore the path for each cable segment is solved independently. This 
-approach is advantageous since cable segments with different sets of wrap 
+and therefore the path for each cable segment is solved independently. This
+approach is advantageous since cable segments with different sets of wrap
 obstacles may require a different number of optimization iterations to converge.
 
 Note that a CableSpan is a geometric object, not a force or constraint element.
@@ -250,7 +250,7 @@ public:
     @param station_B The via point's station in body fixed coordinates.
     @return The index of the added via point in this cable. **/
     CableSpanViaPointIndex addViaPoint(
-        MobilizedBodyIndex viaPointBody, 
+        MobilizedBodyIndex viaPointBody,
         const Vec3& station_B);
 
     /** Get the number of via points added to the path. **/
@@ -301,7 +301,7 @@ public:
     ground frame). The force can be obtained by multiplying the result by the
     cable tension. State must be realized to Stage::Position.
     @param state State of the system.
-    @param[out] unitForce_G The resulting unit spatial force in ground frame. 
+    @param[out] unitForce_G The resulting unit spatial force in ground frame.
     **/
     void calcOriginUnitForce(
         const State& state,
@@ -311,7 +311,7 @@ public:
     (in ground frame). The force can be obtained by multiplying the result by
     the cable tension. State must be realized to Stage::Position.
     @param state State of the system.
-    @param[out] unitForce_G The resulting unit spatial force in ground frame. 
+    @param[out] unitForce_G The resulting unit spatial force in ground frame.
     **/
     void calcTerminationUnitForce(
         const State& state,
@@ -390,7 +390,7 @@ public:
 
     /** Get the index of the mobilized body that the via point is attached to.
     @param ix The index of the via point in this CableSpan.
-    @return The index of the mobilized body that the via point is attached to. 
+    @return The index of the mobilized body that the via point is attached to.
     **/
     const MobilizedBodyIndex& getViaPointMobilizedBodyIndex(
         CableSpanViaPointIndex ix) const;
@@ -413,7 +413,7 @@ public:
     @param station_B The station of the via point in body fixed coordinates. **/
     void setViaPointStation(
         CableSpanViaPointIndex ix,
-        const Vec3& station_B); 
+        const Vec3& station_B);
 
     ///@}
 
@@ -611,7 +611,7 @@ public:
     @param ix The index of the via point in this CableSpan.
     @return The location of the via point in the ground frame. **/
     Vec3 calcViaPointLocation(
-        const State& state, 
+        const State& state,
         CableSpanViaPointIndex ix) const;
 
     /** Calculate the unit force exerted by the cable at the specified via point
--- a/Simbody/include/simbody/internal/MultibodySystem.h
+++ b/Simbody/include/simbody/internal/MultibodySystem.h
@@ -35,7 +35,7 @@ class SimbodyMatterSubsystem;
 class ForceSubsystem;
 class DecorationSubsystem;
 class GeneralContactSubsystem;
-
+class CableSubsystem;
 
 /** The job of the MultibodySystem class is to coordinate the activities of 
 various subsystems which can be part of a multibody system. We insist on 
@@ -68,6 +68,10 @@ public:
     GeneralContactSubsystem&       updContactSubsystem();
     bool hasContactSubsystem() const;
 
+    int setCableSubsystem(CableSubsystem&);
+    const CableSubsystem& getCableSubsystem() const;
+    CableSubsystem&       updCableSubsystem();
+    bool hasCableSubsystem() const;
 
     /// Calculate the total potential energy of the system.  The state must
     /// be at Dynamics stage or later.
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -28,7 +28,7 @@
 using namespace SimTK;
 
 // Define a unique index for cable segments.
-namespace { 
+namespace {
 SimTK_DEFINE_UNIQUE_INDEX_TYPE(CableSegmentIndex);
 }
 
@@ -625,7 +625,7 @@ private:
 };
 
 //==============================================================================
-//                          Struct Curve Segment
+//                               Curve Segment
 //==============================================================================
 /* The CableSpan's path consists of straight line segments between obstacles and
 via points, and curved segments over the obstacles. This struct represents the
@@ -810,23 +810,21 @@ public:
 
     const CableSubsystem& getSubsystem() const
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "CurveSegment::getSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "CurveSegment::getSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
     CableSubsystem& updSubsystem()
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "CurveSegment::updSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "CurveSegment::updSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
@@ -868,7 +866,7 @@ public:
             getDataInst(state).X_SQ.p());
     }
 
-    bool isInContactWithSurface(const State& state) const 
+    bool isInContactWithSurface(const State& state) const
     {
         const CurveSegmentData::Instance& dataInst = getDataInst(state);
         return dataInst.wrappingStatus ==
@@ -883,11 +881,11 @@ public:
     // contact with the cable.
 
     // Find the first obstacle before this segment that is in contact with the
-    // cable. Returns an invalid index if there is none, e.g., the previous 
+    // cable. Returns an invalid index if there is none, e.g., the previous
     // point in the path is a via point or the path origin.
     ObstacleIndex findPrevObstacleInContactWithCable(const State& state) const;
     // Find the first obstacle after this segment that is in contact with the
-    // cable. Returns an invalid index if there is none, e.g., the next point 
+    // cable. Returns an invalid index if there is none, e.g., the next point
     // in the path is a via point or the path termination.
     ObstacleIndex findNextObstacleInContactWithCable(const State& state) const;
 
@@ -921,7 +919,7 @@ public:
             "CurveSegment::Impl::assertSurfaceBounds",
             "Next point lies inside the surface");
         return nextPoint_S;
-    } 
+    }
 
     //--------------------------------------------------------------------------
     //  Updating CurveSegmentData::Instance Helpers
@@ -1456,7 +1454,7 @@ private:
     ObstacleIndex m_obstacleIndex = ObstacleIndex::Invalid();
 
     // MobilizedBody that the curve segment is attached to.
-    MobilizedBodyIndex m_body;
+    MobilizedBodyIndex m_body = MobilizedBodyIndex::Invalid();
     // Surface to body transform.
     Transform m_X_BS;
 
@@ -1476,7 +1474,7 @@ private:
 };
 
 //==============================================================================
-//                          Struct Via Point
+//                                 Via Point
 //==============================================================================
 /* This struct represents a zero-length cable segment associated with a via
 point in the path. It manages the cached data such as the position of the via
@@ -1495,7 +1493,7 @@ public:
         CableSpanIndex cableIndex,
         ViaPointIndex viaPointIndex,
         MobilizedBodyIndex body,
-        Vec3 station_B) :
+        const Vec3& station_B) :
         m_subsystem(subsystem),
         m_cableIndex(cableIndex),
         m_viaPointIndex(viaPointIndex),
@@ -1514,17 +1512,17 @@ public:
             state,
             Stage::Position,
             Stage::Infinity,
-            new Value<Vec3>());
+            new Value<Vec3>(Vec3(0.)));
         m_indexIncomingDirection = updSubsystem().allocateCacheEntry(
             state,
             Stage::Position,
             Stage::Infinity,
-            new Value<UnitVec3>());
+            new Value<UnitVec3>(UnitVec3(NaN)));
         m_indexOutgoingDirection = updSubsystem().allocateCacheEntry(
             state,
             Stage::Position,
             Stage::Infinity,
-            new Value<UnitVec3>());
+            new Value<UnitVec3>(UnitVec3(NaN)));
     }
 
     void invalidateDirectionCacheEntries(const State& state) const
@@ -1548,7 +1546,7 @@ public:
     }
 
     void setIncomingDirection(
-        const State& state, 
+        const State& state,
         const UnitVec3& direction_G) const
     {
         updIncomingDirection(state) = direction_G;
@@ -1562,7 +1560,7 @@ public:
     }
 
     void setOutgoingDirection(
-        const State& state, 
+        const State& state,
         const UnitVec3& direction_G) const
     {
         updOutgoingDirection(state) = direction_G;
@@ -1593,23 +1591,21 @@ public:
 
     const CableSubsystem& getSubsystem() const
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "ViaPoint::getSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "ViaPoint::getSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
     CableSubsystem& updSubsystem()
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "ViaPoint::updSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "ViaPoint::updSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
@@ -1636,7 +1632,7 @@ public:
     //--------------------------------------------------------------------------
     // Computing cached data
     //--------------------------------------------------------------------------
-    
+
     const Vec3& calcStation_G(const State& state) const
     {
         Vec3& station_G = updStation_G(state);
@@ -1690,9 +1686,9 @@ private:
     ViaPointIndex m_viaPointIndex = ViaPointIndex::Invalid();
 
     // MobilizedBody that the via point is attached to.
-    MobilizedBodyIndex m_body;
+    MobilizedBodyIndex m_body = MobilizedBodyIndex::Invalid();
     // The position of the via point in the body frame.
-    Vec3 m_station_B;
+    Vec3 m_station_B{NaN};
 
     // Topology cache.
     CacheEntryIndex m_indexStation_G = CacheEntryIndex::Invalid();
@@ -1701,13 +1697,13 @@ private:
 };
 
 //==============================================================================
-//                          Struct Cable Segment
+//                              Cable Segment
 //==============================================================================
 /* This is a helper struct for managing segments of the cable path that are
-separated by via points. It stores an ordered list of the obstacle indexes 
+separated by via points. It stores an ordered list of the obstacle indexes
 associated with this segment and indexes to the initial and final via points, if
-they exist. An invalid initial via point index indicates that the segemnt begins
-with the cable origin point. Similarly, an invalid final via point index 
+they exist. An invalid initial via point index indicates that the segment begins
+with the cable origin point. Similarly, an invalid final via point index
 indicates that the segment ends with the cable termination point. **/
 class CableSegment {
 public:
@@ -1724,23 +1720,21 @@ public:
 
     const CableSubsystem& getSubsystem() const
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "CableSegment::getSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "CableSegment::getSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
     CableSubsystem& updSubsystem()
     {
-        if (!m_subsystem) {
-            SimTK_ERRCHK_ALWAYS(
-                m_subsystem,
-                "CableSegment::updSubsystem()",
-                "CableSpan not yet adopted by any CableSubsystem");
-        }
+        SimTK_ERRCHK_ALWAYS(
+            m_subsystem,
+            "CableSegment::updSubsystem()",
+            "CableSpan not yet adopted by any CableSubsystem");
+
         return *m_subsystem;
     }
 
@@ -1796,7 +1790,7 @@ private:
     CableSpanIndex m_cableIndex = CableSpanIndex::Invalid();
     // The index of this CableSegment in the CableSpan.
     CableSegmentIndex m_index = CableSegmentIndex::Invalid();
-    
+
     // The indexes to the wrap obstacles belonging to this segment.
     Array_<ObstacleIndex> m_obstacleIndexes;
 
@@ -1953,27 +1947,27 @@ public:
 
     const CurveSegment& getObstacleCurveSegment(ObstacleIndex ix) const
     {
-        return m_curveSegments[ix];
+        return m_curveSegments.at(ix);
     }
 
     CurveSegment& updObstacleCurveSegment(ObstacleIndex ix)
     {
-        return m_curveSegments[ix];
+        return m_curveSegments.at(ix);
     }
 
     const ViaPoint& getViaPoint(ViaPointIndex ix) const
     {
-        return m_viaPoints[ix];
+        return m_viaPoints.at(ix);
     }
 
     ViaPoint& updViaPoint(ViaPointIndex ix)
     {
-        return m_viaPoints[ix];
+        return m_viaPoints.at(ix);
     }
 
     const CableSegment& getCableSegment(CableSegmentIndex ix) const
     {
-        return m_cableSegments[ix];
+        return m_cableSegments.at(ix);
     }
 
     CableSpanParameters& updParameters()
@@ -2043,20 +2037,20 @@ public:
     // Helper functions: CableSegment initial and final points.
     //--------------------------------------------------------------------------
 
-    Vec3 calcInitialCableSegmentPointInGround(const State& s, 
+    Vec3 calcInitialCableSegmentPointInGround(const State& s,
         const CableSegment& cableSegment) const
     {
-        if (cableSegment.getInitialViaPointIndex().isValid()) { 
+        if (cableSegment.getInitialViaPointIndex().isValid()) {
             return getViaPoint(cableSegment.getInitialViaPointIndex())
                                 .getStation_G(s);
         }
         return calcOriginPointInGround(s);
     }
 
-    Vec3 calcFinalCableSegmentPointInGround(const State& s, 
+    Vec3 calcFinalCableSegmentPointInGround(const State& s,
         const CableSegment& cableSegment) const
     {
-        if (cableSegment.getFinalViaPointIndex().isValid()) {   
+        if (cableSegment.getFinalViaPointIndex().isValid()) {
             return getViaPoint(cableSegment.getFinalViaPointIndex())
                                 .getStation_G(s);
         }
@@ -2067,7 +2061,7 @@ public:
     // Helper functions: ViaPoint direction cache entries.
     //--------------------------------------------------------------------------
 
-    void invalidateViaPointDirectionCacheEntries(const State& s, 
+    void invalidateViaPointDirectionCacheEntries(const State& s,
         const CableSegment& cableSegment) const
     {
         if (cableSegment.getInitialViaPointIndex().isValid()) {
@@ -2092,6 +2086,11 @@ public:
     {
         invalidateTopology();
 
+        SimTK_ASSERT(
+            !m_cableSegments.empty(),
+            "No cable segments found. An initial cable segment should have "
+            "been added by the CableSpan constructor.");
+
         ObstacleIndex obstacleIx(m_curveSegments.size());
 
         m_curveSegments.push_back(CurveSegment(
@@ -2111,10 +2110,15 @@ public:
     }
 
     ViaPointIndex addViaPoint(
-        MobilizedBodyIndex viaPointBody, 
+        MobilizedBodyIndex viaPointBody,
         const Vec3& station_B)
     {
         invalidateTopology();
+
+        SimTK_ASSERT(
+            !m_cableSegments.empty(),
+            "No cable segments found. An initial cable segment should have "
+            "been added by the CableSpan constructor.");
 
         ViaPointIndex viaPointIx(m_viaPoints.size());
 
@@ -2122,7 +2126,7 @@ public:
             m_subsystem,
             getIndex(),
             viaPointIx,
-            viaPointBody, 
+            viaPointBody,
             station_B));
 
         // Mark the end of the last CableSegment with this ViaPoint.
@@ -2468,14 +2472,14 @@ private:
                     const MobilizedBody& mobod = curve.getMobilizedBody();
                     const CurveSegmentData::Position& curveDataPos =
                         curve.getDataPos(s);
-    
+
                     // Ending point and velocity of straight line segment.
                     const Vec3 x_GP = curveDataPos.X_GP.p();
                     const Vec3 v_GP = CalcPointVelocityInGround(mobod, x_GP);
-    
+
                     // Compute the lengthening.
                     lengthDot += dot(UnitVec3(x_GP - x_GQ), v_GP - v_GQ);
-    
+
                     // Set the next straight line starting point and velocity.
                     x_GQ = curveDataPos.X_GQ.p();
                     v_GQ = CalcPointVelocityInGround(mobod, x_GQ);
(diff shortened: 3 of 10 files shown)


adamkewley (changes requested):
Nice work @nickbianco . I think this is an excellent contribution and will certainly make it much easier for downstream projects that use via points a lot (e.g. OpenSim) to integrate `SimTK::CableSpan`s.

All-in-all, I approve of the change, most of what I spotted are either basic C++ changes or formatting/stylistic stuff that can be skipped or deferred to @sherm1 .

I can't comment on the algorithm/maths very much (that was @pepbos 's domain) but, qualitatively, it doesn't seem like these changes particularly touch the core maths (jacobians, optimizers, etc.) and the little bit of engineering maths I saw (calculating points, moments of inertia) roughly match stuff I've seen in OpenSim (and assume to be mostly true).

I haven't actually compiled+ran the examples yet, so I can't verify that this all superficially works. I will try and give it a go and get back, but don't let that hold you up.

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OV6qnsL9r7VurXpXLRw:b-2t2ljr)* status: all files reviewed, 12 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/src/CableSpan_SubsystemTestHelper_Impl.cpp` line 601 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6odhZA91vz0weFzpH:-OV6odhZA91vz0weFzpI:b82w5cr) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan_SubsystemTestHelper_Impl.cpp#L601)):*
> ```cpp
> 
> CableSubsystemTestHelper::CableSubsystemTestHelper(
>     CableSubsystemTestHelper&& source) noexcept = default;
> ```

Variable names are unnecessary in these instances.
___
*[`Simbody/tests/TestCableSpan.cpp` line 416 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6pWeN2zrhm_z51UQs:-OV6pWeN2zrhm_z51UQt:b-kqnow5) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L416)):*
> ```cpp
> 
>     // A dummy body.
>     Body::Rigid aBody(MassProperties(1., Vec3(0), Inertia(1)));
> ```

If you're going to write `1.`, why not write `0.` and `1.` in the other real-number-using cases?
___
*[`Simbody/tests/TestCableSpan.cpp` line 500 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6q2CS7_UVJWvVAOAn:-OV6q2CS7_UVJWvVAOAo:bcea8st) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L500)):*
> ```cpp
>     // Sum all straight line segment lengths.
>     const Real lengthTolerance = 1e-5;
>     const Real expectedTotalCableLength = 1. + 2. + 4. + std::sqrt(5.);
> ```

Simbody provides its own `sqrt`, which should be preferred because `std::sqrt` might not support ADL on custom floating-point types (not something that actually matters, but @sherm1 has gone through the effort of ensuring simbody could theoretically work with a user-defined `SimTK::Real` which, I assume, requires using `SimTK::` numeric functions).
___
*[`Simbody/src/MultibodySystemRep.h` line 331 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6onYmB4RJLQM6U1IJ:-OV6onYmB4RJLQM6U1IK:b-3elb27) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/MultibodySystemRep.h#L331)):*
> ```objc
>     }
>     SubsystemIndex setCableSubsystem(CableSubsystem& c) {
>         assert(!cableSub.isValid());
> ```

Use `SimTK` assertions, rather than the C library assertions, because the project-specific ones have alternatives like `ASSERT_ALWAYS` (which might be what you want) or just `ASSERT`, and the simbody project could  feasibly provide compile-time options that upgrade all `SimTK_ASSERT`s to `ALWAYS` without having to compile in debug mode (e.g. OSC has this behavior). Additionally, SimTK assertions can fail in a simbody-defined way (e.g. exception, rather than `exit`)

This pops up a few times in the PR but I won't mention it again.
___
*[`Simbody/src/CableSpan.cpp` line 1467 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6UYHb2BuEyT7z2B7a:-OV6UYHb2BuEyT7z2B7b:b-c7gwg3) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1467)):*
> ```cpp
> 
> //==============================================================================
> //                          Struct Via Point
> ```

*Class (or just drop the usage of the words "struct", e.g. "Represents a zero-legnth cable" above the class is clear enough)
___
*[`Simbody/src/CableSpan.cpp` line 1475 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Ulgf9lDVmdgn1q-A:-OV6Ulgf9lDVmdgn1q-B:b-fu0i2f) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1475)):*
> ```cpp
> class ViaPoint final {
> public:
>     //--------------------------------------------------------------------------
> ```

Naming class sections like "Constructors.", "Getters", etc. is (imo) more of an OpenSim convention.

I'm not bothered either way but it might be the case that @sherm1 has a preference.
___
*[`Simbody/src/CableSpan.cpp` line 1481 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6VmHKEjNAgOJH_JAQ:-OV6VmHKEjNAgOJH_JAR:b1jl796) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1481)):*
> ```cpp
>     ViaPoint() = default;
> 
>     ViaPoint(
> ```

This constructor does not initialize two member variables that are without a member initializer (m_body, m_station_b). Either put a deterministic member initializer on the member itself (mentioned below) or initialize it in each constructor.

Otherwise, the initial state of `ViaPoint` is undefined. In your use-case, this might not actually matter if you always call a function that initializes it but, in my experience, can cause extremely annoying-to-debug bugs when the class is changed/updated in the future in such a way that the implementation reads the undefined value to initialize some other algorithm or similar.
___
*[`Simbody/src/CableSpan.cpp` line 1501 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6WahWAaEblCaIwJaX:-OV6WahWAaEblCaIwJaY:busktje) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1501)):*
> ```cpp
>     void realizeTopology(State& state)
>     {
>         m_indexStation_G = updSubsystem().allocateCacheEntry(
> ```

I'd also recommend zero-initializing the cache entries.

I understand that reading a cache entry that's marked as invalid is bad coding downstream but (imo) you really don't want to debug a program in prod that fails randomly on a particular release/arch/build (you may not even have access to a development machine that matches the user's) because some invalid downstream code reads an undefined value.
___
*[`Simbody/src/CableSpan.cpp` line 1585 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YNFD9fYc8T_dK5jl:-OV6YNFD9fYc8T_dK5jm:b-kyguc0) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1585)):*
> ```cpp
>     {
>         if (!m_subsystem) {
>             SimTK_ERRCHK_ALWAYS(
> ```

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
___
*[`Simbody/src/CableSpan.cpp` line 1596 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YTcCBnpftCRiImo9:-OV6YTcD61yu56g8AgFG:b-kyguc0) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1596)):*
> ```cpp
>     {
>         if (!m_subsystem) {
>             SimTK_ERRCHK_ALWAYS(
> ```

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
___
*[`Simbody/src/CableSpan.cpp` line 1681 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6VKsD2ZbmBZ4B2KCM:-OV6VKsD2ZbmBZ4B2KCN:b-m962fi) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1681)):*
> ```cpp
> 
>     // MobilizedBody that the via point is attached to.
>     MobilizedBodyIndex m_body;
> ```

If the `ViaPoint` class is going to have a `ViaPoint() = default;` then I'd strongly recommend ensuring that all class members default-construct in a deterministic way. IIRC the index types and `SimTK::Vec3` trivially construct with whatever junk is on the stack/heap if they aren't explicitly initialized with a value.
___
*[`Simbody/src/CableSpan.cpp` line 1692 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YZpo20r8ODuOX1pV:-OV6YZpo20r8ODuOX1pW:b-78j9d3) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1692)):*
> ```cpp
> 
> //==============================================================================
> //                          Struct Cable Segment
> ```

Struct->class or drop
___
*[`Simbody/src/CableSpan.cpp` line 1697 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YgUi9cfiRf53aQQh:-OV6YgUi9cfiRf53aQQi:bu8xxq) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1697)):*
> ```cpp
> separated by via points. It stores an ordered list of the obstacle indexes 
> associated with this segment and indexes to the initial and final via points, if
> they exist. An invalid initial via point index indicates that the segemnt begins
> ```

*segment
___
*[`Simbody/src/CableSpan.cpp` line 1716 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Yony6l5GS2oolcdQ:-OV6Yony6l5GS2oolcdR:b-kyguc0) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1716)):*
> ```cpp
>     {
>         if (!m_subsystem) {
>             SimTK_ERRCHK_ALWAYS(
> ```

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
___
*[`Simbody/src/CableSpan.cpp` line 1727 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YtdB9QsGhW8hAhyT:-OV6YtdB9QsGhW8hAhyU:b-kyguc0) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1727)):*
> ```cpp
>     {
>         if (!m_subsystem) {
>             SimTK_ERRCHK_ALWAYS(
> ```

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
___
*[`Simbody/src/CableSpan.cpp` line 1787 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YmQSEiKzRt74yiC2:-OV6YmQSEiKzRt74yiC3:bufx4py) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1787)):*
> ```cpp
>     // The index of this CableSegment in the CableSpan.
>     CableSegmentIndex m_index = CableSegmentIndex::Invalid();
>     
> ```

Whitespace.

I'm going to shut up about whitespace for the remainder of this review, but you should use a python script or IDE setting etc. to try and ensure that any new lines that are added contain no trailing whitespace. This PR contains many lines with trailing whitespace.
___
*[`Simbody/src/CableSpan.cpp` line 1922 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Z7Co6lFksnMsTr8L:-OV6Z7Co6lFksnMsTr8M:by02got) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1922)):*
> ```cpp
>     }
> 
>     const ViaPoint& getViaPoint(ViaPointIndex ix) const
> ```

Might be a good idea to use bounds-checked accessors for these (e.g. `vector.at(i)`), if they're available, because if the algorithm surrounding figuring out via points etc. is tricksy then it might be dividends for maintenance to have the program deterministically fail on OOB conditions.
___
*[`Simbody/src/CableSpan.cpp` line 2058 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6_6BLDI2Dp_QRRlkp:-OV6_6BM9hBV5SzyHXWy:b-uhn38i) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2058)):*
> ```cpp
>             m_subsystem,
>             getIndex(),
>             m_cableSegments.back().getIndex(),
> ```

Might be useful to put a DEBUG assert along the lines of `!m_cableSegments.empty() && "an initial cable segment should've been added by the CableSpan constructor"`, just so that invariant is understood - I had to dig much further down this file to figure out that there's an "add initial cable segment" step after the `Impl` is already constructed.
___
*[`Simbody/src/CableSpan.cpp` line 2087 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6bFox4MTgyMwYNHzP:-OV6bFox4MTgyMwYNHzQ:bf7b4zl) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2087)):*
> ```cpp
> 
>         // Mark the end of the last CableSegment with this ViaPoint.
>         m_cableSegments.back().setFinalViaPointIndex(viaPointIx);
> ```

Similar DEBUG assertion to the above might be useful but, again, this is only an optional suggestion to make maintenance a little bit easier.
___
*[`Simbody/src/CableSpan.cpp` line 2533 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6ccd51540HVdIl1Ph:-OV6ccd51540HVdIl1Pi:b-ljlpee) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2533)):*
> ```cpp
>     // in the current cable segment.
>     for (ObstacleIndex ix(m_obstacleIndex); 
>             ix > cableSegment.getObstacleIndexes().front(); --ix) {
> ```

Unsure about context, but would this be problematic if the given cableSegment contains no obstacles?
___
*[`Simbody/src/CableSpan.cpp` line 2555 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6dPk6B-fWqyof5saA:-OV6dPk6B-fWqyof5saB:bujj6z9) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2555)):*
> ```cpp
>     // cable segment.
>     for (ObstacleIndex ix(m_obstacleIndex + 1); 
>             ix < cableSegment.getObstacleIndexes().back()+1; ++ix) {
> ```

Same as above, it isn't clear to me that it's guaranteed that every `cableSegment` will have at least one obstacle.
___
*[`Simbody/src/CableSpan.cpp` line 2671 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6e1ed35eXvwwDcMvh:-OV6e1ed35eXvwwDcMvi:bjio1uk) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2671)):*
> ```cpp
> 
>     // The via point moment arm in ground.
>     const Vec3& r_G = viaPoint.getStation_G(s) - x_GB;
> ```

A vector subtraction returns a new `Vec3` value, not a reference to a `Vec3`.

Quickly looking at the maths, `r_G` might be "the location of the via point station in the body" rather than "the location of the via point in ground"? The `_G` is making me think "these are quantities in ground"? But the SpatialVec maths below implies it's all in the same coordinate system.
___
*[`Simbody/src/CableSpan.cpp` line 2757 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6fgF611bmAjxK8uFg:-OV6fgF611bmAjxK8uFh:b6wm7jj) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2757)):*
> ```cpp
>             s,
>             unitForce_G * tension,
>             bodyForcesInG);
> ```

I'm beginning to worry that the `OpenSim::ForceProducer` API was ill-conceived - simbody's internal manipulation of the body forces vector makes it quite hard to introspect which OpenSim component is applying which forces :shaking_face: 
___
*[`Simbody/src/CableSpan.cpp` line 3565 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6gvjS1rXJ2PkRlW2S:-OV6gvjS1rXJ2PkRlW2T:b-maguj2) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3565)):*
> ```cpp
>     // solver output, and write to the cached CableSpanData.
>     auto updDataPosFromSolverResult =
>         [&](const State& s, const CableSegment& cableSegment,
> ```

Mayaswell linebreak each argument rather than arbitrarily deciding that one line should have two, another should have one, etc.
___
*[`Simbody/src/CableSpan.cpp` line 3612 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6hZndEPj3A6Y8op66:-OV6hZndEPj3A6Y8op67:bc8o39l) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3612)):*
> ```cpp
>     for (CableSegmentIndex cix(0); cix < getNumCableSegments(); ++cix) {
>         const CableSegment& cableSegment = getCableSegment(cix);
>         Array_<ObstacleIndex> obstacleIndexes = 
> ```

This is being allocated for each iteration of the outer loop
___
*[`Simbody/src/CableSpan.cpp` line 3627 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6huDdCIAjqwg00AVr:-OV6huDdCIAjqwg00AVs:bi0dou1) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3627)):*
> ```cpp
>             // and will update each curve's WrappingStatus.
>             for (ObstacleIndex ix : obstacleIndexes) {
>                 const CurveSegment& curveSegment = getObstacleCurveSegment(ix);
> ```

Could still be a one-liner, but with a range-foor loop
___
*[`Simbody/src/CableSpan.cpp` line 3647 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6iK3FBFUvnZ2b_0ya:-OV6iK3FBFUvnZ2b_0yb:bzh0hgz) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3647)):*
> ```cpp
>             }
> 
>             // Apply the corrections to each CurveSegment to reduce the path 
> ```

Sticking to some kind of (quite low) column limit isn't really worth adding line breaks like this.

I've spotted many instances in this PR that break up a comment line, or make some silly boundary comment (`// ------`) shorter, or linebreak a function call etc. in order to obey a column limit. This just makes the PR noisier and sometimes even creates negative side-effects, such as a comment going from a readable single line to one where there's a single word orphaned on the next line, or a function call like `f(x,y,z)` becoming something like `f(x,y\n          z);`. Imo, if it's between being a few characters over an (arbitrary) column limit and readability, prefer readability.

I'm going to shut up about column limits for the remainder of this review because it's tedious. @sherm1 can call it however he likes it.
___
*[`Simbody/src/CableSpan.cpp` line 4417 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6nEzT9w4yx2hi1HVz:-OV6nEzT9w4yx2hi1HW-:b-8a6nxi) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L4417)):*
> ```cpp
>         terminationStation))
> {
>     CableSpanIndex ix = subsystem.updImpl().adoptCable(*this);
> ```

This constructor is now setting up a class invariants for `Impl` (e.g. "it is constructed with one cable segment").

I think a clearer design would be to have `Impl` accept a reference to its wrapping `CableSpan` as an argument so that it can set up all three conditions (adopted into the subsystem, index set, first cable span created) in `Impl`'s constructor, so that everything is in one place.

Ultimately `CableSpan` is just a wrapper handle for `Impl` anyway. Qt has a similar relationship between `QWidget` and `QWidgetPrivate` where the relationship  between the two is handled like `QWidgetPrivate(QWidget& owner, ctor_args...)`. OSC does something similar for its UI/PIMPL design where it's necessary to have access to the outer handle from the inner Impl.

nickbianco (commented):
Thanks for the quick and detailed review @adamkewley. I've addressed most comments; the others might require a little input from @sherm1.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OV8u_l75Xu-jmEEfg9l:b-uvfgjj)* status: 4 of 10 files reviewed, 12 unresolved discussions (waiting on @adamkewley and @sherm1)

___
*[`Simbody/src/CableSpan.cpp` line 1475 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Ulgf9lDVmdgn1q-A:-OV8FLla09_t0QkFY00W:b-8e1rmk) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1475)):*

Naming class sections like "Constructors.", "Getters", etc. is (imo) more of an OpenSim convention.

I'm not bothered either way but it might be the case that @sherm1 has a preference.
</blockquote></details>

I'm also fine either way. I was just mimicking the style of `CurveSegment` for `ViaPoint` and `CableSegment`.
___
*[`Simbody/src/CableSpan.cpp` line 1481 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6VmHKEjNAgOJH_JAQ:-OV8HCzYELbrCyUTqInd:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1481)):*

This constructor does not initialize two member variables that are without a member initializer (m_body, m_station_b). Either put a deterministic member initializer on the member itself (mentioned below) or initialize it in each constructor.

Otherwise, the initial state of `ViaPoint` is undefined. In your use-case, this might not actually matter if you always call a function that initializes it but, in my experience, can cause extremely annoying-to-debug bugs when the class is changed/updated in the future in such a way that the implementation reads the undefined value to initialize some other algorithm or similar.
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1501 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6WahWAaEblCaIwJaX:-OV8IoRj7EFTEQ_XJS4m:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1501)):*

I'd also recommend zero-initializing the cache entries.

I understand that reading a cache entry that's marked as invalid is bad coding downstream but (imo) you really don't want to debug a program in prod that fails randomly on a particular release/arch/build (you may not even have access to a development machine that matches the user's) because some invalid downstream code reads an undefined value.
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1585 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YNFD9fYc8T_dK5jl:-OV8J4ed9WvNP9Mvw8Pk:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1585)):*

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1596 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YTcCBnpftCRiImo9:-OV8J5FGAJQLZz17alko:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1596)):*

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1681 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6VKsD2ZbmBZ4B2KCM:-OV8J6am1FiqivjohTy_:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1681)):*

If the `ViaPoint` class is going to have a `ViaPoint() = default;` then I'd strongly recommend ensuring that all class members default-construct in a deterministic way. IIRC the index types and `SimTK::Vec3` trivially construct with whatever junk is on the stack/heap if they aren't explicitly initialized with a value.
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1716 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Yony6l5GS2oolcdQ:-OV8JPjU5Wd_qLoh_1Fe:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1716)):*

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1727 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YtdB9QsGhW8hAhyT:-OV8JQ4l3PoKPBhdMqil:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1727)):*

`SimTK_ERRCHK_ALWAYS` already checks that `m_subsystem` is non-nullptr, so the `if` isn't necessary in this case
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 1787 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6YmQSEiKzRt74yiC2:-OV8JlDwEOOf_niTi0RP:bxbuigz) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1787)):*

Whitespace.

I'm going to shut up about whitespace for the remainder of this review, but you should use a python script or IDE setting etc. to try and ensure that any new lines that are added contain no trailing whitespace. This PR contains many lines with trailing whitespace.
</blockquote></details>

Yes, I just realized that my "trim whitespace" IDE setting was disabled (based on Sherm's review of another recent PR). I'll re-enable and try to correct unintentional whitespaces in the files touched by this PR.
___
*[`Simbody/src/CableSpan.cpp` line 2533 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6ccd51540HVdIl1Ph:-OV8WgK8072hnM6bJBEN:b-ja90yc) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2533)):*

Unsure about context, but would this be problematic if the given cableSegment contains no obstacles?
</blockquote></details>

Good catch! I'll add some checks and run everything in DEBUG to confirm.
___
*[`Simbody/src/CableSpan.cpp` line 2671 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6e1ed35eXvwwDcMvh:-OV8XihlCciq5yc87j85:b-qk35d8) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2671)):*

A vector subtraction returns a new `Vec3` value, not a reference to a `Vec3`.

Quickly looking at the maths, `r_G` might be "the location of the via point station in the body" rather than "the location of the via point in ground"? The `_G` is making me think "these are quantities in ground"? But the SpatialVec maths below implies it's all in the same coordinate system.
</blockquote></details>

Done.

`viaPoint.getStation_G(s)` returns the location of the via point in ground and when subtracting the body origin location in ground, `x_GB`, we get the moment arm in ground, `r_G`. 
___
*[`Simbody/src/CableSpan.cpp` line 2757 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6fgF611bmAjxK8uFg:-OV8Zm8LEp_OGjgKKoNo:bq69uff) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L2757)):*

I'm beginning to worry that the `OpenSim::ForceProducer` API was ill-conceived - simbody's internal manipulation of the body forces vector makes it quite hard to introspect which OpenSim component is applying which forces :shaking_face:
</blockquote></details>

I don't know if it was ill-conceived, but the reason this PR adds `calcCurveSegmentUnitForce`, `calcViaPointUnitForce`, etc., to `CableSpan` was to make it easier to conform to the `ForceProducer` API in OpenSim. In my current draft of the new OpenSim wrapping class, I essentially recreate `CableSpan::applyBodyForces` in the `produceForces` implementation. We could discuss further outside of this PR.
___
*[`Simbody/src/CableSpan.cpp` line 3612 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6hZndEPj3A6Y8op66:-OV8afPE9MC9G7qlc8HJ:b-896fix) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3612)):*

This is being allocated for each iteration of the outer loop
</blockquote></details>

Done.
___
*[`Simbody/src/CableSpan.cpp` line 3647 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6iK3FBFUvnZ2b_0ya:-OV8gTGQ6icIgv2YvSmG:b-3t1rqo) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3647)):*

Sticking to some kind of (quite low) column limit isn't really worth adding line breaks like this.

I've spotted many instances in this PR that break up a comment line, or make some silly boundary comment (`// ------`) shorter, or linebreak a function call etc. in order to obey a column limit. This just makes the PR noisier and sometimes even creates negative side-effects, such as a comment going from a readable single line to one where there's a single word orphaned on the next line, or a function call like `f(x,y,z)` becoming something like `f(x,y\n          z);`. Imo, if it's between being a few characters over an (arbitrary) column limit and readability, prefer readability.

I'm going to shut up about column limits for the remainder of this review because it's tedious. @sherm1 can call it however he likes it.
</blockquote></details>

This is a fair comment and I'm definitely guilty of being obsessive about the 80 column limit in my code. I'm fine reverting some of these changes if they've hurt readability too much. However, the Simbody `CONTRIBUTING.md` guide does suggest a hard 80 column limit.
___
*[`Simbody/src/CableSpan.cpp` line 4417 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6nEzT9w4yx2hi1HVz:-OV8tQLi68sNkecMOq53:b-7hyhoj) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L4417)):*

This constructor is now setting up a class invariants for `Impl` (e.g. "it is constructed with one cable segment").

I think a clearer design would be to have `Impl` accept a reference to its wrapping `CableSpan` as an argument so that it can set up all three conditions (adopted into the subsystem, index set, first cable span created) in `Impl`'s constructor, so that everything is in one place.

Ultimately `CableSpan` is just a wrapper handle for `Impl` anyway. Qt has a similar relationship between `QWidget` and `QWidgetPrivate` where the relationship  between the two is handled like `QWidgetPrivate(QWidget& owner, ctor_args...)`. OSC does something similar for its UI/PIMPL design where it's necessary to have access to the outer handle from the inner Impl.
</blockquote></details>

I agree this would be better, but it seems like the `CableSpan::Impl` needs to be fully constructed before the `CableSpan` is added to the subsystem array when calling `adoptCable`. Otherwise, a segfault will occur downstream (e.g., in `CableSubsystem::Impl::realizeSubsystemTopologyImpl`).
___
*[`Simbody/src/MultibodySystemRep.h` line 331 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6onYmB4RJLQM6U1IJ:-OV8U1oPAikFGTTdWj4r:b5wdt9x) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/MultibodySystemRep.h#L331)):*

Use `SimTK` assertions, rather than the C library assertions, because the project-specific ones have alternatives like `ASSERT_ALWAYS` (which might be what you want) or just `ASSERT`, and the simbody project could  feasibly provide compile-time options that upgrade all `SimTK_ASSERT`s to `ALWAYS` without having to compile in debug mode (e.g. OSC has this behavior). Additionally, SimTK assertions can fail in a simbody-defined way (e.g. exception, rather than `exit`)

This pops up a few times in the PR but I won't mention it again.
</blockquote></details>

Happy to make this change, but the rest of this file does use the C library assertions. Not sure if there is a specific reason for that or not, maybe @sherm1 could chime in here.
___
*[`Simbody/tests/TestCableSpan.cpp` line 500 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6q2CS7_UVJWvVAOAn:-OV8TsRR4XEdPwcGirUk:bcthqeu) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L500)):*

Simbody provides its own `sqrt`, which should be preferred because `std::sqrt` might not support ADL on custom floating-point types (not something that actually matters, but @sherm1 has gone through the effort of ensuring simbody could theoretically work with a user-defined `SimTK::Real` which, I assume, requires using `SimTK::` numeric functions).
</blockquote></details>

Simbody defines `sqrt` for `Vec`, `Mat`, and other composite numerical types. `SimTK::sqrt` fails to compile and do I not see any alternatives (unless I am missing something).

adamkewley:
*[`Simbody/src/CableSpan.cpp` line 3647 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6iK3FBFUvnZ2b_0ya:-OV93nG9CRGxJdTHa826:b6fmxny) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3647)):*

This is a fair comment and I'm definitely guilty of being obsessive about the 80 column limit in my code. I'm fine reverting some of these changes if they've hurt readability too much. However, the Simbody `CONTRIBUTING.md` guide does suggest a hard 80 column limit.
</blockquote></details>

Fair enough. This isn't a sticking point for me at all, just thought I'd mention it in one place and leave it up to you

adamkewley:
*[`Simbody/src/CableSpan.cpp` line 4417 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6nEzT9w4yx2hi1HVz:-OV93yBx4gvUDloV1vwO:b25p2ul) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L4417)):*

I agree this would be better, but it seems like the `CableSpan::Impl` needs to be fully constructed before the `CableSpan` is added to the subsystem array when calling `adoptCable`. Otherwise, a segfault will occur downstream (e.g., in `CableSubsystem::Impl::realizeSubsystemTopologyImpl`).
</blockquote></details>

Ah yes, that can be a problem if the code you're calling into immediately uses the wrapper etc.

adamkewley:
*[`Simbody/src/MultibodySystemRep.h` line 331 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6onYmB4RJLQM6U1IJ:-OV94Cd4EGX6id-JDsQx:bv3ikvt) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/MultibodySystemRep.h#L331)):*

Happy to make this change, but the rest of this file does use the C library assertions. Not sure if there is a specific reason for that or not, maybe @sherm1 could chime in here.
</blockquote></details>

Fine either way for me, just thought I'd mention it!

adamkewley:
*[`Simbody/tests/TestCableSpan.cpp` line 500 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6q2CS7_UVJWvVAOAn:-OV94HxY7s4Ob7FPHHa4:b-4adz9q) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L500)):*

Simbody defines `sqrt` for `Vec`, `Mat`, and other composite numerical types. `SimTK::sqrt` fails to compile and do I not see any alternatives (unless I am missing something).
</blockquote></details>

Ah, I must have misunderstood: I recall something unusual being done with simbody AD or similar where `Real` was redefined to an auto-differentiable form and I guessed that it would require a very strict approach to free-functions like `std::sqrt` `std::pow`, etc.

adamkewley (approved):
:lgtm: (although, again, I haven't ran it myself and I'm not 100 % aware of the mathematical aspects of this part of the codebase)

> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OV94ZF2BH9shiQHoUpa:bxtleuu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

adamkewley:
Sorry for the spam - I'm not familiar with Reviewable and I always forget to bunch everything up into an atomic review/response

nickbianco (commented):
No worries, everything is grouped up fine on my end.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OV9AWQjAYngFouP5ZuP:b4mb98r)* status: 9 of 10 files reviewed, all discussions resolved (waiting on @adamkewley)

___
*[`Simbody/tests/TestCableSpan.cpp` line 500 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6q2CS7_UVJWvVAOAn:-OV9A9MK-dDDem0ggIdn:b-l767cj) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L500)):*

Ah, I must have misunderstood: I recall something unusual being done with simbody AD or similar where `Real` was redefined to an auto-differentiable form and I guessed that it would require a very strict approach to free-functions like `std::sqrt` `std::pow`, etc.
</blockquote></details>

That works is still on Antoine Falisse's `simbodyAD` branch which hasn't been merged into master. FWIW, I had also assumed Simbody had it's own `sqrt` function. I'm always surprised when Simbody doesn't have something I'm looking for :)

nickbianco:
After running in DEBUG, I caught one error where I was trying to adopt the `CableSubsystem` twice after adding `MultibodySystem::setCableSubsystem()`, which adopts the subsystem internally. All tests and examples work as expected after fixing that.

adamkewley (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

adamkewley:
Nice work @nickbianco . Just so you know, I am away for a while (>3 weeks), so if you need further checks/reviews within a reasonable time then I'd recommend finding another reviewer.

sherm1 (approved):
Great work, Nick! :lgtm_strong: with a few minor comments

> Reviewed 4 of 10 files at r1, 3 of 6 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OXG0s0M4M5i11GJCV4g:bcfqak3)* status: all files reviewed, 8 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/src/CableSpan.cpp` line 1475 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6Ulgf9lDVmdgn1q-A:-OXFp7Ee6pT_fdkuFksR:b-la05rs) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L1475)):*

I'm also fine either way. I was just mimicking the style of `CurveSegment` for `ViaPoint` and `CableSegment`.
</blockquote></details>

Author's choice IMO. Good to be consistent with existing style also. (Strongly preferred within a file, optional otherwise.)
___
*[`Simbody/src/CableSpan.cpp` line 3647 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6iK3FBFUvnZ2b_0ya:-OXFudp4EbjUv5Re3uTy:bi01tev) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/CableSpan.cpp#L3647)):*

Fair enough. This isn't a sticking point for me at all, just thought I'd mention it in one place and leave it up to you
</blockquote></details>

While Adam makes a fair point, I prefer to be strict about the style guide's 80-character column length. There has to be some limit and 80 allows for three side-by-side open files in IDEs with a font readable by, um "senior" programmers.
___
*[`Simbody/src/MultibodySystemRep.h` line 331 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6onYmB4RJLQM6U1IJ:-OXFwo3XAIikZXJs4Niu:b-r6jgw6) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/src/MultibodySystemRep.h#L331)):*

Fine either way for me, just thought I'd mention it!
</blockquote></details>

My guess is that this old code may have preceded SimTK_ASSERT, so probably should be updated. SimTK_ASSERT generates trap-able exceptions rather than aborts, can produce a helpful message, and can optionally be included in optimized builds for tracking down those annoying problems which won't reproduce in Debug. So better to use it in new code.
___
*[`Simbody/tests/TestCableSpan.cpp` line 500 at r1](https://reviewable.io/reviews/simbody/simbody/828#-OV6q2CS7_UVJWvVAOAn:-OXG-9ua8WGX9rZT_u-M:b-37vhx5) ([raw file](https://github.com/simbody/simbody/blob/3dbf7d538bb4c125164425802c8fcf2a333cf49f/Simbody/tests/TestCableSpan.cpp#L500)):*

That works is still on Antoine Falisse's `simbodyAD` branch which hasn't been merged into master. FWIW, I had also assumed Simbody had it's own `sqrt` function. I'm always surprised when Simbody doesn't have something I'm looking for :)
</blockquote></details>

No worries here -- since the argument is a C++ double constant the built-in sqrt is totally up to the task!
___
*[`Simbody/include/simbody/internal/CableSpan.h` line 128 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFicsb4wO_R6siu6HF:-OXFicsb4wO_R6siu6HG:b-zfsbo5) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/include/simbody/internal/CableSpan.h#L128)):*
> ```objc
> //==============================================================================
> /** This class represents the path of a frictionless cable from an origin point
> fixed to a body, over geometric obstacles and through via points fixed to other 
> ```

nit: there are trailing blanks at the end of many of the comment lines in this file.
___
*[`Simbody/include/simbody/internal/CableSpan.h` line 318 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFjfzp0AhlEQTeVOjB:-OXFjfzp0AhlEQTeVOjC:bijtpgg) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/include/simbody/internal/CableSpan.h#L318)):*
> ```objc
>     void calcTerminationUnitForce(
>         const State& state,
>         SpatialVec& unitForce_G) const;
> ```

BTW is these just the unit vectors parallel to the cable at the origin and termination points? If so I'm wondering if there is a better term than "unit force" for these geometric quantities, for example "OriginDirection" and "TerminationDirection"? If you still want to call these forces, consider at least modifying the documentation to note the alternative interpretation as direction vectors. Calling them "OriginForceDirection" and "TerminationForceDirection" might also be clearer than "UnitForce".

There may be other uses for this information than computing forces -- giving them more generic names might help someone find them.
___
*[`Simbody/src/CableSpan.cpp` line 315 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFnruW3H06hZwQgFSJ:-OXFnruX8HbJd7gxCk15:b-5ppj2e) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L315)):*
> ```cpp
>         Array_<UnitVec3, ViaPointIndex> viaPointInTangents_G;
>         // Tangent vectors outgoing from the via points, in the ground frame.
>         Array_<UnitVec3, ViaPointIndex> viaPointOutTangents_G;
> ```

BTW does "incoming" mean "from the origin" and "out going" mean "to the termination"? Could use a definition here of the new terms (or just spell it out).
___
*[`Simbody/src/CableSpan.cpp` line 1650 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFqLVr336ixvggnsqD:-OXFqLVr336ixvggnsqE:b-g4y7fq) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L1650)):*
> ```cpp
> //                              Cable Segment
> //==============================================================================
> /* This is a helper struct for managing segments of the cable path that are
> ```

nit: struct -> class
___
*[`Simbody/src/CableSpan.cpp` line 1655 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFqiRF1IIDsHRx0ZrN:-OXFqiRF1IIDsHRx0ZrO:bnke32v) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L1655)):*
> ```cpp
> they exist. An invalid initial via point index indicates that the segment begins
> with the cable origin point. Similarly, an invalid final via point index
> indicates that the segment ends with the cable termination point. **/
> ```

nit: should terminate with `*/` since we opened with `/*`
___
*[`Simbody/src/CableSpan.cpp` line 1745 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFr2wz2_QJzK9msk-P:-OXFr2wz2_QJzK9msk-Q:b-lue1fs) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L1745)):*
> ```cpp
>     Array_<ObstacleIndex> m_obstacleIndexes;
> 
>     // The indexes to the initial and final via points belonging to this segment.
> ```

nit: line length
___
*[`Simbody/src/CableSpan.cpp` line 2555 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFt0cND_I9b7rTx54J:-OXFt0cND_I9b7rTx54K:b-d5hoxq) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L2555)):*
> ```cpp
>     // cable segment.
>     for (ObstacleIndex ix(m_obstacleIndex + 1);
>             ix < obstacleIndexes.back()+1; ++ix) {
> ```

BTW consider using `<=`



_Suggestion:_
```cpp
    for (ObstacleIndex ix(m_obstacleIndex + 1);
            ix <= obstacleIndexes.back(); ++ix) {
```
___
*[`Simbody/tests/TestCableSpan.cpp` line 505 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXG-P5y0MJEUqUk0K8J:-OXG-P5y0MJEUqUk0K8K:b-lue1fs) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/tests/TestCableSpan.cpp#L505)):*
> ```cpp
>         std::abs(expectedTotalCableLength - gotTotalCableLength) <
>             lengthTolerance,
>         "Expected cable length (=%f) does not match computed cable length (=%f), error = %e",
> ```

nit: line length

nickbianco (commented):
Thanks Sherm! I've addressed all of the remaining comments, except for the one regarding the naming of the "unit force" calculation methods. I've left my reasoning for keeping the existing method names under that comment below.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OXxxk013KGnly15ttDy:bfvg4le)* status: 6 of 10 files reviewed, 1 unresolved discussion (waiting on @adamkewley and @sherm1)

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 128 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFicsb4wO_R6siu6HF:-OXxpFr58Yj-SjDXCUTr:bj4w8ks) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/include/simbody/internal/CableSpan.h#L128)):*

nit: there are trailing blanks at the end of many of the comment lines in this file.
</blockquote></details>

Gah! I thought I fixed my IDE for this. Fixed now.
___
*[`Simbody/include/simbody/internal/CableSpan.h` line 318 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFjfzp0AhlEQTeVOjB:-OXxqQFX0cW8AKB1JGwg:bt2qsfk) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/include/simbody/internal/CableSpan.h#L318)):*

BTW is these just the unit vectors parallel to the cable at the origin and termination points? If so I'm wondering if there is a better term than "unit force" for these geometric quantities, for example "OriginDirection" and "TerminationDirection"? If you still want to call these forces, consider at least modifying the documentation to note the alternative interpretation as direction vectors. Calling them "OriginForceDirection" and "TerminationForceDirection" might also be clearer than "UnitForce".

There may be other uses for this information than computing forces -- giving them more generic names might help someone find them.
</blockquote></details>

The translational component of the returned `SpatialVec`s are the unit vectors parallel to the cable at the origin/termination (i.e., the force direction). But the `SpatialVec` also contains a rotational component since the cable origin and termination points might not be located at the origins of the mobilized bodies they're attached to. So I think "unit force" is appropriate here, but let me know if additional comments would be useful here.
___
*[`Simbody/src/CableSpan.cpp` line 315 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFnruW3H06hZwQgFSJ:-OXxvOC8C1AKot5bxPqr:b-416cf0) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/src/CableSpan.cpp#L315)):*

BTW does "incoming" mean "from the origin" and "out going" mean "to the termination"? Could use a definition here of the new terms (or just spell it out).
</blockquote></details>

These vectors are tangent to the cable segments immediately prior (incoming) and immediately following (outgoing) each via point in the path. I rephrased to make this clearer.
___
*[`Simbody/tests/TestCableSpan.cpp` line 505 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXG-P5y0MJEUqUk0K8J:-OXxxFrD1edG3xCN1jbm:b-k7tkt) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/tests/TestCableSpan.cpp#L505)):*

nit: line length
</blockquote></details>

Fixed this line, and other similar lines in this file.

sherm1 (approved):
> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 318 at r4](https://reviewable.io/reviews/simbody/simbody/828#-OXFjfzp0AhlEQTeVOjB:-OXyRY3ADOICyLba1_HE:bg2j5hf) ([raw file](https://github.com/simbody/simbody/blob/4f371c1c7e22c0d4cd0cf7e730206bc1232ceec2/Simbody/include/simbody/internal/CableSpan.h#L318)):*

The translational component of the returned `SpatialVec`s are the unit vectors parallel to the cable at the origin/termination (i.e., the force direction). But the `SpatialVec` also contains a rotational component since the cable origin and termination points might not be located at the origins of the mobilized bodies they're attached to. So I think "unit force" is appropriate here, but let me know if additional comments would be useful here.
</blockquote></details>

Makes sense, thanks.

sherm1 (approved):
Thanks, Nick -- merging now.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/828#-:-OXyS88t8gDvRsxoNwPB:bcq03fs)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

Issue #831, opened 2025-07-25
`CableSpan` behaves incorrectly during time-stepping integration


nickbianco:
### Summary
I've begun testing the use of `CableSpan` during a simulation (using `SimTK::TimeStepper`), and I am noticing odd behavior during time-stepping integration. Here is a test program based on the README double pendulum example with a wrapping cylinder and `CableSpan` added:

```c++
int main() {
    // Define the system.
    MultibodySystem system;
    SimbodyMatterSubsystem matter(system);
    GeneralForceSubsystem forces(system);
    CableSubsystem cables(system);
    Force::Gravity gravity(forces, matter, -YAxis, 9.8);

    // Describe mass and visualization properties for a generic body.
    Body::Rigid bodyInfo(MassProperties(1.0, Vec3(0), UnitInertia(1)));
    bodyInfo.addDecoration(Transform(), DecorativeSphere(0.1));

    // Create the moving (mobilized) bodies of the pendulum.
    MobilizedBody::Pin pendulum1(matter.Ground(), Transform(Vec3(0)),
            bodyInfo, Transform(Vec3(0, 1, 0)));
    MobilizedBody::Pin pendulum2(pendulum1, Transform(Vec3(0)),
            bodyInfo, Transform(Vec3(0, 1, 0)));

    // Add a cable.
    CableSpan cable(
        cables,
        matter.Ground(),
        Vec3{0.25, 0.0, 0.0},
        pendulum2,
        Vec3{0.0, 0.5, 0.0});
    cable.addObstacle(
        pendulum1,
        Transform(Vec3(0.2, 0.0, 0.0)),
        std::shared_ptr<ContactGeometry>(
            new ContactGeometry::Cylinder(0.1)),
        Vec3{0.1, 0.0, 0.0});
    cable.setAlgorithm(CableSpanAlgorithm::Scholz2015);

    // Visualize the system.
    system.setUseUniformBackground(true ); // no ground plane in display
    Visualizer viz(system);
    viz.addDecorationGenerator(new CableDecorator(system, cable));
    system.addEventReporter(new Visualizer::Reporter(viz, 0.01));

    // Initialize the system and s.
    system.realizeTopology();
    State state = system.getDefaultState();

    // Initialize the system and state.
    pendulum2.setRate(state, 5.0);

    // Simulate for 20 seconds.
    SemiExplicitEuler2Integrator integ(system);
    TimeStepper ts(system, integ);
    ts.initialize(state);
[...]

PR #791, opened 2024-05-27
Scholz2015 wrapping algorithm

pepbos:
# Scholz2015 Wrapping Algorithm
Implementation of a new algorithm for computing the path of a weightless, and frictionless cable spanning over multiple smooth obstacle surfaces[^fn].

[^fn]: Scholz, A., Sherman, M., Stavness, I. et al (2016). A fast multi-obstacle muscle wrapping method using natural geodesic variations. Multibody System Dynamics 36, 195–219.
## Algorithm Overview
**Creating a cable**
A new subsystem class `CableSubsystem` is added to manage the cables in a simulation:
```cpp
MultibodySystem system;
CableSubsystem cables(system);
```
A cable is represented by the new class `CableSpan`, and allows a path to be spanned between two points:
```cpp
CableSpan cable(
        cables,
        body1, Vec3(0.), // Path origin
        body2, Vec3(0.)); // Path termination
```
a_cable
Obstacles can be added to the cable to make it more interesting:
```cpp
cable.addObstacle(obstacleBody, obstacleOffset, obstacleGeometry);
```
For example, we can add a torus to the scene:

https://github.com/simbody/simbody/assets/46680867/49ea834d-bf1b-49fd-9d11-b90ef7dfeb69

**Initializing the Path**
To help control the cable's path you can choose a contact point hint for each obstacle. For example, placing the contact point within the torus:
```cpp
cable.setContactPointHint(obsIndex, -torusRadius * Vec3{YAxis} / 2);
```
allows spanning the cable through the torus:
set_contact_point_hint

**Obstacle Ordering**
Currently, the order in which the cable can come into contact with the obstacle is ordered based on the order in which the obstacles are added to the cable. This was done to reduce both the algorithm complexity and simplify building your path, while being useful in a large set of problems.

For the previous example this means that a `CableSpan` will not collide with the torus twice:

https://github.com/simbody/simbody/assets/46680867/94917a9f-1a10-485c-be53-dad9d78cfa4f

or in the case of two obstacles, that collision with the first obstacle will not be detected after contact with the second obstacle:

https://github.com/simbody/simbody/assets/46680867/38173be5-de17-44e5-ae77-48ee69d7a3a0

It is still possible to make contact with a single *surface geometry* multiple times. But you have to add that surface as an obstacle multiple times. For example, if we wish to make contact with a torus twice, it can be added as an obstacle twice:
```cpp
// Start with torus obstacle
cable.addObstacle(ground, offset, torusGeometry, contactPoint1);
[...]

PR #832, opened 2025-07-28
Implement `realizeSubsystemPosition` and `realizeSubsystemVelocityImpl()` for `CableSubsystem`

nickbianco:
### Summary
This change implements the `Subsystem` virtual methods `realizesSubsystemPositionImpl` and `realizeSubsystemVelcoityImpl` to fix `CableSpan` behavior during time-stepping integration.  

Previously, `CableSpan` would exhibit the behavior described in #831, where the path would jump around and find invalid configurations during a simulation. After implementing the `Subsystem` realization interface, the behavior improves. Here is the result with `SemiExplicitEuler2Integrator`:

https://github.com/user-attachments/assets/7904fdce-98eb-4187-a783-865c0715f76c

Note that if simulating this double pendulum example with higher-order integrators that take larger steps sizes (e.g., `RungeKuttaMersonIntegrator`), an upper bound on the maximum step size will need to be set, otherwise the path may "step through" the wrap cylinder during the simulation. This seems to be the same behavior as the "path escaping the torus" example described in [the original `CableSpan` PR](https://github.com/simbody/simbody/pull/791).

### `CurveSegment` cache access

To implement the `Subsystem` realization interface, I needed to replace usages of `CurveSegment::getDataPos` with `CurveSegment::updDataPos` in any place that they appeared within a call stack originating from `CableSpan::Impl::calcDataPos`. This is because the `State` provided to `realizesSubsystemPositionImpl` will only be realized to `Stage::Time`, and so the position-level cache entries of `CurveSegment` are only accessible via `updCacheEntry` and not `getCacheEntry`.

However, I'm not sure this is the ideal solution. It's a bit of an odd situation, since `CurveSegment` will calculate position-level information internally that is then used by `CableSpan::Impl` to perform its own position-level calculations. It makes sense that the `CurveSegment` information should be fixed within `CableSpan::Impl` calculations, but both will share a `State` at the same `Stage`, which leads to the issue I encountered. I suppose an alternative would be lower the "earliest" `Stage` set by `allocateCacheEntry` for `CurveSegmentData::Position` to `Stage::Time`. I'm not sure which is preferred here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/832)

### DIFF
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -1522,6 +1522,21 @@ public:
         }
     }
 
+    void realizeModel(State& state) const
+    {
+        // No choices at the moment.
+    }
+
+    void realizeInstance(const State& state) const
+    {
+        // Nothing to compute here yet.
+    }
+
+    void realizeTime(const State& state) const
+    {
+        // Nothing to compute here yet.
+    }
+
     void realizePosition(const State& state) const
     {
         if (getSubsystem().isCacheValueRealized(state, m_indexDataPos)) {
@@ -1539,6 +1554,21 @@ public:
         }
         calcDataVel(state, updDataVel(state));
         getSubsystem().markCacheValueRealized(state, m_indexDataVel);
+    }
+
+    void realizeDynamics(const State& state) const
+    {
+        // Nothing to compute here yet.
+    }
+
+    void realizeAcceleration(const State& state) const
+    {
+        // Nothing to compute here yet.
+    }
+
+    void realizeReport(const State& state) const
+    {
+        // Nothing to compute here yet.
     }
 
     const CableSpanData::Position& getDataPos(const State& state) const
@@ -3640,19 +3670,28 @@ int CableSubsystem::Impl::realizeSubsystemTopologyImpl(State& state) const
 
 int CableSubsystem::Impl::realizeSubsystemModelImpl(State& state) const
 {
-    // No choices at the moment.
+    for (CableSpanIndex ix(0); ix < cables.size(); ++ix) {
+        getCable(ix).getImpl().realizeModel(state);
+    }
+
     return 0;
 }
 
 int CableSubsystem::Impl::realizeSubsystemInstanceImpl(const State& state) const
 {
-    // Nothing to compute here.
+    for (CableSpanIndex ix(0); ix < cables.size(); ++ix) {
+        getCable(ix).getImpl().realizeInstance(state);
+    }
+
     return 0;
 }
 
 int CableSubsystem::Impl::realizeSubsystemTimeImpl(const State& state) const
 {
-    // Nothing to compute here.
+    for (CableSpanIndex ix(0); ix < cables.size(); ++ix) {
+        getCable(ix).getImpl().realizeTime(state);
+    }
+
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


nickbianco:
@sherm1 I think this one is ready for review. There may be other lingering bugs when integrating a system with `CableSpan`, but after testing with Simbody's integrator suite, this seems to fix the issue in #831 (provided that an upper bound on the integrator step size is set).

sherm1 (approved):
Thanks, Nick. :lgtm:
One suggestion below, possibly avoiding future trouble of the same kind.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/832#-:-OWc4T9cDL5u2WUT8g8h:bf25tw4)* status: all files reviewed, 2 unresolved discussions (waiting on @nickbianco)

___
*[`Simbody/src/CableSpan.cpp` line 589 at r1](https://reviewable.io/reviews/simbody/simbody/832#-OWc3q8VDCFnhQNAgkYy:-OWc3q8VDCFnhQNAgkYz:b-c66qpu) ([raw file](https://github.com/simbody/simbody/blob/57be5feb4cf8b0037e327cb5060ec484012efdb5/Simbody/src/CableSpan.cpp#L589)):*
> ```cpp
> 
>     int realizeSubsystemVelocityImpl(const State& state) const override;
> ```

BTW Instance, Acceleration missing? (see comment below)
___
*[`Simbody/src/CableSpan.cpp` line 3645 at r1](https://reviewable.io/reviews/simbody/simbody/832#-OWc2u_e6waJcX8mQH4r:-OWc2u_e6waJcX8mQH4s:bkc1zbe) ([raw file](https://github.com/simbody/simbody/blob/57be5feb4cf8b0037e327cb5060ec484012efdb5/Simbody/src/CableSpan.cpp#L3645)):*
> ```cpp
> 
>     return 0;
> }
> ```

BTW CableSubsystem::Impl is missing the other usual "realize" calls. For example, see CableTrackerSubsystem_Impl.h. Why not implement all of those? Is that because CableSpan's don't implement the corresponding "realize" methods?

I'm thinking it would be better to fill out those to make the Subsystem more boilerplate. It's fine if some of the realize methods don't currently do anything -- they might be more useful in the future.

sherm1 (approved):
I was thinking that the subsystem realize methods should all invoke the corresponding CableSpan realize methods via a loop over all the contained cables (like the realizePosition/Velocity methods do now). That way a future programmer would need only to fill out the CableSpan methods to get proper behavior. As written here, implementing a CableSpan::realizeInstance() (for example) would be allowed but it would never get called.

I don't feel strongly about that -- but I'd be inclined _not_ to stub the subsystem methods here if they aren't going to run through the cables to invoke their realize methods. I'm worried that someone will see that they are implemented and assume (unwisely, without checking) that they are fully implemented. So my preference would be either (a) go back to the previous state where these are unimplemented, or (b) implement the loops at the subsystem level and stub the low-level realize methods.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/832#-:-OWg4aGKAKUoXmesLBmR:b-9z6e0a)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

sherm1 (approved):
Perfect, thanks Nick! Merging now.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/832#-:-OWq8Ug629tE4SBONaUM:bsjsixs)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nickbianco)

PR #840, opened 2025-10-23
Fix out-of-bounds accesses in Mat constructors

adamkewley:
Fixes a potential out-of-bounds error that more modern compilers (e.g. gcc13) have started to warn about.

---

One warning is related to `NActualElements`. The original implementation puts a trailing gap on each `Row` element, for ease-of-access, but this means that `Row`s that are `reinterpret_cast`ed from arrays of `Element`s (e.g. in `Mat`) create an object with a representation that is out-of-bounds of the array, triggering a compiler warning.

The solution is to calculate the actual number of elements in the representation without any trailing elements.

---

Another warning is that the `Mat` elements-row-by-row overloads are out-of-bounds if the number of elements in the constructor exceeds the number of elements in the matrix. (e.g. if you write something like `Mat22{a,b,c,d,e,f}`). The original implementation works around this problem with a runtime `assert` but modern compilers still spot it as a potential issue - even if the overload is never used.

The solution is to use `enable_if` to only enable overloads that are valid for the given matrix. This informs the compiler that the overload is only valid if that precondition is met.

It would be very very ugly to write `std::enable_if` logic for the 15 or so overloads of this function. Fortunately, varadaic template packs make it possible to write a generic overload for N arguments, with a `std::enable_if` check, and with perfect forwarding of the arguments to their final location in `Mat`'s representation.

Unfortunately, writing that kind of generic code without it being a giant PITA requires C++17. So the required C++ level was also changed to C++17. C++17 has excellent coverage on all operating systems released since around 6 years ago, if that's acceptable for compatibility?

For context, opensim-core already compiles with C++17. OpenSim Creator compiles with C++20 (good compatibility on Ubuntu20+, MacOS from 3-4 years ago, Windows). A C++20 version of the overload would be even cleaner, because `std::enable_if_t` can be ported to a `requires` clause and `typename... Elements` can be ported to `std::convertible_to<E>... Elements` 😄

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/840)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ PRs and issues that were included in a particular release.
 --------------------
 * Add LLT (Cholesky) factorization
 * Add quaternion multiplication (Hamilton product) 
+* Required C++ level was increased from C++11 to C++17
+* Fixed an out-of-bounds warning/error emitted by overloads of SimTK::Mat accessing
+  out-of-bounds rows
 
 
 3.8 (May 2025)
--- a/SimTKcommon/CMakeLists.txt
+++ b/SimTKcommon/CMakeLists.txt
@@ -145,7 +145,7 @@ target_compile_definitions(SimTKcommon
 # (maintaining intent of previous behavior, but more universally applied and effective)
 # See https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#language-standard-flags
 # for info re: ordering of the compiler flag(s) for language standards and CMAKE_<LANG>_FLAGS
-target_compile_features(SimTKcommon PUBLIC cxx_std_11)
+target_compile_features(SimTKcommon PUBLIC cxx_std_17)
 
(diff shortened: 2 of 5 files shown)


sherm1:
Awesome, thanks Adam. Will review. I'd be happy to move to C++20 as soon as it isn't a problem for OpenSim. Sounds like 17 is the best we can do for now?

nickbianco:
@adamkewley we currently target Ubuntu 22 and later on the OpenSim CI, so I think we're probably safe to bump up to C++20 (and it's probably time we do).

nickbianco:
> The only thing that's a drag is that std::format isn't supported by Ubuntu 22's libstdc++.

Bummer. For now, we could upgrade to C++20 and let that settle, then give Ubuntu 24 a shot. I tried adding it the CI a few months ago, but ran into some issues and gave up.

PR #841, opened 2025-10-29
Fix SimTK::Factor implementations memory leak on copy assign

adamkewley:
Fixes a memory leak found downstream during libASAN screening in [OpenSim Creator](https://www.opensimcreator.com/). The source of the leak was via an `OpenSim::Model`, which allocates a `CableSubsystem`, which allocates an internal datastructure that includes `FactorQTZ` and friends.

At some point, a copy assignment happens somewhere with a struct containing a `FactorQTZ`, but the current implementation does not handle copy assignment correctly - it overwrites with a `clone` but doesn't account for deleting the left-hand side of the expression.

I decided the best solution is to just convert all `Factor`s to use a `ClonePtr`, which is effectively what all of the code is doing: this reduces the boilerplate and brings move construction/assignment along for the ride.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/841)

### DIFF
--- a/SimTKmath/LinearAlgebra/src/Eigen.cpp
+++ b/SimTKmath/LinearAlgebra/src/Eigen.cpp
@@ -60,31 +60,30 @@ EigenRepBase* EigenDefault::clone() const {
    ///////////
    // Eigen //
    ///////////
-Eigen::~Eigen() {
-    delete rep;
-}
-Eigen::Eigen() {
-   rep = new EigenDefault();
-}
+Eigen::~Eigen() = default;
+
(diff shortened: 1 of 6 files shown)


sherm1 (approved):
Beautiful! Thanks, Adam. :lgtm:

> @sherm1 reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/841#-:-Od6Nl1vC9Mdt6RKWA7V:bhxbmph)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adamkewley)

Issue #843, opened 2025-11-04
A mobilizer inspired by cantilever beams for modeling spinal joints


nickbianco:
I've been kicking around an idea for a mobilizer that could be used to model spinal joints defined using cantilever beam bending equations. The mobilizer would utilize a body-fixed XYZ Euler sequence to define rotations (with hopefully an alternative option for quaternions), and the translation of the child mobilizer frame would be defined based on the deflection of the end of a cantilever beam under a supplied load. This is a visualization of what the beam mobilizer would look like, where here I am just posing a frame based on the proposed mobilizer transform (see below).

https://github.com/user-attachments/assets/2441e52d-2ea8-4859-9de8-ca950c173dc9

The drawn red "beam" is just a visualization reference, and the green point is the load application point. The amount of deflection the beam undergoes for a given orientation can be controlled by moving the loading point up or down the beam. The X and Y rotations incur beam deflections, and the Z rotation allows child frame rotation about an axis tangent to the end of the beam. Since the child frame is tangent to the end of the beam, this definition permits chaining multiple beam mobilizers together in a smooth and continuous way.

The intent for this mobilizer is for modeling the spine in a musculoskeletal model in a semi-realistic way without needing many DOFs or kinematic constraints to define how adjacent vertebrae move with respect to each other. This would be helpful for modeling dynamic movements from athletes where it is important to model accurately upper-body movement (e.g., for tracking markers) but not necessarily important to know the movement or loads for individual vertebrae.

I will lay out some definitions and my current working implementation, with some thoughts about potential improvements.

## Definitions
- $L$: beam length
- $z$: a point along the beam
- $z_{load}$: the load application point along the beam
- $\theta_x, \theta_y, \theta_z$: body-fixed XYZ Euler rotation sequence

## Beam deflection

A deflection coefficient, $$\tilde d(z)$$, can be computed via the following equation,

```math
\tilde d(z) =
\begin{cases}
\frac{z^2}{3 z_{load}^2} (3 z_{load} - z) &  0 \leq z \leq z_{load} \\
\frac{1}{3}(3 z - z_{load}) &  z_{load} \leq z \leq L
\end{cases}
```

[...]

PR #844, opened 2025-11-11
Add `MobilizedBody::CantileverFreeBeam`

nickbianco:
Implements the mobilizer proposed in #843. 

Here's a brief example demonstrating the new `MobilizedBody::CantileverFreeBeam` in the context of a simulation: 

```c++
int main() {
    // Create the system.
    MultibodySystem system;
    SimbodyMatterSubsystem matter(system);

    // Add a mobilized body whose transform we will manipulate via the beam
    // deflection equations.
    Body::Rigid bodyInfo(MassProperties(1.0, Vec3(0), Inertia(1.0)));
    bodyInfo.addDecoration(Transform(), DecorativeFrame(0.1).setColor(Red));
    MobilizedBody::CantileverFreeBeam beamJoint(
            matter.updGround(), Transform(Rotation(-Pi/2, XAxis)),
            bodyInfo, Transform(Rotation(-Pi/2, XAxis)), L);

    // Add prescribed motion constraints.
    Constraint::PrescribedMotion(matter,
            new Function::Sinusoid(Pi/4.0, 1.0 * Pi, 0.0),
            beamJoint, MobilizerQIndex(0));
    Constraint::PrescribedMotion(matter,
            new Function::Sinusoid(Pi/6.0, 2.0 * Pi, Pi/10.0),
            beamJoint, MobilizerQIndex(1));
    Constraint::PrescribedMotion(matter,
            new Function::Sinusoid(Pi/2.0, 0.5 * Pi, Pi/8.0),
            beamJoint, MobilizerQIndex(2));

    // Visualize the system.
    Visualizer viz(system);
    system.addEventReporter(new Visualizer::Reporter(viz, 0.01));

    // Initialize the system and state.
    State state = system.realizeTopology();
    system.realize(state, Stage::Position);

    // Simulate the system.
    RungeKuttaMersonIntegrator integ(system);
    TimeStepper ts(system, integ);
    ts.initialize(state);
    ts.stepTo(20.0);
}
```

https://github.com/user-attachments/assets/9722b959-b31c-4b2f-a224-5dfb934d9063

## Summary of changes

This change adds `MobilizedBody_CantileverFreeBeam.h`, `RigidBodyNodeSpec_CantileverFreeBeam.h`, and all necessary related changes for a `MobilizedBody` implementation. The only testing is via the addition to `TestAngleConversions`, but that could be expanded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/844)

### DIFF
--- a/Simbody/tests/TestMobilizedBody.cpp
+++ b/Simbody/tests/TestMobilizedBody.cpp
@@ -652,13 +652,13 @@ void testCantileverFreeBeamEnergyConservation() {
 
     // Simulate the system and check that the energy is conserved at regular
     // intervals.
-    Real dt = 1e-3;
+    Real dt = 1e-2;
     Real time = 0.0;
     for (int i = 0; i < 1000; ++i) {
         time += dt;
         ts.stepTo(time);
         const State& s = ts.getState();
-        SimTK_TEST_EQ_TOL(initialEnergy, system.calcEnergy(s), 1e-10);
+        SimTK_TEST_EQ_TOL(initialEnergy, system.calcEnergy(s), 1e-8);
     }
 }
 


sherm1:
Oh, actually I think the Motion object is restricted to treating all the dofs the same, which is probably why you used the more-flexible Constraint version! Nvm ...

nickbianco:
Yes, I'm familiar with `Motion`, we use them in OpenSim Moco to perform "prescribed motion" optimal control problems. But yes, the constraint version was easier to make my example look like a [wacky wavy inflatable tube man](https://www.google.com/search?q=wacky+wavy+inflatable+tube+man&sca_esv=2286498253ea9ec4&sxsrf=AE3TifOiz3MTmPRTYbqXSZ5QMLH472R0aw%3A1762897478722&ei=Rq4TaaLeK8Tep84P0MLswAQ&ved=0ahUKEwiimfrviOuQAxVE78kDHVAhG0gQ4dUDCBM&uact=5&oq=wacky+wavy+inflatable+tube+man&gs_lp=Egxnd3Mtd2l6LXNlcnAiHndhY2t5IHdhdnkgaW5mbGF0YWJsZSB0dWJlIG1hbjIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzIKEAAYsAMY1gQYRzINEAAYgAQYsAMYQxiKBTINEAAYgAQYsAMYQxiKBUjFBlAAWABwAXgBkAEAmAEAoAEAqgEAuAEDyAEAmAIBoAICmAMAiAYBkAYKkgcBMaAHALIHALgHAMIHAzAuMcgHAg&sclient=gws-wiz-serp) :D

mitiguy (commented):
Hi Nick -- the description of the mobilizer associated with as a "Beam" confuses me.
Perhaps this is OK, but I wonder if this is the best name for this mobilizer?

Why?  For me, the word "beam" conjures thoughts of how the beam is connected, e.g., pin-pin, pin-roller, cantilever-free, etc.  I also wonder whether it is an Euler beam (long/slender), Timshenko beam (shorter), as well as the beam's material and geometric properties, e.g., beam stiffness (Young's Elastic Modulus E * Area moment of Inertia I), etc.  From what I can see, my physical understanding of a beam (e.g., material properties such as E) do not play into this (or at least it is not obvious to me). 

I was trying to connect the math and physics, starting with a simple version of this beam mobilizer, e.g., one in which there is only one simple rotation (one angle, e.g., theta).   Anyway, this is my confusion.  Perhaps it may be more fruitful to talk about this over a video call. 

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Odp51ue1vPB9xM5n5M-:behdwpt)* status: 0 of 8 files reviewed, all discussions resolved

nickbianco:
The name `Beam` for the mobilizer is more or less a placeholder, so happy to change it to something better. The equations are based on Euler-Bernoulli beam theory, so maybe `MobilizedBody::EulerBeam` is better?

>as well as the beam's material and geometric properties, e.g., beam stiffness (Young's Elastic Modulus E * Area moment of Inertia I), etc. From what I can see, my physical understanding of a beam (e.g., material properties such as E) do not play into this (or at least it is not obvious to me).

The [deflection of the end of a cantilever under a point load](https://home.engineering.iastate.edu/~shermanp/STAT447/STAT%20Articles/Beam_Deflection_Formulae.pdf), as a function of the elastic modulus, inertia, and load $$P$$, is the following:

```math
d = \frac{P L^3}{3 E I}
```
and the deflection angle is,

```math
\theta = \frac{P L^2}{2 E I}
```

For this mobilizer, we want the deflection as a function of deflection angle, or $$d(\theta)$$. We can use the deflection angle formula to solve for $$P$$, and then substitute it into the first formula. Doing this cancels out $$E$$ and $$I$$, leaving us with the following,

```math
d(\theta) = \frac{2 \theta L}{3}
```

Essentially, the deflection is directly a function of the deflection angle, regardless of the applied load or material properties. Given a deflection angle, you can calculate what the load $$P$$ should be for material properties $$E$$ and $$I$$. Maybe that could be added as a utility function if users would care about beam loading.

For this mobilizer, the generalized coordinates $$q_0 = \theta_x$$ and $$q_1 = \theta_y$$ represent deflection angles that induce translations in the Fy and -Fx directions, respectively, and both also contribute a -Fz displacement. The superposition of the deflections/displacements induced by $$\theta_x$$ and $$\theta_y$$ fully the define the translation of the M frame. Then, $$q_2 = \theta_z$$ provides a twist about the beam tangent, Mz. 

Happy to include these details in the mobilizer documentation if useful. I actually started to write out some of the formulas, but I was having trouble getting Doxygen and LaTeX talking to each on my machine.

nickbianco:
Hi Paul, these are good questions. I honestly wasn't sure which choice of beam would make the most sense for the spine, so I wanted to start with something simple to try. Indeed, this is a cantilever-free beam (so maybe `MobilizedBody::CantileverFreeBeam` is appropriate). I started with this because 1) the deflection-angle expression were simple and 2) it allows chaining together multiple joints to create one continuous spine model. It may be that another type of beam is more appropriate for a spine model.

Below is a mockup in OpenSim where I use three beam mobilizer to construct the spine: one each for the lumbar, thoracic, and cervical spines.

Screenshot 2025-11-14 at 11 58 10 AM

The white and green geometry are for reference, the red line is the curve defined by the three connected beam mobilizers. I've included the joint frames to denote where each segment of the spine begin and end. A nice aspect of the current cantilever-free beam approach is that it is easy/simple to make two subsequent beams create a continuous curved segment. But, it should be possible to achieve a similar effect with other beam types. 

The goal of this joint type is not necessarily to create models for studying the spine: if you want to do that, you will probably need to model vertebrae, intervertebral passive structures, etc., explicitly. I'm looking for a joint type with minimal mobilities for modeling the spine with enough fidelity to provide accurate kinematics for the torso and head during dynamic movement. 

I had a slide deck from an internal development meeting that I could send you and Sherm for more context, if that's helpful. I'd also be help to chat over a Zoom call.

mitiguy:
Hi Nick -- thank you for the explanation.  If you have time, let's talk by phone or video.  I would like to learn/understand more.

nickbianco:
Hi @mitiguy, I was actually just about to ping you. As we discussed, I'd be happy to rename to `MobilizedBody::CantileverFreeBeam`, or something similar. Other than that, I was mostly waiting on you and @sherm1 to see if you're happy with the current implementation or if there's any changes you'd like to see.

One tricky thing is that it's hard for me to say if this mobilizer is fully feature-complete for my needs. I may find down the road that having more customization (e.g., over the beam deflection shape) may be important for fitting spine models, and I won't be able to say that until have I something like the current mobilizer to work with. However, I think the implementation here is simple enough to make future changes simple enough. 

I think the most logical next change would be to control the angle of the force applied at the beam end, which would change the beam deflection expressions. If you'd prefer having control for the beam shape included now, I could try adding it in. Or, if it's okay with you both, we could merge this implementation and add more features later.

mitiguy:
Hi Nick,

Thanks for the quick response.  It is difficult to fully know how useful this will be until it is used by other folks. So, after talking to Sherm. perhaps this would be good:  ``MobilizedBody::CantileverFreeBeam``

If later, we want to add a different shape that corresponds to something like a uniform load (rather than an end load), would this be able to retrofitted, e.g., with an optional argument to the constructor?

nickbianco:
Sounds good to me, I'll update the name to `MobilizedBody::CantileverFreeBeam` and we can proceed from there.

> If later, we want to add a different shape that corresponds to something like a uniform load (rather than an end load), would this be able to retrofitted, e.g., with an optional argument to the constructor?

Yes, could have another constructor argument for the beam loading type (e.g., transverse point load, uniform distributed load, etc.), and adjust the beam equations/coefficients underneath the hood. This would mean we would need to find a better way to implement `MobilizedBody::BeamImpl::calcDecorativeGeometryAndAppendImpl` which currently rewrites the beam expressions in order to support visualization. Ideally, this implementation could somehow access the underlying beam equations.

nickbianco:
@mitiguy, @sherm1, I've updated the name to `MobilizedBody::CantileverFreeBeam`. Ready for a full review now.

mitiguy (commented):
> @mitiguy reviewed 2 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844)* status: 6 of 8 files reviewed, 13 unresolved discussions (waiting on @nickbianco and @sherm1)

___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 36 at r3](https://reviewable.io/reviews/simbody/simbody/844#-Og-mWQXAYZdjrYyBb1n:-Og-mWQXAYZdjrYyBb1o:beq4tay) ([raw file](https://github.com/simbody/simbody/blob/74aa887e321c64be02eaaff56be83e4801eab278/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L36)):*
> ```objc
> /** Three mobilities -- rotation with coordinated translation based on the
> deflection path of a cantilever beam with a free end subjected to a transverse
> point load.
> ```

nit  Would it be possible to provide a little more detail here.  I provided my best guess.



_Suggestion:_
```objc
/** Although this mobilizer has only three mobilities (all rotational),
the mobilized frame M translates relative to the fixed frame F as a 
specified function of this mobilizers rotational generalized coordinates 
q0, q1, q2. The specified function is based on the deflection shape of a 
cantilever-free beam, whose free end is subjected to a transverse point load.
Although this shape is inspired by a cantilever-free beam, this mobilizer 
does not actually model a beam, in that there is no elastic restoring 
forces/torques when q0 or q2 or q3 ≠ 0. In the undeformed state (when 
q0 = q2 = q3 = 0), the position from Fo (frame F's origin) to 
Mo (frame M's origin) is L Fz.
```
___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 44 at r3](https://reviewable.io/reviews/simbody/simbody/844#-Og-tIV68_nCy1fQUS_R:-Og-tIV68_nCy1fQUS_S:b-jj4bsw) ([raw file](https://github.com/simbody/simbody/blob/74aa887e321c64be02eaaff56be83e4801eab278/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L44)):*
> ```objc
> 
> The first two rotations induce translations according to the beam deflection
> formula for a cantilever beam under a transverse point load applied at the end
> ```

nit: Is it possible to provide equations here for x as a function of L, q0, q1, q2?
Similar for for y and z ?  If they are not too long, it will avoid needing to have a reference book to understand.

nickbianco (commented):
@sherm1, @mitiguy, I've fixed all minor issues, added equations to the `MobilizedBody::CantileverFreeBeam` documentation, and added a test to `TestMobilizedBody.cpp`. I've also started working on a test for `TestReverseMobilizers.cpp`, but currently half of the checks fail starting with the velocity-level checks. I haven't fully confirmed, but it seems like it might be related to `setUToFitVelocity()` and the related methods. I'll dig into it a bit more tomorrow.

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Og51be2EpvnqWlOGPqa:b63b0xb)* status: 5 of 10 files reviewed, 2 unresolved discussions (waiting on @mitiguy and @sherm1)

___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 60 at r2](https://reviewable.io/reviews/simbody/simbody/844#-OfvKK6UBfsKStMFOf79:-Ofzs2JW0ibMTBDYbhVz:b-l9gdek) ([raw file](https://github.com/simbody/simbody/blob/f6117b831083805b3058673cb889c7a56946c6e0/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L60)):*

BTW you can use Unicode (±π/2) in comments for improved readability if you like. I find https://www.unicodeit.net/ very handy for that -- converts Latex commands to Unicode characters.
</blockquote></details>

Nice! I will try incorporating this in my comments more often.
___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 36 at r3](https://reviewable.io/reviews/simbody/simbody/844#-Og-mWQXAYZdjrYyBb1n:-Og0HUhZ53AmT9RtRMtP:bpxlyc1) ([raw file](https://github.com/simbody/simbody/blob/74aa887e321c64be02eaaff56be83e4801eab278/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L36)):*
<details><summary><i>Previously, mitiguy (Mitiguy) wrote…</i></summary><blockquote>

nit  Would it be possible to provide a little more detail here.  I provided my best guess.
</blockquote></details>

Great suggestions -- I'll substitute these in.

nickbianco (commented):
> There seems to be a linear relationship between translation and rotation, but the vice-versa leaves q\_2 undefined (which may be OK).

This is intended: the last rotation, `q_2`, is independent from the beam bending relationships.

>  I was thinking it would be interesting to do a guess and check on what happens when:  
p\_x = p\_y = 0, p\_z = 1.01_L and then to solve for q\_0, q\_1, q\_2.  
Similarly for  
p\_x = p\_y = 0, p\_z = 0.99_L and then to solve for q\_0, q\_1, q\_2.

It is not possible to set arbitrary translations with this mobilizer. If you force p_x = p_y = 0, then that means q_0 = q_1 = 0 based on the mobilizer definition. Therefore, if the beam is length L, then is not possible to achieve p_z = 0.99 L or p_z = 1.01 L. This is discussed in a little more detail in the implementation of `setQToFitTranslationImpl()` for this mobilizer (which was copied from the `Ellipsoid` mobilizer).

> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Og9E9O2EA4htxhYE5gn:bc1tvdb)* status: 8 of 12 files reviewed, 2 unresolved discussions (waiting on @mitiguy and @sherm1)

___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 139 at r6](https://reviewable.io/reviews/simbody/simbody/844#-Og98AqY0SbIcp-ssdNr:-Og9CMRt52BBhb5_XPvP:bnngy7h) ([raw file](https://github.com/simbody/simbody/blob/210b1ee02636caeb880e7e1a1f44e08e3ef64c8e/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L139)):*
<details><summary><i>Previously, mitiguy (Mitiguy) wrote…</i></summary><blockquote>

nit  It seems the results for p_x, p_y, p_z are linear in q_0, q_1, q_2.  In addition to linear beam theory, this seems to also use small angles.  If this is true, consider something like what is shown below
</blockquote></details>

I wouldn't include this suggestion, because the relationships for `p_x`, `p_y`, and `p_z` here are the exact relationships that define the mobilizer, i.e., they are the relationships used for all `q`, not just small angles. I acknowledge that, from beam theory, these relationships are only valid for small angle deflections, but I am claiming (at least for now) that it is good enough for my purposes. The comment in the doxygen below also tries to address this.

mitiguy (commented):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844)* status: 8 of 12 files reviewed, 2 unresolved discussions (waiting on @nickbianco and @sherm1)

___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 139 at r6](https://reviewable.io/reviews/simbody/simbody/844#-Og98AqY0SbIcp-ssdNr:-OgAO6r2Dr2l2qllWeAu:b-iab5xm) ([raw file](https://github.com/simbody/simbody/blob/210b1ee02636caeb880e7e1a1f44e08e3ef64c8e/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L139)):*

I wouldn't include this suggestion, because the relationships for `p_x`, `p_y`, and `p_z` here are the exact relationships that define the mobilizer, i.e., they are the relationships used for all `q`, not just small angles. I acknowledge that, from beam theory, these relationships are only valid for small angle deflections, but I am claiming (at least for now) that it is good enough for my purposes. The comment in the doxygen below also tries to address this.
</blockquote></details>

OK -- sounds good.

mitiguy (commented):
Hi Nick.  I reviewed some of the unicode with a few suggestions.  I like the unicode better than the Latex as it looks better in nearly every case (except perhaps the limits on the integral, but that is minor).

> @mitiguy partially reviewed 1 file and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Ogia4CtEEQND0paYWpI:b-f2garh)* status: 8 of 12 files reviewed, 3 unresolved discussions (waiting on @nickbianco and @sherm1).

___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 72 at r8](https://reviewable.io/reviews/simbody/simbody/844#-OgiZMVBAj6cVHrPG2v5:-OgiZMVC3Jehf66G9uaz:b-wxvjsx) ([raw file](https://github.com/simbody/simbody/blob/711e4942ec1fa21cdf4566444af62b034649a9c1/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L72)):*
> ```objc
> relationship between the deflection and angle of the beam's endpoint:
> 
> y = (2⁄3) θ L
> ```

nit  consider y(L)



_Suggestion:_
```objc
y(L) = (2⁄3) θ L
```
___
*[`Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h` line 100 at r8](https://reviewable.io/reviews/simbody/simbody/844#-Ogi_Jj35AuRUbkP1udK:-Ogi_Jj35AuRUbkP1udL:bbsv9jn) ([raw file](https://github.com/simbody/simbody/blob/711e4942ec1fa21cdf4566444af62b034649a9c1/Simbody/include/simbody/internal/MobilizedBody_CantileverFreeBeam.h#L100)):*
> ```objc
> an X-Y-Z body-fixed Euler sequence. The three generalized speeds u for
> this mobilizer are also the same as for a Gimbal mobilizer: the time derivatives
> of the generalized coordinates, that is, u = qdot.
> ```

nit  consider 



_Suggestion:_
```objc
uᵢ = q̇ᵢ (i.e., uᵢ = qdotᵢ), (i = 0, 1, 2).
```

sherm1:
Sorry Nick, holiday break got in the way. I'm back today and Paul is tomorrow. Will look ASAP.

nickbianco:
No worries, just checking in to make sure I wasn't the bottleneck.

sherm1 (approved):
A thought: Paul is concerned about whether the new mobilizer is energy conserving. IMO it has to be from the math. But a test showing that would resolve @mitiguy's worries. You could make a test case that consists of a single body connected to Ground by a CantileverFreeBeam mobilizer, and let it fall under gravity like a weird pendulum. Starts with just gravitational potential energy then falls and the total energy should remain constant to many decimal places. Then maybe Paul will OK a merge!

(Body would have to have mass and inertia, and motion must avoid making the middle angle singular.)

> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Oj2kQTZChX2ZMp-K0C-:b42yeqo)* status: all files reviewed, 1 unresolved discussion (waiting on @mitiguy).

sherm1 (approved):
Beautiful! Thanks Nick. I have one minor suggestion below. Otherwise ready to merge!
:lgtm:

> @sherm1 reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/844#-:-Oj8HJON-mAWZ-28HCeF:bv74y71)* status: all files reviewed, 1 unresolved discussion (waiting on @nickbianco).

___
*[`Simbody/tests/TestMobilizedBody.cpp` line 661 at r11](https://reviewable.io/reviews/simbody/simbody/844#-Oj8Fg9JBCCBR_touFzY:-Oj8Fg9JBCCBR_touFzZ:bohlg46) ([raw file](https://github.com/simbody/simbody/blob/9a906abd5868dbb15f278836190f6196a863ac12/Simbody/tests/TestMobilizedBody.cpp#L661)):*
> ```cpp
>         ts.stepTo(time);
>         const State& s = ts.getState();
>         SimTK_TEST_EQ_TOL(initialEnergy, system.calcEnergy(s), 1e-10);
> ```

BTW Great that it was conserved to 1e-10! However, I'd suggest loosening that a little, say to 1e-9 or even 1e-8, and running a little longer like to 10s rather than 1s (could make dt=1e-2). Loosening the tolerance a little more than you absolutely need is generally a good idea for robustness -- other computers or compilers might cause different (but legit) roundoff errors. I would certainly be convinced if energy is conserved to 1 part in 100 million for 10s!

PR #845, opened 2026-01-26
TinyXML migration to TinyXML2

### LINKED ISSUE #839: TinyXML migration to TinyXML2
[`TinyXML2`](https://github.com/leethomason/tinyxml2) has replaced `TinyXML`. 
There are known memory leaks in the original `TinyXML` described here: https://github.com/simbody/simbody/issues/229 and https://github.com/simbody/simbody/issues/227 .

`TinyXML2` features significantly less allocations according to [their documentation](https://github.com/leethomason/tinyxml2?tab=readme-ov-file#tinyxml-1-vs-tinyxml-2):
```md
TinyXML-2 long been the focus of all development. It is well tested and should be used instead of TinyXML-1.

TinyXML-2 uses a similar API to TinyXML-1 and the same rich test cases.
But the implementation of the parser is completely re-written to make it more appropriate for use in a game. 
It uses less memory, is faster, and uses far fewer memory allocations.

TinyXML-2 has no requirement or support for STL.
```

It would be worthwhile to investigate migrating to `TinyXML2`.

alexbeattie42:
Resolves #839

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/845)


sherm1:
Did you change your mind about this or did it die of neglect?

alexbeattie42:
> Did you change your mind about this or did it die of neglect?

I still believe that this is a beneficial and useful change. It will ultimately be quite a large change when it is fully working. I got the conversion to the point where it would build with the new TinXML2 library and some tests were passing. Unfortunately, it will take considerably more work to get it to the point where it is contribution ready. I don't have the resources to contribute to it currently which is why I closed it.

PR #852, opened 2026-02-27
Make various ctors/constants constexpr

adamkewley:
Fixes issue found when compiling simbody as part of [OpenSim Creator](https://www.opensimcreator.com/) / [OPynSim](https://github.com/opynsim/opynsim), which use a variety of different compilers, linters, libASAN, libUBSAN, etc.  This is part of OPynSim's [custom patchset](https://github.com/opynsim/opynsim/tree/17dbb1df84c70efb306554d386712421d3f60bdc/libosim/simbody-patches) that is applied downstream and ran for a while  (e.g. in OpenSim Creator) before upstreaming here.

The specific issue this addresses is related to static initialization order. Therefore, whether it manifests or not strongly depends on whether the downstream codebase is dynamically or statically linking to simbody. With dynamic linkage, it's "the order in which each binary is loaded (linker-dependent *within* a binary, though)", with static linkage, it's "good luck guessing the initialization order (entirely linker-dependent)".

The static initialization issue is that some constants in `simbody` are defined using `extern` symbols. From a linkage point-of-view, what this means is that downstream code doesn't "see" that (e.g.) `Eps == 0.000000000000001`, it "sees" `load symbol SimTK::Eps -> mov xmm0 [SimTK::Eps]` (or similar). Static initialization does not guarantee that those symbols are populated at that address in the process image before other static initializers are ran. Therefore, when the compiler sees:

```c++
// VisualizerGeometry.cpp
static const Vec3 DefaultBodyColor = Gray;
static const Vec3 DefaultPointColor = Magenta;
```

It generates:

```text
mov [SimTK::DefaultBodyColor] [SimTK::Gray]
```

As a runtime static initializer. This means that `DefaultBodyColor` might be populated with whatever junk/zeroed memory exists at the address of `SimTK::Gray`, rather than a `Gray` color. This is also applies for any static initializers in downstream code when static linking and, indeed, OPynSim/opensim-creator trigger errors when running libASAN/libUBSAN because exactly this problem manifests. Granted, it's a special case because those projects statically compile the universe into one mega binary in order to make shipping easier/smaller.

This PR doesn't fix all known static initialization issues in simbody, only the ones I have detected using runtime analyzers. A robust solution requires going through all constants in the codebase and ensuring that they are either `constexpr` (i.e. in a header) or `constinit` (C++20 feature, which can be used to mean "there's no runtime initialization required for this extern symbol). I figured it's best to PR what I'm currently patching to open a dialogue about gradually fixing it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/852)


adamkewley:
Because i cba'd dancing with GitHub, here's a minimal docker setup to figure out the compiler issues:

`docker/ubuntu22_apt.txt`:
```txt
git            # VCS
cmake          # C++
liblapack-dev  # Simbody: linear algebra
freeglut3-dev  # Simbody: visualizer
libxi-dev      # Simbody: visualizer
libxmu-dev     # Simbody: visualizer
doxygen        # Simbody: documentation
gcc-12         # C++20: newer compiler
g++-12         # C++20: newer compiler
clang-15       # C++20: newer compiler
clang++-15     # C++20: newer compiler
```

`docker/ubuntu22.Dockerfile`:

```dockerfile
FROM ubuntu:22.04

# Avoid interactive prompts during package installs
ENV DEBIAN_FRONTEND=noninteractive

# Copy apt requirements file into the container
COPY ubuntu22_apt.txt /tmp/ubuntu22_apt.txt

# Install system dependencies (+cleanup)
RUN apt-get update && \
    apt-get install -y software-properties-common curl ca-certificates && \
    apt-get install -y $(sed 's/#.*//' /tmp/ubuntu22_apt.txt | xargs) && \
    rm -rf /var/lib/apt/lists/* /tmp/ubuntu22_apt.txt

# Set the default shell
cmd ["/bin/bash"]
```

With that, this works and passes all tests:

```bash
#!/usr/bin/env bash

set -xeuo pipefail

cd docker && docker build -t simbody-ubuntu22 -f ubuntu22.Dockerfile . && cd -
sudo docker run --rm -v "${PWD}:/simbody" -w /simbody -it simbody-ubuntu22 bash -c "CC=clang-15 CXX=clang++-15 cmake --preset dev-default && cmake --build --preset dev-default -j$(nproc) && ctest --preset dev-default -j$(nproc)"
sudo rm -rf out
sudo docker run --rm -v "${PWD}:/simbody" -w /simbody -it simbody-ubuntu22 bash -c "CC=gcc-12 CXX=g++-12 cmake --preset dev-default && cmake --build --preset dev-default -j$(nproc) && ctest --preset dev-default -j$(nproc)"
```

Which is to say, Ubuntu 22 needs gcc12 and clang15 in order to compile simbody with C++20 support. Updating the compiler shouldn't change the runtime libstdc++ version (it should only update the language support).

Given the CI build uses an unversioned clang `CMakePresets.json` preset to build, I'd recommend either using `update-alternatives` to make `clang` point to `clang-15`, or consider dropping a hard-coded compiler as a CMakePreset, and instead lifting it to the CI level by just having a `ci-linux-debug` target and using the traditional environment variables to control it: `CC=clang-15 CXX=clang++-12 cmake --preset ci-linux-debug`.

@sherm1  what would you prefer? Reconfigure the CI image to alias `clang` to `clang-15` or drop the `-clang-` presets and instead leave it up to the caller to specify the compiler via `CC`/`CXX`. For my own projects, I do the latter, entirely because I might use backported compilers, aliases, shims etc. on a per-CI-node basis.

sherm1:
@adamkewley sorry looks like I lost track of this PR. Is it still active?

adamkewley:
@sherm1 iirc it cannot be merged unless a decision is made about changing the presets and/or CI environment to accommodate for the fact that the changeset includes C++20 code that won't compile with Ubuntu 22's default compiler toolchain (see prev. post)

sherm1:
IMO the only substantive constraint is what OpenSim needs to build with. Otherwise I'm fine with making everyone use a more recent compiler. @aymanhab any thoughts?

nickbianco:
OpenSim has CI builds for Ubuntu 22 and Ubuntu 24 with C++20. I'd guess that our CI scripts use whatever default GCC version comes with Ubuntu 22, which would cause the same issues downstream. But if you want to pin the compiler to make it work here, I can do the same in OpenSim.

aymanhab:
Thanks for checking @sherm1 and @adamkewley 👍  Based on my recent experience building python wheels for opensim using specific compilers required by the python ecosystem (manylinux docker containers on linux) It's critical that we "can" compile with these compilers for manylinux_2_28, so sharing the environment with other common tools is possible as needed by an ever-growing community of OpenSim users.

Specific builds for OSC or others can use more restrictive compilers as long as the code still compiles with the list attached here. 
Screenshot 2026-05-06 141620

adamkewley:
@nickbianco / @sherm1 ready for review. I decided to change simbody's Linux CI image to match OpenSim's, to reduce any friction between the two projects (they're basically joined at the hip!).

There was also a couple of minor changes to the C-style casts in the test suite. The reason for those changes is that the CI image uses a newer gcc (gcc-14), which now more strongly prohibits strict-aliasing violations. The casts are safe from a data/access PoV but modern C++ has gotten much more aggressive about type aliasing because optimizers are starting to make more assumptions about lifetimes, data accesses, reference escapes, etc. which might now break.

sherm1 (approved):
Thanks, Adam! Reviewed, :lgtm: except I'd like to see the tricky semantics of no-computation negation and conjugation continue to be verified in some way if possible. In both cases the idea is that changing the interpretation of the existing bits (by changing the type) is sufficient to negate a real or the imaginary part of a complex.

@aymanhab does this all look good from an OpenSim point of view?

> @sherm1 reviewed 9 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-OyEsRk7BkhT-_GTgvmj:b-b84y02)* status: all files reviewed, 3 unresolved discussions (waiting on adamkewley).

___
*[`SimTKcommon/SmallMatrix/include/SimTKcommon/internal/Vec.h` line 1013 at r3](https://reviewable.io/reviews/simbody/simbody/852#-OyEs4wy1xIaYlP4GKrw:-OyEs4wy1xIaYlP4GKrx:bmfuyju) ([raw file](https://github.com/simbody/simbody/blob/a7c576018b4cc4025555324ae9ca1a6d3c0a10fa/SimTKcommon/SmallMatrix/include/SimTKcommon/internal/Vec.h#L1013)):*
> ```objc
>     template<typename ElementsRowByRowTuple, int... Idx>
>     requires (sizeof...(Idx) == M)
>     constexpr void assignElementsFromTuple(ElementsRowByRowTuple&& els, std::integer_sequence<int, Idx...>)
> ```

nit: line length per styleguide (sorry, Adam!)
___
*[`SimTKcommon/tests/MatVecTest.cpp` line 332 at r3](https://reviewable.io/reviews/simbody/simbody/852#-OyEqz0716EM4u9xCOuL:-OyEqz0716EM4u9xCOuM:b-kzj3vy) ([raw file](https://github.com/simbody/simbody/blob/a7c576018b4cc4025555324ae9ca1a6d3c0a10fa/SimTKcommon/tests/MatVecTest.cpp#L332)):*
> ```cpp
>     Vec<2,negator<complex<float> >,1>& negCv2 = -cv2;
>     Vec<2,conjugate<float>,1> conjCv2{cv2};
>     Vec<2,negator<conjugate<float> >,1>& negConjCv2 = -conjCv2;
> ```

minor: IIRC the purpose of these lines was to verify that negation and conjugation could be achieved merely by changing the interpretation of the existing bits and not actually performing any computation (hence the reference types). Some of the changes appear to switch from type casting to computation. Is the problem that the new compilers don't like the clunky old school type conversions? I'd prefer to see that addressed with static_cast if possible.

Really this test should verify that the address of the lhs remains the same as the rhs.

sherm1 (approved):
(Apologies, @adamkewley -- I had reviewed this a while back in Reviewable and apparently forgot to hit Publish!)

> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-OzrGRBt384BmiLmcHZM:b-qzkois)* status: all files reviewed, 3 unresolved discussions (waiting on adamkewley).

aymanhab (commented):
Thanks @sherm1 for checking, as I mentioned above the main concern is if this compiles with c++20 which is required for manylinux_2_28 mentioned above, if so then we're all good. Going further ahead complicates life for users on mainstream python packages that are built with older compiler for wider use.  I believe this has been verified by @adamkewley changing ci to use manylinux_2_28

> @aymanhab reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-OzrNHOz3JmfofNpEftk:b-pzhzc)* status: all files reviewed, 3 unresolved discussions (waiting on adamkewley and sherm1).

aymanhab (commented):
Maybe be good to build opensim against this simbody commit to verify and avoid surprises downstream.

> @aymanhab made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-OzrQ1U555Ngf9r__agP:b7eiwd6)* status: all files reviewed, 3 unresolved discussions (waiting on adamkewley and sherm1).

nickbianco:
> Maybe be good to build opensim against this simbody commit to verify and avoid surprises downstream.

Testing here: https://github.com/opensim-org/opensim-core/pull/4396

nickbianco:
The CI builds for https://github.com/opensim-org/opensim-core/pull/4396 fail, but only because the `Vec<T>` constructors have changed we need to to update the bindings to accommodate. Things look fine on the C++ side.

adamkewley (commented):
@sherm1 thanks for reviewing - I fixed the line lengths and explained why you may not want zero-computation `conjugate` punning. If you desperately believe it's a strong necessity for `simbody` then I'd recommend adding a type-punning `.conjugate` into the API or having tests like `static_assert(sizeof(Conj) == sizeof(NotConj) && alignof(Conj) == sizeof(NotConj))`, but assertions like that don't guarantee as much as they might seem (e.g. what if the byte layout changes but has the same size+alignment?).

> @adamkewley reviewed 9 files and all commit messages, made 3 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-P-UNEqa-C81NgkXd-mO:bqipius)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on sherm1).

___
*[`SimTKcommon/SmallMatrix/include/SimTKcommon/internal/Vec.h` line 1013 at r3](https://reviewable.io/reviews/simbody/simbody/852#-OyEs4wy1xIaYlP4GKrw:-P-UHZt2D6HlpnyjyhhJ:b-ese405) ([raw file](https://github.com/simbody/simbody/blob/a7c576018b4cc4025555324ae9ca1a6d3c0a10fa/SimTKcommon/SmallMatrix/include/SimTKcommon/internal/Vec.h#L1013)):*

nit: line length per styleguide (sorry, Adam!)
</blockquote></details>

Fixed in 02ae6b21
___
*[`SimTKcommon/tests/MatVecTest.cpp` line 332 at r3](https://reviewable.io/reviews/simbody/simbody/852#-OyEqz0716EM4u9xCOuL:-P-UHd6I39apZYKTXFYB:b-vlzkw9) ([raw file](https://github.com/simbody/simbody/blob/a7c576018b4cc4025555324ae9ca1a6d3c0a10fa/SimTKcommon/tests/MatVecTest.cpp#L332)):*

minor: IIRC the purpose of these lines was to verify that negation and conjugation could be achieved merely by changing the interpretation of the existing bits and not actually performing any computation (hence the reference types). Some of the changes appear to switch from type casting to computation. Is the problem that the new compilers don't like the clunky old school type conversions? I'd prefer to see that addressed with static_cast if possible.

Really this test should verify that the address of the lhs remains the same as the rhs.
</blockquote></details>

@sherm1 switching to type-punning (or casting) traditionally behaved fine on older compilers, but it's UB due to the strict aliasing rule. Modern compilers are permitted to rearrange code in incompatible ways if you do something like construct a `SimTK::Vec<T>` and then pun it to a `SimTK::Vec<U>`, even if you "know" `U` has the same byte layout as `T`.

Code that does it with more-modern compilers these days also typically need to combine it with something like `std::launder` (which tells the compiler "I'm laundering memory over here, don't try and optimize around it). IIRC, the only non-UB usages of a bare `reinterpret_cast` are something like std::bytes <--> unsigned char <--> &T (also <--> void*, but that's also possible via `static_cast`).

For this particular change, 2 of the 3 lines are actually improvements. `-cv2` calls:

`const TNeg&  operator-() const { return negate(); }`

Which already codifies a `reinterpret_cast`. Using the in-built negator is an upgrade from the original C cast because simbody is at least explicitly encoding "negation is zero computation, we know what we're doing, leave it to us". However, this line:

Vec<2,conjugate<float>,1>& conjCv2 = (Vec<2,conjugate<float>,1>&)cv2;

Is problematic. `Vec` doesn't have an implicit conversion operator like `operator Vec<CNT, conjugate<T>>&` or a method like `Vec<CNT, conjugate<T>>& conjugate()`. It's relying on a bare C cast to do the check, but the C cast is effectively just a `reinterpret_cast` on the caller's side. `reinterpret_cast` can convert any reference to one type into a reference to another type and is always zero computation, meaning the test wouldn't be testing very much (you could `reinterpret_cast` and `int` to a `Vec` and have it compile and superficially seem ok also).

`static_cast` isn't a suitable alternative for this, because you cannot `static_cast` a reference to one type into a reference to another type. That requires that the outgoing type has an implicit conversion operator to a reference to the new type. You can `static_cast` references up inheritance hierarchies, or between integer types, or even downcast pointers (incl. `void*`). You can even `static_cast` to a completely different **value** if the new value has a constructor from the old value, but you cannot `static_cast` pointers or references to one type into pointers or references to another type (in this case, two separate template instantations of `Vec` are "two types") because that specific use-case is a `reinterpret_cast` - with all the strings attached.

So my recommendation is to keep the negation lines (they're strictly improvements) and perform a value cast (i.e. a construction) for the conugate. If zero-cost conjugate casting is desirable then it should be encoded in simbody's type system (e.g. `Vec::conjugate`), rather than testing it with a unit test. Otherwise, it's an unclear/implicit API design that downstream code really ought to not rely on: "this particular type-pair can be safely type-punned with a `reinterpret_cast`", because it means you'll end up with a bunch of `reinterpret_cast`s in that downstream code that will silently break if you later decide "actually, it requires *a little* computation" and they update+recompile.

sherm1 (approved):
Thanks, Adam. We can live without zero-cost conjugation. Feel free to merge this when ready.

> @sherm1 reviewed 1 file and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/852#-:-P-V7b5k1iPy9z0iOW3w:b-7uxa34)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on adamkewley).

___
*[`SimTKcommon/tests/MatVecTest.cpp` line 332 at r3](https://reviewable.io/reviews/simbody/simbody/852#-OyEqz0716EM4u9xCOuL:-P-V6tLJ6rIHEwS0i0kg:b-haqm0s) ([raw file](https://github.com/simbody/simbody/blob/a7c576018b4cc4025555324ae9ca1a6d3c0a10fa/SimTKcommon/tests/MatVecTest.cpp#L332)):*

@sherm1 switching to type-punning (or casting) traditionally behaved fine on older compilers, but it's UB due to the strict aliasing rule. Modern compilers are permitted to rearrange code in incompatible ways if you do something like construct a `SimTK::Vec<T>` and then pun it to a `SimTK::Vec<U>`, even if you "know" `U` has the same byte layout as `T`.

Code that does it with more-modern compilers these days also typically need to combine it with something like `std::launder` (which tells the compiler "I'm laundering memory over here, don't try and optimize around it). IIRC, the only non-UB usages of a bare `reinterpret_cast` are something like std::bytes <--> unsigned char <--> &T (also <--> void*, but that's also possible via `static_cast`).

For this particular change, 2 of the 3 lines are actually improvements. `-cv2` calls:

`const TNeg&  operator-() const { return negate(); }`

Which already codifies a `reinterpret_cast`. Using the in-built negator is an upgrade from the original C cast because simbody is at least explicitly encoding "negation is zero computation, we know what we're doing, leave it to us". However, this line:

Vec<2,conjugate<float>,1>& conjCv2 = (Vec<2,conjugate<float>,1>&)cv2;

Is problematic. `Vec` doesn't have an implicit conversion operator like `operator Vec<CNT, conjugate<T>>&` or a method like `Vec<CNT, conjugate<T>>& conjugate()`. It's relying on a bare C cast to do the check, but the C cast is effectively just a `reinterpret_cast` on the caller's side. `reinterpret_cast` can convert any reference to one type into a reference to another type and is always zero computation, meaning the test wouldn't be testing very much (you could `reinterpret_cast` and `int` to a `Vec` and have it compile and superficially seem ok also).

`static_cast` isn't a suitable alternative for this, because you cannot `static_cast` a reference to one type into a reference to another type. That requires that the outgoing type has an implicit conversion operator to a reference to the new type. You can `static_cast` references up inheritance hierarchies, or between integer types, or even downcast pointers (incl. `void*`). You can even `static_cast` to a completely different **value** if the new value has a constructor from the old value, but you cannot `static_cast` pointers or references to one type into pointers or references to another type (in this case, two separate template instantations of `Vec` are "two types") because that specific use-case is a `reinterpret_cast` - with all the strings attached.

So my recommendation is to keep the negation lines (they're strictly improvements) and perform a value cast (i.e. a construction) for the conugate. If zero-cost conjugate casting is desirable then it should be encoded in simbody's type system (e.g. `Vec::conjugate`), rather than testing it with a unit test. Otherwise, it's an unclear/implicit API design that downstream code really ought to not rely on: "this particular type-pair can be safely type-punned with a `reinterpret_cast`", because it means you'll end up with a bunch of `reinterpret_cast`s in that downstream code that will silently break if you later decide "actually, it requires *a little* computation" and they update+recompile.
</blockquote></details>

OK. I don't think anyone is really depending on a zero-computation conjugate anyway. At the time I wrote Simbody's math classes I was just enamored with the idea that it was possible to do that! Sigh ... I suppose we can live without it :).

adamkewley:
Great :smile: - and @sherm1 the conjugate stuff could be added in a feature PR if there's some performance corner for it? All i mean is that it should probably be a member method so that you can control the `reinterpret_cast` from `Vec` rather than having downstream code do it because it's less risky from a codebase-evolution PoV (the strict aliasing rules would be violated, though!)

PR #855, opened 2026-03-12
Add `CableSpan::calcResampledDecorativePathPoints()`

nickbianco:
`CableSpan::calcResampleDecorativePathPoints()` is a method for generating decorative path points across an entire `CableSpan`. Unlike `CableSpan::calcDecorativePathPoints()`, points along a curve segment are resampled to desired equal length intervals in the same way as `CableSpan::calcCurveSegmentResampledPoints()`.

This change will serve to address the following OpenSim issue: https://github.com/opensim-org/opensim-core/issues/4279.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/855)

### DIFF
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -515,9 +515,10 @@ public:
         const State& state,
         const std::function<void(Vec3 point_G)>& sink) const;
 
-    /** Compute points on the path spanned by this cable suitable for
-    visualization purposes where the points along each obstacle's curve segment
-    are computed at equal length intervals.
+    /** Compute `numCurveSegmentSamples` equally-spaced points for each curve
+    segment along the path spanned by this cable and emit them to `sink`.
+    This can be useful for visualization purposes, where it can be important to
+    limit the amount of visualized path points for performance reasons.
     State must be realized to Stage::Position.
     @param state System State.
     @param numCurveSegmentSamples The number of samples to take per curve
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -515,6 +515,19 @@ public:
         const State& state,
         const std::function<void(Vec3 point_G)>& sink) const;
 
+    /** Compute points on the path spanned by this cable suitable for
+    visualization purposes where the points along each obstacle's curve segment
+    are computed at equal length intervals.
+    State must be realized to Stage::Position.
+    @param state System State.
+    @param numCurveSegmentSamples The number of samples to take per curve
+                                  segment.
+    @param sink Where the path points (in ground frame) will be written to. **/
+    void calcResampledDecorativePathPoints(
+        const State& state,
+        int numCurveSegmentSamples,
+        const std::function<void(Vec3 point_G)>& sink) const;
+
     ///@}
 
     /** @name Curve segment computations */
--- a/Simbody/src/CableSpan.cpp
+++ b/Simbody/src/CableSpan.cpp
@@ -2182,6 +2182,38 @@ public:
         int numSamples,
         const std::function<void(Vec3 point_G)>& sink) const;
 
+    void calcResampledDecorativePathPoints(
+        const State& state,
+        int numCurveSegmentSamples,
+        const std::function<void(Vec3 point_G)>& sink) const
+    {
+        // Write the initial path point.
+        const CableSpanData::Position& dataPos = getDataPos(state);
+        sink(dataPos.originPoint_G);
+
+        // Write the path points for each cable segment.
+        for (CableSegmentIndex ix(0); ix < getNumCableSegments(); ++ix) {
+            const CableSegment& cableSegment = getCableSegment(ix);
+
+            // Write the path points for each obstacle in the cable segment.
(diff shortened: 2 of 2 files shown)


adamkewley (approved):
@nickbianco nice work, I think it's looking good.

My comments are more like stylistic musings than anything important. If the implementation works the way you think it should etc. then don't let those comments stop you 
:lgtm:

> @adamkewley reviewed 2 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/855#-:-OnaVJYQAvVneYYUVKbB:bxoibm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 518 at r1](https://reviewable.io/reviews/simbody/simbody/855#-OnaORQeB_rcET6V-AAL:-OnaORQeB_rcET6V-AAM:b-f88dlu) ([raw file](https://github.com/simbody/simbody/blob/c89caee451a79a368b8142afc13d71258827dab3/Simbody/include/simbody/internal/CableSpan.h#L518)):*
Nice explanation. Personally, I'd prefer describing the behavior first and then mentioning the visualization benefits as a note, rather than a top-line API description:

> Computes `numCurveSegmentSamples` equally-spaced points for each curve segment along the path spanned by this cable and emits them to `sink`.
>
> This can be useful for visualization purposes, where it can be important to limit the amount of visualized path points for performance reasons.
>
> @param state System State. Must be realized to at least Stage::Position.
> etc.

But it's entirely up to you. It's just that I prefer the first/second sentence of an API string to describe the functionality and then the rest can mention benefits, edge-cases, etc.
___
*[`Simbody/src/CableSpan.cpp` line 2204 at r1](https://reviewable.io/reviews/simbody/simbody/855#-OnaPwCw5-G5L14vnpiz:-OnaPwCw5-G5L14vnpj-:b-3mf6u8) ([raw file](https://github.com/simbody/simbody/blob/c89caee451a79a368b8142afc13d71258827dab3/Simbody/src/CableSpan.cpp#L2204)):*
> ```cpp
>             }
> 
>             // Write the path points for the final via point in the cable
> ```

The singular/plural dual is a little unusual here. You're saying "write the path POINTS for the final via POINT" - is there one final via point or many final via points? If there's many, are they equally final, etc.?

If it's the simple case, isn't `calcDecorativePathPoints` effectively the same as asking for the final via point station/location and throwing that through `sink`? Why the dance with `calcDecorativePathPoints`? I'm assuming it's plurally named to try and be consistent with path elements (segments, etc.) that may emit multiple points, rather than because a single via point is capable of generating multiple decorative points?
___
*[`Simbody/src/CableSpan.cpp` line 4806 at r1](https://reviewable.io/reviews/simbody/simbody/855#-OnaQeW_36DKC5GTwjx4:-OnaQeW_36DKC5GTwjx5:boi2pt) ([raw file](https://github.com/simbody/simbody/blob/c89caee451a79a368b8142afc13d71258827dab3/Simbody/src/CableSpan.cpp#L4806)):*
> ```cpp
>     const std::function<void(Vec3 point_G)>& sink) const
> {
>     getImpl().calcResampledDecorativePathPoints(state, numCurveSegmentSamples,
> ```

I get there's some automated formatting going on here but line-breaking a function's arguments like this because a text formatter has decided it doesn't make maintaining the code easier.

Remember, a code formatter isn't responsible for maintaining the code, developers are. Developers read code like they read the newspaper, or Harry Potter, not like how clang-format reads it. Which would you prefer to actually read?:

```c++
 getImpl().calcResampledDecorativePathPoints(state, numCurveSegmentSamples,
     sink);

 getImpl().calcResampledDecorativePathPoints(
     state, numCurveSegmentSamples, sink);

 getImpl().calcResampledDecorativePathPoints(
     state,
     numCurveSegmentSamples,
     sink
  );

// oooh nooo, naughty naughty
 getImpl().calcResampledDecorativePathPoints(state, numCurveSamples, sink);

```

This isn't a blocking point, just one that's worth pointing out, because maintenance is something like 80 % of development in legacy projects (by legacy, I mean: still useful, still relevant, but older and with backwards-compatibility concerns, not "bad").

sherm1 (approved):
:lgtm: too -- just need to respond to Adam's comments then ready to merge

> @sherm1 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/855#-:-OniNlKRAq3wp4ppXB04:b53s4u4)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

adamkewley (approved):
> @adamkewley reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/855)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

sherm1 (approved):
> @sherm1 reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/855)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

nickbianco:
Thanks @adamkewley and @sherm1!

PR #857, opened 2026-05-29
Add `CableSpan` path tangent direction calculation methods

nickbianco:
This change adds methods to `CableSpan` for calculating path tangent directions at origin, termination, and via points. This will make it easier for downstream applications to calculate force contributions from a `CableSpan` as a series of point forces (see https://github.com/opensim-org/opensim-core/pull/4331).

The methods implemented return point forces in the direction of cable tension, e.g., along the path away from origin, termination, and via points. This is most intuitive to me, but it is different from the way `CableSpan::calcCurveSegmentInitialFrenetFrame()` and `CableSpan::calcCurveSegmentFinalFrenetFrame()` return Frenet frames, where the tangent axis (x-axis) of each frame both point going from the origin to the termination. I'm happy to switch the convention for the tangent direction calculations if we think that is better.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/857)

### DIFF
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -307,6 +307,13 @@ public:
         const State& state,
         SpatialVec& unitForce_G) const;
 
+    /** Calculate the tangent direction of the cable span at the origin. This is
+    the direction along the path away from the origin point, i.e., the direction
+    of cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    **/
+    UnitVec3 calcOriginTangentDirection(const State& state) const;
+
     /** Calculate the unit force exerted by the cable at the cable termination
     (in ground frame). The force can be obtained by multiplying the result by
     the cable tension. State must be realized to Stage::Position.
@@ -316,6 +323,13 @@ public:
     void calcTerminationUnitForce(
         const State& state,
         SpatialVec& unitForce_G) const;
+
+    /** Calculate the tangent direction of the cable span at the termination.
+    This is the direction along the path away from the termination point, i.e.,
+    the direction of cable tension. State must be realized to Stage::Position.
+    @param state State of the system.
+    **/
+    UnitVec3 calcTerminationTangentDirection(const State& state) const;
 
     ///@}
 
@@ -639,6 +653,26 @@ public:
         const State& state,
         CableSpanViaPointIndex ix,
         SpatialVec& unitForce_G) const;
+
+    /** Calculate the initial tangent direction of the cable span at the
+    specified via point. This is the direction along the path from the via point
+    towards the origin point, i.e., the direction of cable tension. State must
+    be realized to Stage::Position.
+    @param state State of the system.
+    @param ix The index of the via point in this CableSpan. **/
+    UnitVec3 calcViaPointInitialTangentDirection(
+        const State& state,
+        CableSpanViaPointIndex ix) const;
+
+    /** Calculate the final tangent direction of the cable span at the specified
+    via point. This is the direction along the path from the via point towards
+    the termination point, i.e., the direction of cable tension. State must be
+    realized to Stage::Position.
+    @param state State of the system.
+    @param ix The index of the via point in this CableSpan. **/
+    UnitVec3 calcViaPointFinalTangentDirection(
+        const State& state,
+        CableSpanViaPointIndex ix) const;
 
     ///@}
 
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -307,9 +307,9 @@ public:
         const State& state,
         SpatialVec& unitForce_G) const;
 
-    /** Calculate the tangent direction of the cable span at the origin. This is
-    the direction along the path away from the origin point, i.e., the direction
-    of cable tension. State must be realized to Stage::Position.
+    /** Calculate the cable span tangent direction at the origin. The direction
+    points along the path from the origin towards the termination.
+    State must be realized to Stage::Position.
     @param state State of the system.
     **/
     UnitVec3 calcOriginTangentDirection(const State& state) const;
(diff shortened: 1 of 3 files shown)


nickbianco:
@adamkewley assigning you a review here since this will benefit OSC downstream via https://github.com/opensim-org/opensim-core/pull/4331.

adamkewley:
All seems like sensible stuff to ask from the `CableSpan` lgtm, merge it if it's holding you up, etc.

The only thing I wasn't so sure about is whether the API is saying too much. There's a tangent vector from the origin to the next via point (e.g. A ---> B) and a then there's a tangent vector from each via point towards the origin (e.g. C --> B --> A) and then there's a tangent from the last via point to the insertion point (e.g. B --> C) but I'm guessing, given they're all unit vectors from each point, that this could all be exposed as `calcViaPointTangentDirectionTowardsOrigin` and the caller could do all the shuffling/negating etc. - it's more a question of where the complexity should be (upstream vs. downstream).

nickbianco:
Thanks @adamkewley. I've redefined the new methods so that the tangent directions are always point along the path from the origin to the termination. I've also renamed the via point methods to use "incoming"/"outgoing" instead of "initial"/"final" which I think is clearer and consistent with our internal terminology. I've also added unit testing.

sherm1 (approved):
:lgtm: with one clarification request

> @sherm1 reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/857#-:-Ou34A_Z4lWo5Kr_eKxw:b-7cb12q)* status: all files reviewed, 1 unresolved discussion (waiting on adamkewley and nickbianco).

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 332 at r2](https://reviewable.io/reviews/simbody/simbody/857#-Ou32XCe-R4x0A1YqWx7:-Ou32XCf8cEcVoJHH6eq:b-w2c2yx) ([raw file](https://github.com/simbody/simbody/blob/466bcd1b1d9b82bb0380cd0b5ce0d438b5a6e93c/Simbody/include/simbody/internal/CableSpan.h#L332)):*
> ```objc
>     @param state State of the system.
>     **/
>     UnitVec3 calcTerminationTangentDirection(const State& state) const;
> ```

BTW I'm wondering how this relates to the existing calcTerminationUnitForce() method above. I'm guessing that if the unit force is (0, f) at the termination point then this method return f? Or is it -f? Or am I misunderstanding? Can you clarify? Probably its the UnitForce method that is ambiguous.

nickbianco (commented):
> @nickbianco made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/857)* status: all files reviewed, 1 unresolved discussion (waiting on adamkewley and sherm1).

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 332 at r2](https://reviewable.io/reviews/simbody/simbody/857#-Ou32XCe-R4x0A1YqWx7:-Ou38A5t7K4iNGx9fxVw:b-3mgbqb) ([raw file](https://github.com/simbody/simbody/blob/466bcd1b1d9b82bb0380cd0b5ce0d438b5a6e93c/Simbody/include/simbody/internal/CableSpan.h#L332)):*

BTW I'm wondering how this relates to the existing calcTerminationUnitForce() method above. I'm guessing that if the unit force is (0, f) at the termination point then this method return f? Or is it -f? Or am I misunderstanding? Can you clarify? Probably its the UnitForce method that is ambiguous.
</blockquote></details>

calcTerminationUnitForce() will apply a unit force in the opposite direction of this method. So if the unit force is (0, f), then this method would return -f. 

Should I add some clarification to the unit force calculation docs to explain which direction forces are being, emphasizing that it is based on a tensile force?

sherm1 (approved):
> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/857)* status: all files reviewed, 1 unresolved discussion (waiting on adamkewley and nickbianco).

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 332 at r2](https://reviewable.io/reviews/simbody/simbody/857#-Ou32XCe-R4x0A1YqWx7:-Ou3NEki9eaBGBrqIbj6:b-yqx2jx) ([raw file](https://github.com/simbody/simbody/blob/466bcd1b1d9b82bb0380cd0b5ce0d438b5a6e93c/Simbody/include/simbody/internal/CableSpan.h#L332)):*

calcTerminationUnitForce() will apply a unit force in the opposite direction of this method. So if the unit force is (0, f), then this method would return -f. 

Should I add some clarification to the unit force calculation docs to explain which direction forces are being, emphasizing that it is based on a tensile force?
</blockquote></details>

Yes, I think that's what's missing. Your new documentation is perfect but the old doc is ambiguous. Good to fix that opportunistically while you're touching that code.

sherm1 (approved):
> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/857)* status: all files reviewed, 1 unresolved discussion (waiting on adamkewley and nickbianco).

___
*[`Simbody/include/simbody/internal/CableSpan.h` line 332 at r2](https://reviewable.io/reviews/simbody/simbody/857#-Ou32XCe-R4x0A1YqWx7:-Ou3NbHR4aujIsikyS91:b-22aggy) ([raw file](https://github.com/simbody/simbody/blob/466bcd1b1d9b82bb0380cd0b5ce0d438b5a6e93c/Simbody/include/simbody/internal/CableSpan.h#L332)):*

Yes, I think that's what's missing. Your new documentation is perfect but the old doc is ambiguous. Good to fix that opportunistically while you're touching that code.
</blockquote></details>

Also I'd be inclined to mention in each routines' docs that the direction is opposite the other method. That's likely to be a surprise.

nickbianco:
@sherm1 updated the unit force docs. Let me know what you think.

sherm1 (approved):
Looks great, Nick -- thanks! Ready to merge?

> @sherm1 reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/857#-:-Ou3yKhX8BaoifixMgvA:b-y3nm4s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on adamkewley).

nickbianco:
Yep! Ready when you are.

PR #858, opened 2026-07-01
CableSpan: add support for enabling/disabling solver warm starts

nickbianco:
### Motivation
This change is motivated by [a recently discovered bug in OpenSim](https://github.com/opensim-org/opensim-core/issues/4330), where we found that moment arm calculations could change based on different default values for a coordinate. A path would first compute an initial solution at the default coordinate values. Then, when computing moment arms at coordinate values far from the default, the moment arm calculation would differ from when the default coordinate value is near the coordinate value used to calculate the moment arm. (In OpenSim Creator, this would manifest as moment arm values "flipping signs" when moving the coordinate slider, since moment arms would be recalculated treating the value on the slider as the coordinate's default.)

### Summary of changes
This change adds the accessors `getUseWarmStart()` and `setUseWarmStart()` to `CableSpan`, which internally controls a flag on `CableSpanParameters` to dictate when path warm starts are used (or not). If enabled, the auto-update discrete variable containing a curve's wrapping state is populated from the previous path solution (i.e., the current behavior). If disabled, it instead uses a plain cache entry at `Stage::Instance` to store intermediate path solutions which get (manually, see below) wiped on `realizePosition()` to force the path to always start from the contact hint. 
 
### Looking for feedback
When warm starts are disabled, the current approach must manually wipe the instance cache data on every `realizePosition()` call to get the desired behavior (always solve from the contact hints). Ideally, the cache entry would get wiped automatically, but setting the "earliest" stage to `SimTK::Position` makes the cache value via `getCacheEntry()` inaccessible since the position stage hasn't yet been reached during while the path solver is computing the optimal path. The alternative to the current approach is to instead call `updCacheEntry()` within `getDataInst` which will allow access to the cache value before hitting `Stage::Position`. I'm happy to switch if desired, but since call a `upd` method within `get` feels wrong, I've opted for the current approach which manually clears the instance cache data.

### Drive-by CI change

I've updated the generator in the Windows CMake preset to `Visual Studio 18 2026`, since the GitHub Actions `windows-latest` runner no longer supports `Visual Studio 17 2022`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/858)

### DIFF
--- a/Simbody/include/simbody/internal/CableSpan.h
+++ b/Simbody/include/simbody/internal/CableSpan.h
@@ -138,17 +138,19 @@ straight line. Similarly the cable can touchdown on the obstacle if the surface
 obstructs the straight line segment again. The cable will always slide freely
 through any via points present in the path.
 
-The path is computed as an optimization problem using the previous optimal path
-as the warm start. This is done by computing natural geodesic corrections for
-each curve segment to compute the locally shortest path, as described in the
-following publication:
+The path is computed as an optimization problem by computing natural geodesic
+corrections for each curve segment to compute the locally shortest path,
+as described in the following publication:
 
     Scholz, A., Sherman, M., Stavness, I. et al (2015). A fast multi-obstacle
     muscle wrapping method using natural geodesic variations. Multibody System
     Dynamics 36, 195–219.
 
 The overall path is locally the shortest, allowing winding over an obstacle
-multiple times, without flipping to the other side.
+multiple times, without flipping to the other side. The wrapping solver can
+optionally use the solution from the previous time step as a warm start to
+improve performance during forward dynamics simulations (see method
+`getUseWarmStart()` for more information).
 
 During initialization the path is assumed to be in contact with each obstacle at
 the user defined contact-point-hint. At each contact point a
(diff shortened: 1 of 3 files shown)


nickbianco:
@adamkewley would you be interested/available for a review?

adamkewley (approved):
Seems like a reasonable thing to customize. I'd guess that OpenSim should expose the behavior via a runtime property (defaulted to "don't warm-start"), so that it behaves similarly to the existing wrapping code but has a performance/behavioral flag?
:lgtm:

> @adamkewley reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/858#-:-Ox-krN_9RWomfQT_UXF:b-pkq36s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

nickbianco:
@adamkewley thanks! And yep, I've already got that started here: https://github.com/opensim-org/opensim-core/pull/4342.

nickbianco:
@sherm1 could we merge this?

sherm1 (approved):
:lgtm:, thanks Nick!

> @sherm1 reviewed 4 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/858#-:-OyZSqYU0y1miCvE7W5_:bqc6cmk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on nickbianco).

PR #860, opened 2026-07-22
[portability] Fix scalar constant SIOF: make `NTraits` getters and `Scalar` constants `constexpr`

helmesjo:
> **Note:** These changes were extracted from a build2 port of Simbody. All fix
> genuine issues in the upstream codebase and none is build2-specific.

## What

Makes the `NTraits<T>` scalar getters and the `Scalar.cpp` global constants
`constexpr`, eliminating the Static Initialization Order Fiasco (SIOF) for the
core SimTK numeric constants.

### NTraits.h

- Marks all `NTraits<T>::get*()` functions `constexpr`, including the
  `sqrt`-derived getters `getSignificant`, `getSqrtEps`, and `getTiny`. Those
  three are implemented via `SimTK::constexpr_sqrt` (see `Math.h` below), which
  on C++23 and later (`__cpp_lib_constexpr_cmath >= 202202L`) delegates to
  `std::sqrt`, and on earlier standards uses an IEEE 754 biased-exponent seed
  followed by Newton-Raphson iteration with correct handling of all IEEE 754
  special cases (NaN, negative, +/-0, +inf, subnormals).
- Rewrites `getLosslessNumDigits` as a closed-form expression using only
  compile-time constants from `std::numeric_limits`, making it proper `constexpr`
  under C++20 with no conditionalization.

### Scalar.cpp

- Changes all definitions (`SimTK::NaN`, `SimTK::Infinity`, `SimTK::Eps`,
  `SimTK::SqrtEps`, `SimTK::TinyReal`, `SimTK::SignificantReal`, `SimTK::Pi`,
  `SimTK::Zero`, `SimTK::One`, and all remaining numeric constants) from `const`
  to `constexpr`.

### SimTKcommon/internal/Math.h (new)

- Adds `SimTK::constexpr_sqrt(double)` and `SimTK::constexpr_sqrt(float)`, the
  compile-time square root helpers used by the `NTraits<T>` getters above.

## Why

`SimTK::NaN`, `SimTK::Infinity`, and the other `Scalar.cpp` constants are
defined as `extern const Real` with dynamic initialization: each calls a
`NTraits<Real>` getter that returns a function-local static. Dynamic
initialization is not ordered across translation units. In a static-library
build, `RegisterTypes_osim*` translation units in opensim-core construct default
component instances (via `constructProperties()`) before `Scalar.cpp`'s
initializers run, so any property that defaults to `SimTK::NaN` or
`SimTK::Infinity` silently reads as zero.

Making the `NTraits` getters `constexpr` allows the downstream `Scalar.cpp`
constants to be given constant initialization, which the C++ standard guarantees
runs before any dynamic initialization regardless of link order.

The changes have no observable effect in shared-library builds, where
initialization order is well-defined.

## How to test

Build Simbody as a shared library and run its test suite to confirm no
regression:

    cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
    cmake --build build -j4
    ctest --test-dir build -C Release --output-on-failure

To exercise the SIOF fix end-to-end, build opensim-core as a static library
against a static Simbody install using this branch and run its test suite. All
property defaults that use `SimTK::NaN` or `SimTK::Infinity` should read
correctly. This has been validated: 100% pass rate on Windows with MSVC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/860)


adamkewley:
@helmesjo , are you ok?

helmesjo:
@adamkewley Barely, it's been quite a few iterations with all these fixes (including OpenSim).

adamkewley:
Alright,

- The cmake-level install/shared lib etc. stuff looks fine but should be handled separately from the const-ing and static initialization order fiasco fixes etc.
- The changes to the initializers make sense, but are almost 1:1 handled by #852 , which is where that should be handled. The blocker for that is just getting the CI system updated to support C++20, which needs to happen anyway. #852 is a little safer because it uses C++20 to handle the bounds checking of the Vec initializer at compile-time (via `requires`) whereas this changeset will only cause failures at runtime (and only if `assert` is enabled).
- The `ntraits_sqrt` `constexpr` sqrt stuff is cool but might be a bit too much faffing around to fix one getter - unsure. If it were to be merged then I'd recommend doing it as a `consteval` function in a separate header or somewhere less specific to vector maths.

helmesjo:
@adamkewley 
- Yeah, makes sense for the CMake stuff. I'll move it to a separate PR.
- Ok, just let me know what should be dropped (or all?). The stuff handled in this PR does (manually confirmed) make OpenSim + Simbody (both static) work (all tests pass, practically none without the fixes).

adamkewley:
Yeah if it's all compiling etc. then we could merge this and separately handle the C++20 requires etc.

On reflection, the constexpr/consteval sqrt is a nice thing to have, but it feels a little bit gross to put it where it is just to satisfy that one usage point, but there might be a nicer home for it somewhere else - I'll have to look around a bit tomorrow/Friday and get back.

adamkewley:
CI failure is because there's an OOB assignment, which is to be expected, because the compiler is literally compiling an OOB access :smile: - IIRC it might've also been a motivation for eliminating it at compile-time with a C++20 `requires` constraint.

helmesjo:
I'll have a look tomorrow, but if you think that it'd be easier to snag anything from this PR (if necessary) over to the other existing feel free! Then I'll just close this (and open a second with just the CMake stuff).

adamkewley:
@helmesjo

#852 fixes some of the `constexpr` stuff and should hopefully be merged soon (I fixed CI etc. to use a newer compiler) - it just needs a once-over by another core developer (because it's a bigger change, affects CI, etc.).

#861 handles the CMake part of this PR, so my suggestion to you is to change this PR to be "Make NTraits/Scalar `constexpr`" and only cover the changeset that touches those two parts (they're related). I can also look into doing this later, after the other two are handled.

helmesjo:
@adamkewley I made the changes (as I understood them), and added `internal/Math.h` for the `constexpr_sqrt()` functions (and just copy-pasted some existing copyright header).

adamkewley:
Yes this is good - I'll still need to read through it but it's much  easier to review etc. now that it's just changing NTraits/Scalar. There might be a few changes needed but I can try building this locally and maybe suggest anything as a diff you can just whack into this PR

adamkewley:
@helmesjo I think e121132 (pushed here) is how I'd handle removing all static initialization issues in `NTraits`/`Scalar`:

- I moved the `constexpr_sqrt` back into `NTraits` - when I looked around I figured, on balance, your original placement is cleaner than adding a new `Math.h` header for the job.
- I changed `constexpr_sqrt` to `consteval_sqrt` to guarantee that it's only ever used at compile-time, because it's a slower method on non-C++23 compilers and we definitely don't want to risk a runtime regression.
- The various legacy `static` initializers in `NTraits<T>` were removed so that almost all functions now just return a `constexpr` value.
  - **This is a breaking change, but one that should (imo) be done if possible - I'll check if OpenSim has any issues with it (unlikely)**.
  - The original design where `NTraits<T>` functions return references to static storage has drawbacks. It would almost certainly be pessimized by a modern C++ compilers because reference escape analysis may notice that unseen code in another compilation unit is also permitted to write something like `const_cast<double&>(NTraits<double>::getEps()) = 1.0;`, so it can't know at compile-time what the value is.
- The constants in `Scalar.h` were changed into `inline constexpr` values. Similar logic. The original design of `extern`ing them means that the compiler always has to (e.g.) `mov xmm0 [SimTK::Eps]` (i.e. a memory load, potentially via the GOT when simbody is dynamically loaded).
- Any mentions of "memory address" etc. in this part of the codebase were removed. Storing the memory address of constants that typically fit into a couple of CPU registers isn't useful. It's much more valuable that a compiler can see, with certainty, that a procedure returns a *copy* of a known value that no other code could possibly modify (for escape analysis - it means the compiler knows it can eliminate the load entirely).
- Some `#include`s were rearranged because, beforehand, there were situations where `String.h` forward-declared one of these constants - that isn't permitted when porting to `constexpr` (the compiler must know the value).

@nickbianco / @sherm1 if you're available, you could review this? I imagine @sherm1 has opinions about whether it's a good idea to change the reference behavior of the `NTraits<T>` API etc. (as in, he'd know the concrete use-cases where it's necessary, etc.).

I'm (clearly) of the strong opinion that simbody should try and move as much code like this as possible to `constexpr`, so that modern compilers can delete/inline/optimize most of it!

nickbianco (commented):
Did a code review and all changes LGTM. Though, per @adamkewley's comment, I'll leave the final call for @sherm1. 

> @nickbianco reviewed 7 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860#-:-Oye7KjEARBY6WoFQNJT:biw19m1)* status: all files reviewed, 1 unresolved discussion (waiting on helmesjo).

___
*[`SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h` line 68 at r1](https://reviewable.io/reviews/simbody/simbody/860#-Oye3a89DLk2c0aZRozI:-Oye3a89DLk2c0aZRozJ:b3nv8k) ([raw file](https://github.com/simbody/simbody/blob/e121132597f90d2eb023396f1a97523ea42022c6/SimTKcommon/Scalar/include/SimTKcommon/internal/NTraits.h#L68)):*
> ```objc
> // IEEE 754 biased-exponent halving trick. The fallback handles all IEEE 754
> // special cases but may be off by 1 ulp from the correctly-rounded value.
> consteval double consteval_sqrt(double x) noexcept {
> ```

I'm assuming the details of the `consteval_sqrt` methods are implemented faithfully from the IEEE 754 standard and won't pour over the details here. Although it's always fun to see bit shifting in the wild (I knew those random YouTube videos I watched about [the Quake algorithm](https://en.wikipedia.org/wiki/Fast_inverse_square_root) would payoff some day). :D

helmesjo:
What do we think, is this in good enough shape for a merge?

adamkewley:
@sherm1 is there any comments/review points you'd like to make? I have tried to address the points from copilot?

sherm1 (approved):
Awesome! :lgtm_strong: pending two requests:
- check with @aymanhab that this doesn't break OpenSim (if you haven't already)
- please add a unit test for the consteval_sqrt() just to sanity check that it returns the same result as std::sqrt(). Of course it will in recent compilers but in case the hand implementations fail on some older compiler it would be good to catch that with a test failure rather than some obscure wrong answer!

> @sherm1 reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860#-:-OzrINbiDFifQ41XB0XO:b9rar5)* status: all files reviewed, 1 unresolved discussion (waiting on helmesjo).

sherm1 (approved):
Oh, and #852 is still pending. Should that merge first?

> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860#-:-OzrJdKwEZ0rGzSPCJIb:bscpkj5)* status: all files reviewed, 1 unresolved discussion (waiting on helmesjo).

aymanhab:
@helmesjo Would be good to build OpenSim against this commit to make sure nothing breaks. The commit is located here https://github.com/opensim-org/opensim-core/blob/18e99c9d185ea638c89c91425b6a1165756d8785/dependencies/CMakeLists.txt#L196
Just to make sure our current supported compilers as well as SWIG can digest the updated syntax

helmesjo:
> @helmesjo Would be good to build OpenSim against this commit to make sure nothing breaks. The commit is located here https://github.com/opensim-org/opensim-core/blob/18e99c9d185ea638c89c91425b6a1165756d8785/dependencies/CMakeLists.txt#L196 Just to make sure our current supported compilers as well as SWIG can digest the updated syntax

Would it work to create a phony PR in OpenSim that changes the Simbody hash to this branch head, thus queuing the full CI that way? I've since cleaned up my comp and would really love not to install the various deps/set it all up again...

**Edit.**
https://github.com/opensim-org/opensim-core/pull/4391

helmesjo:
> > @helmesjo Would be good to build OpenSim against this commit to make sure nothing breaks. The commit is located here https://github.com/opensim-org/opensim-core/blob/18e99c9d185ea638c89c91425b6a1165756d8785/dependencies/CMakeLists.txt#L196 Just to make sure our current supported compilers as well as SWIG can digest the updated syntax
> 
> Would it work to create a phony PR in OpenSim that changes the Simbody hash to this branch head, thus queuing the full CI that way? I've since cleaned up my comp and would really love not to install the various deps/set it all up again...
> 
> **Edit.** [opensim-org/opensim-core#4391](https://github.com/opensim-org/opensim-core/pull/4391)

All green.

aymanhab:
Looks great, thanks for testing @helmesjo Please go ahead and close this PR and merge the simbody PR. Thank you 👍

helmesjo:
@aymanhab Great! I don't have the permission to complete this though, but feel free whoever can.

sherm1 (approved):
Thanks, @helmesjo & @aymanhab. There is still a pending reviewer request that I didn't see addressed:
> please add a unit test for the consteval_sqrt() just to sanity check that it returns the same result as std::sqrt(). Of course it will in recent compilers but in case the hand implementations fail on some older compiler it would be good to catch that with a test failure rather than some obscure wrong answer!

Ready to merge after that.

> @sherm1 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860#-:-P-0tX-H5nBY7S1Jn9VJ:b-6fa9tw)* status: all files reviewed, 1 unresolved discussion (waiting on helmesjo).

helmesjo:
@sherm1 Alright, understood. @adamkewley, do you still have the environment setup and could squeeze in a unit test for the new `consteval_sqrt()` against `std::sqrt()`? Would be greatly appreciated! 🥇

adamkewley:
Made necessary changes and **will merge early next week if there's no contention** (no rush because it still needs follow-up integration work in OpenSim, which I'll try and do next week):

- Refactored the `consteval_sqrt` implementation for readability in [4ef1a2b](https://github.com/simbody/simbody/pull/860/commits/4ef1a2b0bc0247c808934c0bc30f954cccbad45d)
- Added test for `consteval_sqrt` in [ce3c22f](https://github.com/simbody/simbody/pull/860/commits/ce3c22fd676ab2b6f61621e5d817e52c4cb515f7). Notably, the fallback solver can be off by 1ULP due to how the Newton-Rhapson solver might round the last digit. For information-dense square roots that are irregular and would certainly continue to the ULP (e.g. pi or 2), it's observable. The test cannot assert `result == std::sqrt(input)` and, instead, has to check that the result is *within 1 ULP*.
- A 1 ULP rounding difference here should be interpreted as off by a single step in a 64-bit `double`, or the 15-17th decimal place, or `0.0000000000000001`.

I haven't worked with low-level numerics for a while, so here's what my local friendly LLM has to say on the matter:

> Double-precision floats store numbers using 53 bits of precision (~15 to 17 decimal digits). Every floating-point number has a discrete "neighbor" directly above and below it. For $\sqrt{2}$:
> - True Infinite Value: 1.414213562373095048801688724...
> - Closest IEEE 754 double (std::sqrt(2)): 1.4142135623730951
> - Adjacent Neighbor (+1 ULP): 1.4142135623730954
>
> The physical difference between those two adjacent double values is:
>
> $$\Delta = 0.00000000000000022204... \quad (2.22 \times 10^{-16})$$
>
> To put a $10^{-16}$ relative error into perspective:
> - Measuring the Earth: If you measured the circumference of the Earth at the equator (~40,075 km) with 1 ULP error, your answer would be off by less than 10 nanometers (roughly the width of a single strand of DNA).
> - Distance to the Moon: If you calculated the distance from Earth to the Moon (~384,400 km), a 1 ULP error is off by 60 nanometers.
> - Financial Calculations: On a $1,000,000,000 portfolio, 1 ULP is 0.00002 cents.
>
> Gemini (Flash 3.6)

sherm1 (approved):
That's definitely close enough! Thanks, Adam.
BTW 1 ULP is a very strict test. If that causes trouble on any supported machine we could relax it a little and still have it be a convincing test. Still :lgtm_strong: -- could squash & merge any time IMO.

> @sherm1 reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860#-:-P-_ojVAEsJEaAazmFIj:brk3a9x)* status: all files reviewed, 1 unresolved discussion (waiting on helmesjo).

sherm1 (approved):
> @sherm1 resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on helmesjo).

sherm1 (approved):
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/860)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on helmesjo).

adamkewley:
Squashed my changes + fixups etc. into a single commit but left the original commits intact - will merge after final CI

PR #861, opened 2026-07-23
Move vendored library installation to Platform/

adamkewley:
Related to #860 

When building with `BUILD_SHARED_LIBS=OFF`, the libBLAS/libLAPACK libraries aren't installed because the existing method (IMPORTED_RUNTIME_ARTIFACTS) and other alternatives (e.g. `TARGET_RUNTIME_DLLS`) only work on dynamic targets.

This change instead moves the installation step to `Platform/` where it's known, for sure, that the build is using a vendored libBLAS/libLAPACK (the only vendored dynamic libraries in simbody).

This changes behavior slightly: in the old version, if a caller sets their own libBLAS/libLAPACK (e.g. from OpenBLAS) then cmake will no longer automatically bundle those libraries into the install. This is (imo) acceptable for a library target, because it's up to application developers to figure out what actually needs to be in the user-facing bundle.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/861)


adamkewley:
Note: I also double-checked this on a Windows machine. The resulting installation directories are identical, but the static build now also copies the dlls into bin/, which is within expectations.

halleysfifthinc (approved):

adamkewley:
Thanks for the quick check @halleysfifthinc !

PR #862, opened 2026-07-23
Remove hard-coded Visual Studio version in CMakePresets.json

adamkewley:
Quick fix (won't wait for review): this hard-coded VS needs to be removed because it prevents Simbody from building on CI or developer machines with VS2026 etc.

If there's a desire to change VS version then it should be done via the developer's/CI's environment (e.g. by sourcing vcvars.bat or similar).

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/862)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -18,7 +18,6 @@
                 "lhs": "${hostSystemName}",
                 "rhs": "Windows"
             },
-            "generator": "Visual Studio 17 2022",
             "architecture": {
                 "value": "x64"
             }


halleysfifthinc:
:+1: SGTM. I'm less familiar with building on Windows, but it seems like the CMake auto-sets/discovers the generator? (So the only reason to explicitly set it is for CI build env consistency?)

adamkewley:
@halleysfifthinc it usually auto-detects it, but might require a nudge if it isn't on the PATH. e.g. for OSC I need to `call vcvars.bat` to activate a particular VS so that `cmake`, `ninja`, etc. match the one used in the IDE

PR #863, opened 2026-08-25
Fix OBJ face parsing with mixed whitespace

### LINKED ISSUE #747: The OBJ file parser should ignore all kinds of whitespace for face (`f`) commands
Minor bug, but it popped up when a user supplied custom-computed mesh data to OpenSim Creator (see: https://github.com/ComputationalBiomechanicsLab/opensim-creator/issues/473)

The bug is that SimTK's OBJ parser will fail to parse OBJ files that use tabs as delimiters in the face (`f`) command.

A quick search on the internet indicates that OBJ fields do not have a strictly specified field delimiter. Examples and files I have found on the internet indicate that fields may be separated with:

- A single space
- A newline character (with `\` as a control character)
- A single tab
- A single "other" whitespace character (vertical tabs)
- (maybe) Multiple instances of the above

E.g. Blender's OBJ parser handles all of the above by just treating all control characters the same way:

- (call point): https://github.com/blender/blender/blob/a16dd407b3d3f830c40646fcda5ea5b89b3ad4ee/source/blender/io/wavefront_obj/importer/obj_import_file_reader.cc#L499
- (implementation): https://github.com/blender/blender/blob/a16dd407b3d3f830c40646fcda5ea5b89b3ad4ee/source/blender/io/wavefront_obj/importer/obj_import_file_reader.cc#L390

SimTK's OBJ loader currently handles this fairly liberal specification for vertex commands (`v`), but not face commands (`f`) in OBJ files. This is because the vertex parser uses `std::istream::operator>>` to read each vertex element:

- https://github.com/simbody/simbody/blob/d685ed2/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L174

Where C++ specifies that the operator:

> Extracts a floating-point value potentially skipping preceding whitespace. (see: https://en.cppreference.com/w/cpp/io/manip/skipws)

But the face commands (`f`) use `std::istream::ignore` with a space character (`' '`):

- https://github.com/simbody/simbody/blob/d685ed2/SimTKcommon/Geometry/src/PolygonalMesh.cpp#L191

Where C++ specifies that `ignore` skips that character multiple times until it doesn't see it (in this bug, until it sees `\t` and stops).

The solution to this is to change that code section to skip as much *whitespace* as possible, rather than specifically the space character.

tandede:
## Summary

- let formatted face-token extraction handle tabs, vertical whitespace, form feeds, and repeated separators
- preserve existing OBJ line continuation and slash-delimited vertex/texture/normal parsing
- add focused regression coverage for mixed whitespace across ordinary and continued face records

## Problem

The face parser already reads each vertex token with `operator>>`, which skips leading whitespace. It then calls `ignore(..., ' ')` after every token. When a face uses a separator other than a literal space, that extra call consumes subsequent vertex tokens while searching for a space.

The regression test fails on the previous implementation because a four-vertex face is parsed with fewer than four vertices.

## Implementation

Remove the redundant literal-space `ignore` call and rely on the formatted extraction that already drives the loop. This keeps token boundaries intact for all stream-recognized whitespace without changing the parsing of `v/vt/vn` triplets or backslash-continued lines.

The new test covers tabs, vertical tabs, form feeds, repeated spaces, a continued face line, and texture/normal indices.

## Validation

- `cmake -S . --preset ci-make-macos-release -DINSTALL_DOCS=OFF`
- `cmake --build --preset ci-make-macos-release --parallel 4`
- `ctest --test-dir out/build/ci-make-macos-release --parallel 4 --output-on-failure` (114/114 passed)
- `clang-format --dry-run --Werror` on the added test
- `git diff --check`
- no-tab check on both changed files

Fixes #747.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/863)

### DIFF
--- a/SimTKcommon/Geometry/src/PolygonalMesh.cpp
+++ b/SimTKcommon/Geometry/src/PolygonalMesh.cpp
@@ -326,7 +326,6 @@ void PolygonalMesh::loadObjFile(std::istream& file) {
                     nIndex--;
                     normalIndices.push_back(nIndex);
                 }
-                s.ignore(line.size(), ' ');
             }
             if (getNumFaces() == 0) { // First encountered face, set expectations
                 expectNormals = (indices.size() == normalIndices.size());
@@ -1295,4 +1294,3 @@ createCylinderMesh(const UnitVec3& axis, Real radius, Real halfLength,
 
     return cyl; // just a shallow copy
 }
-
--- a/SimTKcommon/tests/TestPolygonalMesh.cpp
+++ b/SimTKcommon/tests/TestPolygonalMesh.cpp
@@ -114,6 +114,39 @@ void testLoadObjFile() {
     ASSERT(!mesh.hasTextureCoordinates())
 }
 
+void testLoadObjFileWithWhitespaceDelimitedFaces() {
+    string file;
+    file += "v 0 0 0\n";
+    file += "v 1 0 0\n";
+    file += "v 1 1 0\n";
+    file += "v 0 1 0\n";
+    file += "vt 0 0\n";
+    file += "vt 1 0\n";
+    file += "vt 1 1\n";
+    file += "vt 0 1\n";
+    file += "vn 0 0 1\n";
+    file += "f\t1/1/1\t2/2/1\v3/3/1\f4/4/1\n";
+    file += "f  1/1/1\t2/2/1 \\\n";
+    file += "  3/3/1\v4/4/1\n";
+
+    PolygonalMesh mesh;
+    stringstream stream(file);
+    mesh.loadObjFile(stream);
+
+    const Vec2 expectedTextureCoordinates[] = {Vec2(0, 0), Vec2(1, 0),
+                                               Vec2(1, 1), Vec2(0, 1)};
+    ASSERT(mesh.getNumFaces() == 2);
+    for (int face = 0; face < mesh.getNumFaces(); ++face) {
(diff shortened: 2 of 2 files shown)


adamkewley (approved):
:lgtm: - nice fix

Unrelated, but it's rather unfortunate that it also accepts non-standard inputs, like vertical tabs and line continuations, but the original parser already did this. This PR is just fixing a bug where the parser over-eats the input (e.g. if tabs are delimiting two faces then it will eat the second face trying to find the space for `.ignore`.

> @adamkewley reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/863#-:-P-shNWm1F6wq87MxZgS:b1lpd52)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on tandede).

tandede:
Thanks for the review and for clarifying the scope. Agreed—this PR only fixes the input over-consumption caused by `.ignore()` and intentionally leaves the parser's existing support for vertical whitespace and line continuations unchanged.

sherm1 (approved):
Thanks, @tandede! I'll merge this now.

> @sherm1 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/simbody/simbody/863#-:-P-tp5NA-R5F138lr1TL:b-8lzg98)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on tandede).

Issue #856, opened 2026-05-04
Support for scaled system Jacobians


nickbianco:
I'm currently working on a project to develop optimization methods for fitting OpenSim models both to optical motion capture data and video-based data sources. For this, it will be beneficial to have access to methods for computing analytical derivatives of functions computing the distance between two markers or the orientation between two frames. I've already incorporated existing system Jacobian methods in a custom inverse kinematics framework with CasADi, and you can see [an example of that approach here](https://github.com/opensim-org/opensim-fitter/blob/main/src/osimfit/callbacks.py). 

To support model scaling simultaneously with inverse kinematics, I'd like to implement support for _scaled_ versions of the the system (and frame/station) Jacobians, which I'm proposing here. The idea is to add methods similar to the existing Jacobian methods, but that also accept as argument a set of body scale factors to compute what the Jacobians would be if the system were scaled. 

I've already begun protoyping these methods in SimbodyMatterSubystem here: https://github.com/nickbianco/simbody/pull/1/changes.

### Theory

The proposed methods accept a vector of scale factors, one `SimTK::Vec3` for each body in the system. For a given mobilizer, the body scales associated with the parent body will scale the position offset of parent's mobilizer frame F. Similarly, the body scales associated with the child body will scale the position offset of the child's mobilizer frame M. If the mobilizer offset frames are rotated relative to the parent frame, we would need to project  See the mobilizer diagram below fore reference.

<img width="522" height="450" alt="Image" src="https://github.com/user-attachments/assets/8a8a7167-342f-4748-b83b-2b71d59602f4" />&nbsp;&nbsp;&nbsp;

[...]

PR #866, opened 2026-09-03
Add position Jacobian operators w.r.t. mobilizer frames

nickbianco:
Part 2 of addressing #856.

This change builds on #859, adding a new set of Jacobian operators to calculate the change in the position of a mobilized body with respect to changes in inboard and outboard mobilizer frames and their transpose. The following operators are added to `SimbodyMatterSubystem()`:

 - `multiplyByPositionJacobianWrtInboardFramePositions()`
 - `multiplyByPositionJacobianWrtInboardFramePositionsTranspose()`
 - `multiplyByPositionJacobianWrtOutboardFramePositions()`
 - `multiplyByPositionJacobianWrtOutboardFramePositionsTranspose()`

The forward operators take a change or perturbation in mobilizer inboard (F) or outboard (M) frames and calculate the resulting change in mobilizer body (B). The transpose operators are arguably more useful, since they allow taking a gradient-like quantity $`g_{GB}  = \partial{f} / \partial{p_{GB}}`$ and map it to something like $`g_{PF}  = \partial{f} / \partial{p_{PF}}`$. If $`f`$, for example, is a function that computes the distance between a measured point and a body origin, then the transpose operator can be used to assemble a gradient with respect to the mobilizer frame position. This is directly useful if you want to optimize the geometry of a system (e.g., the offsets of inboard/outboard frames) while tracking experimental data.

The underlying math is relatively simple: a shift in an inboard or outboard frame will shift all mobilizer positions down the kinematic chain, and the rotations `R_GB` and `R_GP` re-express the position shifts in ground. The transpose operators accumulate shifts for all children up to the current node, similar to other transpose operators.

The implementation follows a similar pattern to the other Jacobian operators: accessors and argument checking in `SimbodyMatterSubsystem`, an internal representation in `SimbodyMatterSubsystemRep` for creating intermediate data structures (e.g., for handling transpose accumulations), and a per-node implementation in `RigidBodyNode`. Since ground always contributes zero motion and all mobilizers handle inboard and outboard frames the same way, the current implementation is written directly in `RigidBodyNode` with no virtual dispatches.

For testing, I've added `TestJacobians.cpp` that covers all the new operators using a branch multibody system containing a variety of different mobilizers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/simbody/simbody/866)

### DIFF
--- a/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
+++ b/Simbody/include/simbody/internal/SimbodyMatterSubsystem.h
@@ -1224,6 +1224,130 @@ SpatialVec calcBiasForFrameJacobian(const State&         state,
     return JFDotu[0];
 }
 
+/** Calculate the product of the body position Jacobian with respect to
+inboard frame positions, JpF, with a vector of nb position-like quantities,
+dp_PF. If dp_PF is a vector of perturbations in mobilizer inboard frame (F)
+positions, expressed in the parent body P, then the result is the change in
+mobilized body (B) origin positions (Bo) measured and expressed in Ground.
+
+@param[in]      state
+    A State that has already been realized through Position stage.
+@param[in]      dp_PF
+    A vector of nb position-like quantities, one per mobilized body in the order
+    of MobilizedBodyIndex. The 0th entry is ignored since Ground has no inboard
+    frame. This is usually a set of perturbations in mobilizer inboard frame (F)
+    positions, expressed in the parent bodies P.
+@param[out]     dp_GB
+    The product JpF*dp_PF, usually the change in mobilized body (B) origin
+    positions (Bo), measured and expressed in Ground. The 0th entry is set to
+    zero since Ground does not move.
+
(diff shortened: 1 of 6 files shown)


adamkewley:
@nickbianco I can have a look at this from a C++ PoV but the maths/utility etc. is a little beyond me (as in, it's not that this PR is a bad idea, but I'd really have to study up on how Jacobians can be used in various ways etc. etc.)

sherm1:
@aseth1 any interest in reviewing @nickbianco's math?

nickbianco:
I've addressed the issue that Copilot pointed out.

If it's helpful for review here, I've outlined the math related to this PR in a document I've been assembling related to the project that relies on the proposed changes. Chapter 2 contains the related (and hopefully not butchered) math and algorithms: [multibody_jacobians.pdf](https://github.com/user-attachments/files/32251020/multibody_jacobians.pdf).

nickbianco:
@carmichaelong is going to help review this one.

